PR #31, opened 2018-01-25
feature/done-locally

mblomdahl:
Now enables authentication with Libris Login locally.


oBlissing:
We're having a problem with links in the data pointing to ``https://127.0.0.1/<libris-id>``

Unsure where this has to be changed, but since the links are set in the source data, and is set when the example data is imported, I think it's not in this repo, and this pull request should be OK.

Ping @niklasl

PR #33, opened 2018-04-27
Include OAuth2 scopes in authorization call

mxtthias:


mxtthias:
@oBlissing **DON'T MERGE!** This will be part of Monday's changes.

PR #37, opened 2018-05-03
Use ID_PATH for vocab, context and display definitions

oBlissing:
Enables loading vocab, context and display definitions as external resources instead of hardcoded JSON (for display) and instead of libris origin (for context and vocab).

Requires the ID_PATH to be set in api_config.json (look at example).


oBlissing:
Proceed as soon as https://github.com/libris/devops/pull/19 is merged.

PR #67, opened 2018-05-09
Translate strings in footer

sannana:
Removed jquery code and added some other code to make strings in footer translatable again.


viewer/v2client/src/components/layout/footer.vue
```diff
@@ -9,19 +12,32 @@ export default {
     }
   },
   methods: {
+    translate() {
+      const langCode = this.settings.language;
+      let translationEls = document.querySelectorAll('.js-translateable');
+
+      _.each(translationEls, function (translationEl) {
+        console.log(translationEl);
```
mxtthias: Remove

sannana:
Removed console logging and also mapGetters that are not used

PR #74, opened 2018-05-15
Feature/shortcut keys

sannana:
- Added minor navbar style fixes
- Made som typo changes for labels
- Restored keyboard shortcut bindings for inspector
- Added keyboard shortcut listener for field adder window
- Changed keybinding for undo, because IMO cmd/ctrl+z should not be hijacked from the OS
- Restored keybindings for field adder


sannana:
Fixed undo error in previous commit

PR #87, opened 2018-05-24
Use expanded uri in getTermByType

oBlissing:
* Fixes item-vocab's (aka dropdowns) where the list was empty.


oBlissing:
@haraag issuanceType field should work correct now 🎈

PR #95, opened 2018-05-28
Change style of holding information

sannana:


sannana:
Added features from another ticket as they are related

PR #103, opened 2018-05-29
Sort all forms according to display

oBlissing:
* Moved a lot of form-functionality into `form-mixin`, so that it works the same between `entity-form`, `item-sibling` and `item-local`.
* Make `DisplayUtil.getProperties` respect `fresnel:super` and `fresnel:extends`.
* Add `DisplayUtil.getLensById`.
* Fixed so that `subClassChain` always contains curie:s.

Other changes:
* Renamed `specialProperties` to `hiddenProperties`.
* Added flushing of `changeHistory` on document load.
* Disallow `abstract`, `unstable` and `shorthand` properties from being added. (LXL-1097)

### DIFF
--- a/viewer/v2client/src/components/mixins/form-mixin.vue
+++ b/viewer/v2client/src/components/mixins/form-mixin.vue
@@ -1,5 +1,6 @@
 <script>
 import * as DataUtil from '@/utils/data';
+import * as DisplayUtil from '@/utils/display';
 import * as VocabUtil from '@/utils/vocab';
 import * as StringUtil from '@/utils/string';
 import * as _ from 'lodash';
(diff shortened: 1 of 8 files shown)


viewer/v2client/src/utils/display.js
```diff
@@ -68,7 +86,26 @@ export function getProperties(typeInput, level, displayDefs) {
...
+
+    if (props.length > 0) {
+      return props;
+    } else if (level === 'full') { // Try fallback to card level
+      props = getProperties(type, 'cards', displayDefs);
+      if (props.length > 0) {
+        return props;
+      }
+    }
     if (props.length > 0) {
```
mxtthias: With the new `if` statement above, will this case ever be true?

viewer/v2client/src/components/mixins/form-mixin.vue
```diff
@@ -24,6 +25,78 @@ export default {
       'settings',
       'status',
     ]),
+    formType() {
+      return this.formObj['@type'];
+    },
+    filteredItem() {
```
mxtthias: Has this code been updated too or just moved?

PR #104, opened 2018-05-30
Escape double slash in path (except for 'https?://')

mxtthias:
This is needed since Flask treats all slashes as path separators and
unescapes them.


niklasl:
The problem appears to be a double error by Apache and WSGI(!). See this comment for a solution: https://github.com/pallets/flask/issues/900#issuecomment-27378055

If we deem this not feasible (at this time), let's merge this and note the proper fix (to also handle single escaped slashes).

PR #110, opened 2018-05-31
 Add pagination to entity-adder modal and search-window modal

oBlissing:
* Added `modal-pagination` component.
* Added above to `search-window` and `entity-adder`.
* Added translations for "First", "Last", "Previous", "Next" (also added use of these to old pagination code).

Ways to improve on this:
* It shows way too many pages at the same time if you make a big search and that looks pretty bad. Maybe do like the pagination in the main search (show 10 or so pages around the `currentPage`).
* The layout code (html/css) between the two modals differ a lot. We should change this.


viewer/v2client/src/components/inspector/entity-adder.vue
```diff
@@ -717,10 +746,20 @@ export default {
...
+}
+.EntitySearchResult {
+  &-fetchMore {
+    text-align: center;
   }
+}
 
+.ScrollMarginTop {
+  padding-top: 95px !important;
+  // If you question this, feel free to rewrite the layout of this modal.
```
sannana: attitydsproblem

PR #129, opened 2018-06-25
Fixed entity-adder not resetting previous search when adding from the list

oBlissing:
Moved all the reset functionality into one method.

* Removed `searchOpen` property (was unused)
* Removed `chooseLocalType` property (was unused)
* Removed `closeSearch` method (use `hide`)
* Removed `openSearch` method (use `show`)
* Added `resetSearch` method (called in `show`)

Also:
* Removed some scattered code from old pagination


oBlissing:
More changes incoming on this request. Hold the merge!

PR #130, opened 2018-06-26
save user's remote databases

jesperengstrom:
Targets [LXL-1344](https://jira.kb.se/browse/LXL-1344)

* Logged in user's selected remote databases are stored in new settings property `defaultDatabases`.
* `defaultDatabases` is set to `OCLC` by default.
* For non-logged in users nothing is stored or selected by default (maybe a later todo?)


oBlissing:
I approve this request 👍

PR #151, opened 2018-08-15
Enter also triggers summary-action's action

jesperengstrom:
(add entity btn in modal for example)


oBlissing:
Is this only when the button is selected with tab? If that is the case then this looks good to me, too. 👍

PR #153, opened 2018-08-20
Feature/breadcrumb

sannana:


oBlissing:
Looks very good, but I'll hold my approval until the user is warned about navigating away when they have unsaved changes. Just check the computed var `unsavedChanges` in inspector.

sannana:
TY, good input

PR #164, opened 2018-09-03
Feature/update link entity window

sannana:


viewer/v2client/src/components/inspector/search-window.vue
```diff
@@ -330,33 +330,29 @@ export default {
...
-              <select 
+              <filter-select class="EntityAdder-filterSearchInput FilterSelect--insideInput"
+                :class-name="'js-filterSelect'"
+                :custom-placeholder="'All types:'"
+                :options="selectOptions"
+                :options-all="getRange"
+                :is-filter="true"
+                :options-selected="''"
+                v-on:filter-selected="setFilter($event, keyword)"></filter-select>
+              <!-- <select
```
oBlissing: Can we remove this commented out code? Is it used anymore? If not, just remove it. Or maybe we have a good reason to keep it?

PR #170, opened 2018-09-10
Weird cursor focus fix

jesperengstrom:
[LXL-1694](https://jira.kb.se/browse/LXL-1694)
Was caused by a function doing this explicitly with no apparent benefit. Removed it and its reverse sibling 👭.


oBlissing:
I *think* the reason it was there in the first place was to automatically focus new fields when they were added (so as to help keyboard users). But I also think that is a bad way to do things, as it changes the expected behaviour of the browser.

PR #176, opened 2018-09-17
Null links in footer fix

jesperengstrom:
[LXL-1505](https://jira.kb.se/browse/LXL-1505)

Was caused by translate function setting a prop as `innerHTML`, even when it wasn't found, ie `null`.
However, was only rendered by IE & Edge, Chrome skipped it.


sannana:
GOOD FIND!!

jesperengstrom:
at @oBlissing:s suggestion I'll try to swap local translate function with the newer `translatePhrase`. So hold da merge...

PR #177, opened 2018-09-18
update text in search-window

oBlissing:
Clarify flow so that user understands what is happening.


oBlissing:
I am rejecting this myself 🔥

PR #193, opened 2018-10-02
Feature/link entity summaries

sannana:


sannana:
I tried simply removing add from the whole list element and only keeping it on the button. Might be confusing to have a link within a link with different targets/results?

PR #211, opened 2018-10-15
Feature/Close panel-default with ESC-key

sannana:


sannana:
I wrote some thoughts on this in corresponding JIRA-ticket.

PR #217, opened 2018-10-16
Render head before body

sannana:


oBlissing:
We need to confirm that the "old browser" text is displayed without modernizr.

oBlissing:
We have some code that depends on modernizr for detecting flexbox and ie11.

Have you seen this thread about the input localization test?
https://github.com/Modernizr/Modernizr/issues/1998

oBlissing:
We have some code that depends on modernizr for detecting flexbox and ie11.

Have you seen this thread about the input localization test?
https://github.com/Modernizr/Modernizr/issues/1998

PR #222, opened 2018-10-19
Add same functionality to item-sibling that item-local has

sannana:


viewer/v2client/src/components/inspector/item-sibling.vue
```diff
@@ -270,7 +258,7 @@ export default {
     },
   },
   watch: {
-     isEmpty(val) {
+    isEmpty(val) {
       if (val) {
         this.$el.getElementsByClassName('js-expandable')[0].classList.add('is-inactive');
```
oBlissing: This watcher is not needed. Just bind the computed property it in the template.
Example:

`<div :class="{ 'is-inactive': isEmpty }">nånting</div>`

PR #235, opened 2018-10-24
Edge & older Chrome scroll bug fix

jesperengstrom:
Issue [LXL-1916](https://jira.kb.se/browse/LXL-1916)


oBlissing:
Nice find. I'll test it ASAP.

PR #246, opened 2018-10-30
Highlight local entity fix

jesperengstrom:
A fix so that new item highlighting works in case described in [LXL-1964](https://jira.kb.se/browse/LXL-1964). As a bonus, adder (+) focus also works.

Was caused by an index value always being appended to the path, even if it was a single (empty) object, resulting in a path like `somePath[isNaN]`. Fix does a check to prevent this.


jesperengstrom:
@oBlissing can be reviewed again... I think the easiest way to prevent double highlight is to simply remove transparency from blue color :)

PR #248, opened 2018-10-30
Download compiled Marc in Edge

jesperengstrom:
[LXL-1894](https://jira.kb.se/browse/LXL-1894).

This probably has to do with Edge limiting the length of href URL:s, a long standing issue with data URI:s, see [here](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/4282810/)

Implemented Microsoft's recommendation to create a [URL from a blob](https://textslashplain.com/2018/08/06/script-generated-download-files/) instead, although I can't compare the output in local env - just empty files. Thoughts are welcome 🙂


oBlissing:
Perhaps we can spin up this feature branch on dev2 and test it.

jesperengstrom:
Up on dev2, tested and approved by Elisabet @oBlissing

PR #251, opened 2018-11-01
Move type select with portal

jesperengstrom:
[LXL-1987](https://jira.kb.se/browse/LXL-1987): Move the select menu that lets you choose what type of property to add. Today it's absolute positioned on top of entity adder. We want to make it a placeholder where the new item will be placed.

Doing this with dynamically named portals do send the right select menu to the right place. This, in turn requires some extra functionality:

* We want to autofocus it, i.e need to know when it's mounted -> refactored it into own component `<type-select>`
* Possible long distance from (+) button to menu -> implemented autoscroll 
* New solution works badly with clickaway -> added 'remove' button instead. Remove is not available when the field is empty.


jesperengstrom:
@oBlissing evil bug should be solved now ✌️

PR #255, opened 2018-11-05
Disable unused tests

sannana:
I disabled the test  _inputnumber-l10n_ because it we don't use it and it was causing` body` to render before` head`. 

I have also checked for which other test we do not use classes of at this point and disabled them. It seems like we only check for `ie11` and `no-flexbox`, but I kept most of the CSS tests just in case.


oBlissing:
Can we put this up on dev2 and test that old browsers still get the "unsupported" page?

oBlissing:
Also, I know that we have a css selector looking for .no-flexbox
BUT I can't see any flexbox tests (even before this PR). So maybe we should add that?

PR #256, opened 2018-11-05
Change display type

sannana:


sannana:
This will fix inline-block invisble margin problem in IE11.

sannana:
Also removed `max-height: auto` because the value "auto" doesn't exist for that attribute anyway.

PR #267, opened 2018-11-08
Feature/template flow (aka "Berika från mall" + other things)

oBlissing:
It's now possible to download "in progress" files as JSON-LD to embellish a post "from a file".
It checks that the type of the mainEntity of a post is the baseClass or the same of that in the file supplied by the user. The same check is applied to our "official" templates to calculate what templates are valid on what posts.

Also:
Added an "overwrite"-function so that power users (aka Elisabet & her crew) can fix posts like they have in the past. This is only available when the "debug mode" user setting is active.

Bonus:
The "from file" views now have an extra file input button in addition to the "drop your file here"-version.


oBlissing:
@niklasl Which of the features do you mean?

If you're talking about the old "create new from file" which didn't clean @id and similar properties, this is now moved to the "Overwrite file" function (hidden under debug setting). The new "Creat new from file" will strip all those values.

The "embellish from file" uses the same mechanic as the old "embellish from template", so it doesn't edit any properties that already has a value (which means it won't replace @id since it will always be present in some form, maybe just a TEMPID, but it will still be a value). I should probably pipe the file through the same mechanic so that it doesn't add a sameAs if it's missing.

Will push another commit to fix this.

oBlissing:
Actually, that already happens ^^

I have nothing to add.

PR #271, opened 2018-11-13
Use publication.year in facets

mxtthias:
This replaces publication.date, since that field tends to include
additional information.

### DIFF
--- a/viewer/dataaccess.py
+++ b/viewer/dataaccess.py
@@ -81,7 +81,7 @@ sites = {
                     "sortOrder":"asc",
                     "size":100
                 },
-                "publication.date":{
+                "publication.year.keyword":{
                     "sort":"key",
                     "sortOrder":"desc",
                     "size":50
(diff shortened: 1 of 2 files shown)


niklasl:
It seems date has a `raw` in ES config, but does year have one too?

mxtthias:
Hmm.. The mappings I have locally (after running `recreate_storage`) differ from both PROD and DEV. Locally the field is called `raw`, but in the other envs it's called `keyword`. I'll take a look and try to figure out why.

mxtthias:
Found the error, but encountered another one.

Hold off on this PR for a while. I'll ping you when it's ready.

mxtthias:
Updated, but needs the fix in https://github.com/libris/librisxl/pull/254

mxtthias:
`keyword` is a feature in ES, see https://www.elastic.co/blog/strings-are-dead-long-live-strings. The `raw` stuff was a leftover from some forgotten experimentation I had done on my local machine..

PR #283, opened 2018-11-16
Prevent linking of empty entites

jesperengstrom:
[LXL-2101](https://jira.kb.se/browse/LXL-2101)


jesperengstrom:
On second thought, is it better to disable the icon rather than removing it, to show where linking _would_ be possible?

PR #287, opened 2018-11-19
Updates of Cartographic/StillImage templates that we agreed upon workshop

sarwik:


sarwik:
I have tested it locally (together with the definitions updates) and it looks good

PR #291, opened 2018-11-20
Add paste button in inline fields

oBlissing:
Paste button was missing from inline (or nested) fields.


oBlissing:
Merge is approved by test lead.

PR #299, opened 2018-11-21
Remove a lot of code from old client

oBlissing:
* Removed all proprietary JS code
* Removed almost all dependencies
  * Includes removal of bower (moved dependencies to npm usage and rewrote vendor build bash script)
  * Only dependencies left are uglify-js (for vendor.js build), less, jquery, bootstrap, kungbib-styles and some script helpers.
* Rewrote some less that was depending on code inside removed vue components.


oBlissing:
Will build this to dev2 and se if things catch fire or not.

PR #305, opened 2018-11-27
Get ETag from response header instead of data

oBlissing:
Needed for [LXL-2178](https://jira.kb.se/browse/LXL-2178) to work.

### DIFF
--- a/viewer/v2client/src/components/Inspector.vue
+++ b/viewer/v2client/src/components/Inspector.vue
@@ -152,7 +152,7 @@ export default {
       const fetchUrl = `${this.settings.apiPath}/${this.documentId}/data.jsonld?${randomHash}`;
       fetch(fetchUrl).then((response) => {
         if (response.status === 200) {
-          this.documentETag = response.headers.get('ETag').replace(/"/g, ''); // Fix double quoted string
+          this.documentETag = response.headers.get('ETag');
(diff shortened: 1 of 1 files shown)


viewer/v2client/src/components/Inspector.vue
```diff
@@ -151,6 +152,7 @@ export default {
       const fetchUrl = `${this.settings.apiPath}/${this.documentId}/data.jsonld?${randomHash}`;
       fetch(fetchUrl).then((response) => {
         if (response.status === 200) {
+          this.documentETag = response.headers.get('ETag').replace(/"/g, ''); // Fix double quoted string
```
mxtthias: Why strip quoting?

mxtthias: Backend sends a quoted string in the ETag header. I would have assumed that frontend should treat it as an opaque value and not manupulate it in any way?

oBlissing: Got it, will fix and update the PR.

niklasl: The keys is to make the subsequent `If-None-Match` identical. So if we get:
```
    ETag: "abc"
```
Do we send:
```
   If-None-Match: "abc"
```
?

PR #317, opened 2019-01-10
Fix path for extractions and block Concept chain from extraction

oBlissing:
- Removed baseclass `Identity` from extractable classes, this removes `Concept`, `Agent`, `Work` from extraction
- Add `Agent` and `Work` to extractable classes.

- Add `/data` to apiPath when making extraction calls.


jesperengstrom:
👍

PR #329, opened 2019-01-15
feature/sort-class-fix

jesperengstrom:
Current implementation of the sort component only allow one searched `@type` since we need to get relevant sort options/parameters for it. This completely ignored subclasses and the fact that multiple types can be a subset of Instance for example.

This fix allows an array of searched types to be sent to the sort component. It then reduces it and if they all belong to the same recordtype, we then get sort options for it.

For example: good when using the 'type' facets.


oBlissing:
Nice catch and solution 👍

PR #337, opened 2019-01-23
Template functionality updates

oBlissing:
* It's now possible to embellish from an existing post by ID reference
* Automatically updates any heldBy-keys to the active users sigel (so that they can use holding-templates created by another sigel).

+ Implemented DataUtil.rewriteValueOfKey and added 🍾UNIT TESTS 🍾 for that.


jesperengstrom:
On another note, maybe 'x field(s) added from template' qualifies as a `success` notification?

PR #339, opened 2019-01-28
Add workaround for IE11 when downloading Blob

oBlissing:
Same workaround has been used for `Download Compiled Marc` before. This is for the "download json" and "download json (incl unsaved changes)".

We can probably spin this branch on dev2 to check it.


oBlissing:
Tested on IE11 on dev2. Merging.

PR #341, opened 2019-01-30
Save expansion state on toggle

jesperengstrom:
[LXL-2005](https://jira.kb.se/browse/LXL-2005) - save fields' expansion state when toggling between instance/record.

Instead of saving the state of each field in vuex for example (which seemed like a lot of data for nothing _that_ important) I thought about saving their state by simply keeping the component in the DOM.

Today we render a single `<entity-form>` and fill it with different data when toggling focus. This approach renders one `<entity-form>` (incl data) per tab-item. Toggling is then merely a `v-show` switch. 

Everything seems to work as before, but we have to manually close some panels on switch since their parents are never unmounted. 

**Advantages:**
+ state is preserved
+ switching is much faster

**Disadvantages:**
???

### DIFF
--- a/viewer/vue-client/src/components/inspector/entity-adder.vue
+++ b/viewer/vue-client/src/components/inspector/entity-adder.vue
@@ -125,6 +125,9 @@ export default {
           case 'close-modals':
             this.hide();
             break;
+          case 'focus-changed':
+            this.hide();
+            break;
           default:
         }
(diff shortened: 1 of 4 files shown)


oBlissing:
Haven't tried this yet. But as long as the performance isn't worse than before, I really like this solution.

PR #368, opened 2019-02-25
Don't use `.keyword` suffix in facets

mxtthias:
The backend takes care of this for us.


mxtthias:
Must be merged **after** https://github.com/libris/librisxl/pull/306

PR #371, opened 2019-02-26
Adjust search help

oBlissing:
* Only show on toggle, not hover
* Adjusted size and position
* Added tooltip on icon

(If trying on local, do remember that the text will not load, the interaction and position should be good though).


oBlissing:
@jesperengstrom ready to be tested again :)

PR #377, opened 2019-02-26
Guard route

jesperengstrom:
Had an idea to block directory-care route for un-authed users, but it turns out no routes that really require login are actually guarded (ie `/new`, `/create`, `/user`)

Waddabout this solution 🤔


jesperengstrom:
👍 👍

PR #380, opened 2019-02-28
Error notifications instance fetch

jesperengstrom:
Displays the following notfications when fetching flagged instances:

Some failed - "Failed to retrieve x of x resources"
All failed - "Failed to retrieve resources"
Fetch error - "Something went wrong"

Not displaying WHAT id:s failed, why etc. Should we?


oBlissing:
Closing because git shenanigans...

PR #391, opened 2019-03-05
IE fixes

jesperengstrom:
Checked Safari, Firefox, Edge and they look & behave ok as far as I can tell.
In IE there were some strange height issues with flexbox that this hopefully fixes.


jesperengstrom:
Should look ok now

PR #428, opened 2019-04-02
[WIP] Update templates and snippets containing local entities of inScheme:

sarwik:
- Change the type of inScheme to appropriate subClass of ConceptScheme

This only affects local entities of ClassificationScheme. We need to discuss if this should be merged together with https://github.com/libris/definitions/pull/103 or if we should wait until we change mapping of ConceptScheme i marcframe?


oBlissing:
@sarwik is this still being worked on?

oBlissing:
@sarwik Can we close this?

sarwik:
@oBlissing Yes, close it 👍

PR #431, opened 2019-04-04
LXL-2432 Removable filter

jesperengstrom:
Reworked `search-controls` component according to design in [ticket](https://jira.kb.se/browse/LXL-2432) including:

* Display active facets as removable filter badges
* New copy / search info

Also 
* Made type checkboxes + search bar query + searchtype dropdown reactive to route changes to always reflect the actual search (when using the back-button for example) - necessary since we no longer print out the searchphrase in search-controls info.

**U may need to restart Vagrant to include an API fix that fixes facet 'up'-links**

**UPDATE:**
Now also includes [LXL-2453](https://jira.kb.se/browse/LXL-2453) since it's practically an extension of the code above. The current search param (Free text, ISBN, Main title etc) is now stored in user settings and reused on a clean search. 

The 'active' option now falls back like: 
Presence of existing query param --> user setting param --> Fallback/default value

Test: See user story in ticket.

### DIFF
--- a/viewer/vue-client/src/components/search/result-controls.vue
+++ b/viewer/vue-client/src/components/search/result-controls.vue
@@ -110,6 +110,17 @@ export default {
     currentSortOrder() {
       return this.$route.query._sort;
     },
+    resultRange() {
+      if (this.$route.params.perimeter === 'remote') {
+        return `1-${this.limit}`;
+      } 
+      const first = this.pageData.itemOffset + 1;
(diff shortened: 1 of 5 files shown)


oBlissing:
Opinion:
Can we reduce the padding-right on boxes and remove padding on icon?

Compare:
Screen Shot 2019-04-09 at 12 45 19
Screen Shot 2019-04-09 at 12 45 08

jesperengstrom:
Yep, done!

PR #435, opened 2019-04-11
LXL-2489 Fix property sort (without rollback)

jesperengstrom:
**Bug:** 
[LXL-2489](https://jira.kb.se/browse/LXL-2489)

**Probable cause:** 
`getDisplayProperties()` returns nothing for `Work` since it never looks beyond `'full'` level. So the `formObj` is rendered as-is, unsorted.

**Fix:**
Make `'full'` level fall back to `'card'` (like before).

### DIFF
--- a/viewer/vue-client/src/utils/display.js
+++ b/viewer/vue-client/src/utils/display.js
@@ -111,6 +111,15 @@ export function getDisplayProperties(className, displayDefinitions, vocab, setti
       level = 'chips';
     }
   }
+
+  if (level === 'full') {
+    props = getLensPropertiesDeep(cn, displayDefinitions, vocab, settings, context, level, depth);
+    if (props.length === 0 && depth === 0) {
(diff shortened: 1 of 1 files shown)


niklasl:
Looks reasonable in itself, unless @oBlissing intended to remove traversal here to enable full control of this behaviour in the display configuration? (I believe the desired result for Work could be achieved by defining a "full"-lens for Work and having that extend the "card"-lens for Work?)

oBlissing:
@niklasl is correct.

We should never fallback from `full` (or any level for that matter) to anything "smaller". That could potentially break a lot of things.

In a nutshell: If we want a `full`-order for Work, just define it :)

jesperengstrom:
Ok, @klngwll will make a new PR with more 'full' levels defined.

klngwll:
https://github.com/libris/definitions/pull/110 merged

PR #443, opened 2019-04-26
LXL-2102 Highlight embellished

jesperengstrom:
This highlights all added fields after embellish. Highlight arr is cleared on next add, on save, before destroy.

**No need to merge this (right away at least).** Can also be just a reference if we want to use logic in [LXL-2392](https://jira.kb.se/browse/LXL-2392).


oBlissing:
Looks good.

IMO the highlight should dissappear when you change something in the form, kind of like it does now but on all changes (like text input in a text field, removal of something etc).

jesperengstrom:
Yep. Now removing the highlight when Inspector gets a data update (checking that the update is not the embellish action itself), which seems to cover all changes in the form. Also resulted in less code!

PR #444, opened 2019-04-29
[WIP] [LXL-2392] Only scroll on fields added

oBlissing:
* If the added node is a field, ensure that is in viewport and highlight it
* If the added node is note a field, highlight it, wait for it to be inside the viewport and then remove the highlight.

Had to rewrite a bit of the path-code. Will need to test if it has any side-effects.


oBlissing:
Closing this with reason: Stale

PR #449, opened 2019-05-02
[LXL-2501] Decrease paddings in form, remove min-height on rows

oBlissing:


jesperengstrom:
Nice. 

(Maybe `Field-labelContainer`'s `padding-left` should be the same as `Field-content`'s `padding-right`. Whatever the value, doesn't have to be 20px...)

PR #450, opened 2019-05-03
Let item-sibling extend item-local

oBlissing:
We'll go through this together.


oBlissing:
Ah, yes. Copying of it is OK, however you should not be able to duplicate it (ie copying and pasting in the same field).

oBlissing:
(but I am disabling that for now)

oBlissing:
@jesperengstrom I have disabled the options menu for siblings.

oBlissing:
Closing this.

PR #473, opened 2019-06-20
LXL-2500 - Move facets and db handler to the left

oBlissing:
* Moves libris search filters to a leftmost column
* Moves remote database selection to same column instead of side-panel


oBlissing:
Created [LXL-2595](https://jira.kb.se/browse/LXL-2595) to address the checkbox issue.

I think the other this is a discussion we're about to have soon.

Merging.

PR #477, opened 2019-07-04
Adapt code to work on Python 3

niklasl:
This has only been rudimentarily tested, but is a fairly straightforward change to run under both Python 2 and 3. It would be good to have it run for a while on dev2 before merging, especially if we can set up mod_wsgi for Python 3 there (reasonably by using a flag in the devops env setting).

### DIFF
--- a/viewer/__init__.py
+++ b/viewer/__init__.py
@@ -1,5 +1,8 @@
 # -*- coding: UTF-8 -*-
 from __future__ import absolute_import, unicode_literals, print_function
+if bytes is not str:
+    unicode = str
+
 import json
 import operator
 import os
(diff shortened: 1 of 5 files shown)


oBlissing:
@niklasl do we run Python3 on any of the environments?

niklasl:
No, only locally. We need to set it up for all envs; either by doing more provisioning in Fabric, or by upgrading each environment manually, and taking care to support both deploy states (versions) unless everything is upgraded in union. (Specifically IIRC, ensuring mod_wsgi for Python 3,  and ideally a yum-managed pypy3 on xlbuild.)

PR #482, opened 2019-08-13
[LXL-2608] Implement parameterized resource views

olovy:
Forward query parameters used by back-end


viewer/__init__.py
```diff
@@ -250,9 +250,11 @@ def _get_view_data_accept_header(request, suffix):
 
 def _filter_query_params(request_args):
```
niklasl: TBH I don't know why we manage the query parameters here instead of just passing them through. I think it was introduced in da05ffc8bb329c320ef7fa3ba0f3381a7f22e45d (before that we didn't pass them at all). We could probably pass them through as is and skip the filtering altogether.

PR #533, opened 2019-11-25
LXL-2771 Add more options for search form

oBlissing:
Added:
* You can use form to search for `hasTitle.mainTitle` for type `Work`.
* You can use form to search for `prefLabel` for type `Concept`.


praktwerk:
👍

PR #536, opened 2019-11-28
Fix SummaryNode-label when card key not same as item key

olovy:


viewer/vue-client/src/utils/display.js
```diff
@@ -298,7 +298,7 @@ export function getItemSummary(item, displayDefs, quoted, vocab, settings, conte
       } else if (displayGroups.categorization.indexOf(key) !== -1) {
         summary.categorization.push({ property: key, value: v });
       } else {
-        const valueAsArray = isArray(item[key]) ? item[key] : [item[key]];
```
olovy: This breaks for e.g. "altLabel" which is `altLabelByLang` in `item` while key is `altLabel` at this point.
Do we do it this way for a reason?

oBlissing: Yes, there is a reason if I recall correctly. We can talk about that tomorrow and if I'm right I should really do another comment on this PR xD

PR #546, opened 2019-12-03
Fix broken layout in Firefox

praktwerk:
Make FieldContent take up remaining space inside it's parent


praktwerk:
🤦‍♂ missed that Field is not flex <768. Should be ok now!

PR #548, opened 2019-12-03
Apply derivedFrom when extracting entities

oBlissing:
Also:
* Move general code into mixin
* Lock derivedFrom property in app-settings (so it can't be deleted/edited in GUI)


praktwerk:
@oBlissing friendly reminder to add RecordUtil import :)

oBlissing:
> @oBlissing friendly reminder to add RecordUtil import :)

Done :)

PR #551, opened 2019-12-05
Only translate if language in settings is present in item

olovy:
Original behavior.
i.e. https://libris-qa.kb.se/katalogisering/1rznbwb534c3znk5 should not display altLabel in card if user language setting is "sv."


niklasl:
Isn't it "nice" to show some value (e.g. in english) rather than none in general?

oBlissing:
> Isn't it "nice" to show some value (e.g. in english) rather than none in general?

Considering this change doens't only affect `altLabel` but all `...byLang`-properties  it's probably a good idea to have some kind of fallback. Up for discussion I think.

PR #584, opened 2020-02-06
Feature/lxl 2775 sort by reverse links

praktwerk:
## Checklist:

## Description

Add sort select to side panel search results.
Styling adjustments, new style of select / filter in side panel.

### Tickets involved
[LXL-2775](https://jira.kb.se/browse/LXL-2775)

### Solves

The need of sorting side panel search results by relevance/most linked/last updated

### Summary of changes

Added new styling to the Sort and Filter Select component, both have two different styles now, the default one and material design-ish one.

Side panel search is now using the Sort component from main search to sort results.
In the Sort component, added a boolean commonSort param to override which returns common search params even if search result items have different base classes.

### DIFF
--- a/viewer/vue-client/src/components/inspector/search-window.vue
+++ b/viewer/vue-client/src/components/inspector/search-window.vue
@@ -535,16 +535,6 @@ export default {
     }
   }
 
-  &-input {
-    // color: @black;
-    // background-color: #fff;
-    // border: 1px solid @grey-lighter;
-    // &:focus {
-    //   border-color: @brand-primary;
-    // }
-  }
-
-
   &-filterSearchContainer {
     width: 100%;
     display: flex;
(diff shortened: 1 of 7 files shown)


praktwerk:
@oBlissing please have a closer look on https://github.com/libris/lxlviewer/pull/584/commits/67d7086d1088863f3bdd43dda65d219097f7c6b4 and https://github.com/libris/lxlviewer/pull/584/commits/3e6f8eaf8ca2bb55525aaea479acd6c8e41b7ac4 . In order to have sort on results with different base classes, I added a "Common" and "Other" sort definitions (relevance/last updated/most linked), and a commonSort param to the Sort component. 
For example, it makes it possible to sort subject or language by most linked or last updated.
However, I'm unsure if I missed some edge cases where those common sort definitions will brake something/not work as expected.

viewer/vue-client/src/components/inspector/search-window.vue
```diff
@@ -516,28 +536,45 @@ export default {
...
 
   &-input {
-    color: @black;
+    // color: @black;
+    // background-color: #fff;
+    // border: 1px solid @grey-lighter;
+    // &:focus {
+    //   border-color: @brand-primary;
+    // }
   }
```
olovy: Remove these?

olovy:
Code LGTM!

Maybe it's more of a UX-question but IMHO Concept and Agent should have the same sort options as other types (most linked. last modified, alphabetical).

PR #594, opened 2020-02-28
Feature/lxl 3030 tab order

praktwerk:
## Checklist:

## Description
Tab focus ending up in footer when navigating between routes.
Example: Navigate from search results to a post.

### Tickets involved
[LXL-3030](https://jira.kb.se/browse/LXL-3030)

### Solves

Tab navigation issues, specifically when navigating from search results to a single post.

### Summary of changes

Added negative tabindex to file input in Toolbar 

Placed Toolbar before Inspector in HTML, in order to get the tab order right

Added Bootstrap classes to "offset" Toolbar in order to keep it in the same position in the layout

in App.js, added a watch on $route, setting tabindex and focus to router-view when route changes

added a wrapper div in DirectoreCare to prevent the focus thing from throwing an error because of content not yet rendered

### DIFF
--- a/viewer/vue-client/src/App.vue
+++ b/viewer/vue-client/src/App.vue
@@ -18,7 +18,7 @@
             <p>Om felet kvarstår, kontakta <a href="mailto:libris@kb.se">libris@kb.se</a>.</p>
           </div>
         </div>
-        <router-view v-if="resourcesLoaded" />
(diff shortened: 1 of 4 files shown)


praktwerk:
@olovy definitely. Adding this as a separate issue, since it's isolated to a specific component.

PR #598, opened 2020-03-10
show reverse links amount on instance

praktwerk:
## Checklist:

## Description

For now, show reverseLinks for type Instance

### Tickets involved
[LXL-2929](https://jira.kb.se/browse/LXL-2929)

### Solves

Performance improvement (less calls to server)

### Summary of changes

Checking if mainEntity has reverseLinks and recordType is Instance, then show the amount
If false, do the o-query as usual


oBlissing:
Ship it ⛴

PR #617, opened 2020-04-02
Handle null mappings in marcframe view

olovy:
### Solves
https://jira.kb.se/browse/LXL-3084

Handles fields mapped to null in marcframe view template. Template would throw on these resulting in HTTP 500.

example marcframe.json
```
"Multimedia": {
        "[18] [19] [20] [21]": null,
``` 

Also hides some more instances of "TODO" in the marcframe view.

### DIFF
--- a/viewer/templates/marcframeview.html
+++ b/viewer/templates/marcframeview.html
@@ -200,7 +200,7 @@
       <tr>
         <th><div id="${cat}-${tag}-${key}">${key}</div></th>
         <td>
-          % for k, v in sorted((subdfn or {'': 'ignored'}).items())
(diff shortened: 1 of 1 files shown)


viewer/templates/marcframeview.html
```diff
@@ -197,8 +200,8 @@ <h3>${tkey}</h3>
       <tr>
         <th><div id="${cat}-${tag}-${key}">${key}</div></th>
         <td>
-          % for k, v in sorted(subdfn.items())
-            % if not k.startswith('TODO:') and v
+          % for k, v in sorted((subdfn or {'ignored': True}).items())
```
olovy, line 203: This is the fix for null values

PR #639, opened 2020-05-19
Update digitemplate

klngwll:
## Description
Update digi template:

* Add DIGI sigel in bibliography
* Update production/DigitalReproduction
* Remove ISBN from identifiedBy

### DIFF
--- a/viewer/vue-client/src/resources/json/combinedTemplates.json
+++ b/viewer/vue-client/src/resources/json/combinedTemplates.json
@@ -13,7 +13,7 @@
           "bibliography": [
             {
               "@type": "Library",
-              "sigel": [""]
+              "sigel": ""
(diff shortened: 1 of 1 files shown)


viewer/vue-client/src/resources/json/combinedTemplates.json
```diff
@@ -206,7 +206,7 @@
           "bibliography": [
             {
               "@type": "Library",
-              "sigel": [""]
+              "sigel": ["DIGI"]
```
sarwik, line 209: Should not be a list, right?

klngwll: You got that right! Hmm need to fix this in all templates then! Good catch

PR #655, opened 2020-06-15
Fix HTTP 500 on missing Accept header

olovy:
### Solves
Requests to `/<id>` without `Accept:` header would return HTTP 500.

`curl 'http://kblocalhost.kb.se:5000/n0xpcvb8l93wbdhq' -H 'Accept:'`

before:
`< HTTP/1.0 500 INTERNAL SERVER ERROR`

after:
`< HTTP/1.0 406 NOT ACCEPTABLE`

### DIFF
--- a/viewer/__init__.py
+++ b/viewer/__init__.py
@@ -244,9 +244,9 @@ def thingview(path, suffix=None):
 # FIXME make this less sketcy
 def _get_view_data_accept_header(request, suffix):
     mimetype, _ = negotiator.negotiate(request, suffix)
-    if mimetype in ('application/json'):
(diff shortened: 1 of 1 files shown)


viewer/__init__.py
```diff
@@ -244,9 +244,9 @@ def thingview(path, suffix=None):
 # FIXME make this less sketcy
 def _get_view_data_accept_header(request, suffix):
     mimetype, _ = negotiator.negotiate(request, suffix)
-    if mimetype in ('application/json'):
```
niklasl, line 247: This was the actual culprit. Testing for None in a tuple is OK, looking for None in a string is not.

PR #659, opened 2020-07-10
LXL-3230 - Improve work extraction flow (+some refactoring)

oBlissing:
## Checklist:

### Tickets involved
[LXL-3230](https://jira.kb.se/browse/LXL-3230)

### Solves

* Special heading for instanceOf-entities which has big buttons when expanded, and is sticky.
* Preview toggle for the search-window panel.

### Summary of changes

* Added functionality to button-component (a setting which adds bigger button style)
* Added `entity-action` component which is the controls in the heading of entities.
* Refactored part of `field-adder` (and renamed it to `property-adder`) so that it is now just a panel and the button functionality is moved into an `entity-action`


praktwerk (approved):
Great! 👍

PR #688, opened 2020-09-11
Make reverse properties visible on id.kb.se

olovy:
## Description
Make `@reverse` properties visible on id.kb.se
They will be shown as the acual relation, e.g. `@reverse/broader` will be displayed as narrower. 
Both in the html and data views. (Is this good or bad?)

### Tickets involved
[LXL-3291](https://jira.kb.se/browse/LXL-3291)

### Summary of changes
Make the python packend use `_applyInverseOf` when fetching data from backend. (https://github.com/libris/librisxl/pull/722)

### DIFF
--- a/viewer/__init__.py
+++ b/viewer/__init__.py
@@ -230,8 +230,11 @@ def thingview(path, suffix=None):
 
     whelk_accept_header = _get_view_data_accept_header(request, suffix)
     query_params = _filter_query_params(request.args)
-
     resource_id = _get_served_uri(request.url_root, path)
+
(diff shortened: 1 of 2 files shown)


oBlissing (approved):
Go for it 🌠

klngwll (approved):
LGTM! From the users POV it is expected, no idea if this goes against any underlying specs but better working than not so I guess it could be revisited in the future if that is the case. 👍

olovy:
Fixed.

https://github.com/libris/librisxl/commit/62372e264f95cf622436d80aa1c364d1e61fdb7f

PR #690, opened 2020-09-14
LXL-3356-Newly added linked resource does not render as chip until post is saved.

praktwerk:
## Checklist:

## Description

New resource does not render as chip until post is saved.
Caused by a ``null`` value being added to the ``currentValue`` array in ``addLinkedItem``.

### Tickets involved
[LXL-3356](https://jira.kb.se/browse/LXL-3356)

### Summary of changes

Added a null check to prevent a null value being added to the ``currentValue`` array.


oBlissing (approved):
Nice 👍

PR #692, opened 2020-09-14
Search both contribution and instanceOf.contribution for agents

olovy:
## Checklist:

## Description
When showing contributions for an agent, both instance and work documents should be displayed.
In the bottom-right corner of the card and in the corresponding right-panel search.

### Tickets involved
[LXL-3305](https://jira.kb.se/browse/LXL-3305)


oBlissing (approved):
👍

PR #695, opened 2020-09-15
Filter out warnings from rdfxml 'Assertions on rdflib.term.BNode...'

olovy:
I tested this by generating the same warning in `serve.py` and it was filtered correctly.


klngwll (commented):
⭐

PR #698, opened 2020-11-02
Fix summary disappers when extracting work

olovy:
### Tickets involved
[LXL-3378](https://jira.kb.se/browse/LXL-3378)

### Solves
Summary disappearing when extracting work from instance.


oBlissing (approved):
👍

PR #701, opened 2020-11-16
Handle HTTP 400 in Find, i.e. invalid queries.

olovy:
## Checklist:

## Description
Queries with leading wildcards need special handling in backend/ES. These queries might fail, e.g. `*foo "` fails because of unbalanced quotation marks. This change shows a specific error message in this case.

Side panel search is not changed. It shows "No results" on invalid queries.

### Tickets involved
[LXL-3364](https://jira.kb.se/browse/LXL-3364)


oBlissing (approved):
Code looks good. Have not tested (no environment to do that with at the moment).

PR #703, opened 2020-11-20
Add field dropdown to right panel search. Add search option 'Label'.

olovy:
## Checklist:

## Description
Add a "field selection dropdown" to the right panel search. (Free text, Preferred label, ...)

Add a new option "Label" in the search "field selection dropdown" for Concept
The new option searches `prefLabel`, `altLabel`, `hasvariant.prefLabel`.

### Tickets involved
[LXL-3364](https://jira.kb.se/browse/LXL-3364)
[LXL-2586](https://jira.kb.se/browse/LXL-2586)

### Summary of changes
* Make it possible for a "property mapping" to specify multiple fields to be searched.
* Never display "up-link" for fields from the original query (only facets)
* Add option "Label" for Concept
* When clicking "Create/link" on a local entity, the type filter in the search panel has the same type as the local entity selected
* Right panel: Add a "field selection dropdown".
* Right panel: Don't wait 500ms to search after dropdown change (only wait after search phrase change).
* Move function for escaping query string to `HttpUtil` and use it in right panel search also (fixes [LXL-2586](https://jira.kb.se/browse/LXL-2586))

There seems to be a bug in IntelliJ (vue plugin?) that makes it impossible to disable "Strip trailing space on Save" so I finally gave up and commited the changes.

### DIFF
--- a/viewer/vue-client/src/components/inspector/param-select.vue
+++ b/viewer/vue-client/src/components/inspector/param-select.vue
@@ -13,7 +13,7 @@ export default {
     reset: {
       type: Number,
     },
-    contextName: {
+    userPrefKey: {
       type: String,
       default: '',
     },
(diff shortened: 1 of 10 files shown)


viewer/vue-client/src/components/inspector/param-select.vue
```diff
@@ -0,0 +1,115 @@
...
+    <select
+      class="SearchForm-paramSelect SearchForm-select customSelect"
+      v-model="selectedParam"
+      @change="handleChange()"
+      :aria-label="'Choose type' | translatePhrase">
+      <option
+        v-for="prop in availableSearchParams"
+        :key="prop.key"
+        :value="prop">
+        {{prop.key | translatePhrase}}
```
oBlissing, line 105: If the values here are things from the vocab, you can use "labelByLang" instead of "translatePhrase". TranslatePhrase is mainly used to get UI-specific phrases and words translated. labelByLang should be able to handle everything domain-related (aka things present in the vocab).

oBlissing (changes requested):
Adding "Label" to the i18n file is probably not needed. See other note.

viewer/vue-client/src/components/inspector/param-select.vue
oBlissing: And yes, the filter being named `labelByLang` is problematic in itself. But it refers to the label of the class or the prop. It has nothing to do with `...byLang`-properties.

olovy: `translatePhrase` is used in the main search box and is needed because of e.g. "Free text".

PR #707, opened 2020-11-25
Fix: Don't mint main entity URIs with trailing spaces

olovy:
## Checklist:

## Description
All string properties are now trimmed by backend on save but it is frontend that mints id.kb.se URIs from e.g. prefLabel.
Make sure to not create URIs with leading or trailing `%20` in the last path component by trimming the user-supplied string.

### Tickets involved
[LXL-2551](https://jira.kb.se/browse/LXL-2551)


niklasl (approved):
Good catch! 🚢

PR #709, opened 2020-12-02
Fix type selected and input focus when linking empty item

olovy:
## Checklist:

## Description
Clicking "link" on an empty entity would result in an error which leads to
* the search input is not focused
* the correct type is not selected in the search pane

```
TypeError: this.itemInfo is null
--
resetSearch search-window.vue:301
show search-window.vue:274
```

[this is because item-info is set to null if item only contains `@type`](https://github.com/libris/lxlviewer/blob/b74fab2a55f0d5260240d2e2880212f9f9a6200d/viewer/vue-client/src/utils/data.js#L59)

### Tickets involved
[LXL-3447](https://jira.kb.se/browse/LXL-3447)

### DIFF
--- a/viewer/vue-client/src/components/inspector/item-local.vue
+++ b/viewer/vue-client/src/components/inspector/item-local.vue
@@ -563,7 +563,7 @@ export default {
       :all-values-from="allValuesFrom"
       :some-values-from="someValuesFrom"
       :all-search-types="allSearchTypes"
-      :entity-type="entityType" 
+      :entity-type="item['@type']"
       :field-key="fieldKey" 
       :extractable="isExtractable"
       :extracting="extracting" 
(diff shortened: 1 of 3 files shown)


viewer/vue-client/src/components/inspector/item-local.vue
```diff
@@ -563,7 +563,7 @@ export default {
       :all-values-from="allValuesFrom"
       :some-values-from="someValuesFrom"
       :all-search-types="allSearchTypes"
-      :entity-type="entityType" 
+      :entity-type="item['@type']"
```
olovy, line 566: Note that we set `:entity-type="item['@type']"` in the same way on lines 536 and 551 above.

oBlissing (approved):
Ah, right. This stems from "items" being considered "empty" if they only contain `@type`. But in this case we actually need the type so it's fine :)

Code LGTM, haven't tested it in action though.

PR #711, opened 2021-01-07
Fix error when document contains link to itself

olovy:
## Checklist:

## Description
Fixes "the document is missing a reference" error when a document is linking to itself.
* Fix error (datatype should be `entity` not `sibling`)
* Fix chip and link when link points to self (`item-mixin.vue`)

Example (record.inDataset): 
https://libris-dev.kb.se/katalogisering/nrb9xjzsq0rkbtnb

### DIFF
--- a/viewer/vue-client/.editorconfig
+++ b/viewer/vue-client/.editorconfig
@@ -1,7 +1,9 @@
 [*.{js,jsx,ts,tsx,vue}]
+root = true
 indent_style = space
 indent_size = 2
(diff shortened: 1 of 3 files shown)


oBlissing (approved):
As discussed in PM, the only thing that made me hesitate on this was the possiblity of the "sibling"-result never being reached. However, it seems that we don't have any siblings any more (aka graph 2 works), so I say we merge this!

PR #712, opened 2021-01-13
Fix dev/qa/stg indicator blocking tool menu

olovy:
### Solves
dev/qa/stg indicator ("global-msg-corner") overlaps and blocks tool menu on narrow screens.

### Summary of changes
Move it to lower right corner instead of lower left.


oBlissing (approved):
LGTM! 👍

PR #714, opened 2021-02-16
Always show all reverse links in card. Add filters for side panel reverse links view.

olovy:
## Checklist:

## Description
* Show all reverse links ("Used in"/"Används i") for all types. (except Item and Instance, still only holdings).
* Add a dropdown for filtering reverse links list by relation or type (using facets generated by backend).

Depends on: https://github.com/libris/librisxl/pull/854
### Tickets involved
[LXL-3105](https://jira.kb.se/browse/LXL-3105), [LXL-2355](https://jira.kb.se/browse/LXL-2355)

### Solves
All relations were not visible before, e.g. only "contribution.agent" but not "subject" for Agents.
Make the reverse relations result list a little more navigable. It would be nice to be able to sort it but that requires a generic alphabetical sort that we don't have at the moment. (Sorting now requires knowledge of which type(s) the list contains.)

### Summary of changes
* Other changes
    * Remove 1 second delay when we already have the reverse count.
    * Extract facet-mixin

### DIFF
--- a/viewer/vue-client/src/components/mixins/facet-mixin.vue
+++ b/viewer/vue-client/src/components/mixins/facet-mixin.vue
@@ -32,15 +32,23 @@ export default {
         object = object.mainEntity;
       }
       const lang = this.user.settings.language;
-
-      if (object.hasOwnProperty('@id')) {
-        const chains = this.settings.propertyChains;
-        const id = object['@id'];
-        if (chains.hasOwnProperty(id)) {
-          return chains[id][this.user.settings.language];
+      
+      for (const prop of ['@id', '_key']) {
+        if (object.hasOwnProperty(prop)) {
+          const chains = this.settings.propertyChains;
+          const id = object[prop];
+          if (chains.hasOwnProperty(id)) {
+            return chains[id][this.user.settings.language];
+          }
         }
       }
 
+      if (object.hasOwnProperty('propertyChainAxiom')) {
+        return object.propertyChainAxiom
+          .map(o => this.$options.filters.capitalize(this.determineLabel(o)))
+          .join('/');
+      } 
+      
       // TODO: Add chip functionality instead?
       const label = this.getByLang(object, 'prefLabel', lang)
         || this.getByLang(object, 'label', lang)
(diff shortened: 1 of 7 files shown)


viewer/vue-client/src/components/mixins/facet-mixin.vue
```diff
@@ -0,0 +1,85 @@
...
+          if (chains.hasOwnProperty(id)) {
+            return chains[id][this.user.settings.language];
+          }
+        }
+      }
+
+      if (object.hasOwnProperty('propertyChainAxiom')) {
+        return object.propertyChainAxiom
+          .map(o => this.$options.filters.capitalize(this.determineLabel(o)))
+          .join('/');
```
olovy, line 49: How do we want to present this?
Now:
"Metadata/Katalogiseringsregler (446)"
"Seriemedlemskap/Ingår i serie/Instans av Verk (16k)"

andersju (commented):
Looks like there can be a mismatch in the side panel filter between the number of items shown in the "Filtrera" dropdown and the actual number of items? E.g. agent fcrtpljz1qp2bdv (Astrid Lindgren), with the dev example data, shows "Instans (7)" in the dropdown but there are actually 8 items shown if you select that filter. I guess it has to do with the fact that one instance has type Print, while the other seven have type Instance.

olovy:
> Looks like there can be a mismatch in the side panel filter between the number of items shown in the "Filtrera" dropdown and the actual number of items? E.g. agent fcrtpljz1qp2bdv (Astrid Lindgren), with the dev example data, shows "Instans (7)" in the dropdown but there are actually 8 items shown if you select that filter. I guess it has to do with the fact that one instance has type Print, while the other seven have type Instance.

Yes, I forgot to mention that. The count in the aggregations/facets we get from ES only contains the specific type. When doing the search, XL injects all subtypes in the query sent to ES. 

We have the same issue in the "regular" search (look at the type facet):
https://libris-dev.kb.se/katalogisering/search/libris?q=hej%20matematik&%40type=Instance&%40type=Instance&_limit=20&_sort=-meta.modified

viewer/vue-client/src/components/mixins/facet-mixin.vue
andersju: Can't think of anything better at the moment. Maybe best answered by users during testing? So I say we keep it like this for now.

oBlissing:
Not sure this is related at all since I'm not that involved in what these tickets are about, but visually, what properties are visible are controlled in this file

https://github.com/libris/lxlviewer/blob/develop/viewer/vue-client/src/resources/json/displayGroups.json

PR #718, opened 2021-03-04
Linked shelfMark and automatic shelfControlNumber

olovy:
## Checklist:

## Description
Add support for linked shelfMark with automatic generation of shelfControlNumber.

### Tickets involved
[LXL-3465](https://jira.kb.se/browse/LXL-3465)

### Summary of changes
* Allow linking for classes with category `distinct` (`https://id.kb.se/vocab/distinct`)
* Generate `shelfControlNumber` for each `Item` on save if it has a linked `shelfMark` with `nextShelfControlNumber`.
* Add a custom component `shelfControlNumber` which displays the option to generate it if applicable.
* Add template for ShelfMarkSequence (class name may change) in the new template section "Other".
* Limit entity-adder search to active ShelfMarks owned by current sigel

TODO:
* Restrict input `nextShelfControlNumber` to number. (restrict all `xsd:integer`)
* Disallow creating local entities for classes with category `distinct`

### DIFF
--- a/viewer/vue-client/src/components/inspector/item-entity.vue
+++ b/viewer/vue-client/src/components/inspector/item-entity.vue
@@ -47,6 +47,9 @@ export default {
     fullPath() {
       return `${this.parentPath}.{"@id":"${this.item['@id']}"}`;
     },
+    isMaybeMagicShelfMark() {
+      return this.focusData['@type'] === 'ShelfMarkSequence';
+    },
   },
   watch: {
     'inspector.event'(val) {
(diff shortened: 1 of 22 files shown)


viewer/vue-client/src/components/inspector/item-entity.vue
```diff
@@ -99,6 +102,9 @@ export default {
   },
   mounted() {
     this.$nextTick(() => {
+      if (this.isMaybeMagicShelfMark) {
+        this.$root.$emit('maybe-magic-shelf-mark-created', { path: this.actualParentPath, id: this.item['@id'] });
```
oBlissing, line 106: Any specific reason for going via $root instead of using the store?

olovy: This is just the event that signals to the inspector to maybe update the store.
I maybe should break out all the shelfControlNumber methods from inspector so that item-entity can do this by itself.

olovy: Refactored

PR #720, opened 2021-03-09
Feature/paste links

olovy:
## Checklist:

## Description
This fixes the problem when the target of a link is not available to the client leading to a broken chip (and sometimes broken "hover card"). Fixes these cases:
* Pasting an object/entity
* Creating a new doc from a template

Doing a separate HTTP request for each missing thing might be less effective that round-tripping the whole template through embellish? But we don't have an API for that.

Also fixes broken itemOf chip in 'new holding' view.

### Tickets involved
[LXL-3465](https://jira.kb.se/browse/LXL-3465), [LXL-3286](https://jira.kb.se/browse/LXL-3286)


oBlissing (approved):
👍

PR #721, opened 2021-03-11
Add settings.defaultExpandedProperties

olovy:
## Checklist:

## Description
* Add `defaultExpandedProperties` to settings. These properties/fields while be expanded from the start when opening a form.
* Add some default expanded properties
    * `hasComponent`
    * `@reverse/reproductionOf`
    * `@reverse/supplementTo`
    
### Solves
e.g. newly generated `hasComponent.shelfControlNumber` not being visible when saving an `Item`


oBlissing (approved):
Looks good to me! We'll just have to make sure it works in IE11 as well since it doesn't support Array.includes natively, but I think babel solves that anyway.

PR #722, opened 2021-03-11
Don't hide itemOf in relation side panel unless viewing Instance

olovy:
## Checklist:

## Description
itemOf in card is useless if viewing incoming links for an Instance (since itemOf it will be the same Instance).
It is useful if viewing incoming links for other types.


oBlissing (approved):
Sweet 👍

PR #724, opened 2021-03-15
Search: Ranking for property search.

olovy:
## Checklist:

## Description
Fix so that results are ranked by relevancy in Label/Preferred label/Title search

This changes the query from e.g.
`/find?prefLabel=båtar`
to
`/find?prefLabel=båtar&q=båtar`

The first form results in a single `filter` query in ES. [The results will not be ranked and randomly ordered (since filter is only true/false)](https://www.elastic.co/guide/en/elasticsearch/reference/7.11/query-filter-context.html). By adding the free text parameter `q` we also get a `simple_query_string` query in the ES query and relevancy ranking works. (Getting better ranking scores for Concepts is a separate backend task...)

We maybe want to change the logic for ES query building in the `/find` API so that `/find?prefLabel=båtar` gets a "better" ES query with ranking.

### Tickets involved
[LXL-3399](https://jira.kb.se/browse/LXL-3399)

### DIFF
--- a/viewer/vue-client/src/resources/json/i18n.json
+++ b/viewer/vue-client/src/resources/json/i18n.json
@@ -37,6 +37,7 @@
     "Agent": "Agent",
     "Already exists": "Finns redan",
     "Concept": "Koncept",
+    "Subject": "Ämne",
     "Opens in new window": "Öppnas i nytt fönster",
     "The resource has changed": "Resursen har ändrats",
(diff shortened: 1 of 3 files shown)


klngwll (approved):
In tests worked very well with prefLabels.

andersju (approved):
LGTM!

PR #727, opened 2021-03-17
Fix not possible to save digital reproduction with single identifiedBy

olovy:
`identifiedBy` became an object instead of an array with one object which
backend chokes on for this specific property when creating a new doc (HTTP 500).


andersju (approved):
Yep, works fine now! :ship:

PR #734, opened 2021-04-15
Use _sortKeyByLang as sort option in Agent/Concept/All

andersju:
## Checklist:

### Tickets involved
[LXL-3508](https://jira.kb.se/browse/LXL-3508)
### Solves

Previously it wasn't possible to sort results with agents or mixed types alphabetically. This was solved in the backend with two new fields, `_sortKeyByLang.sv` and `_sortKeyByLang.en`, that try to be approximately the same as what would be shown in the header in the UI. This simply makes use of these new fields when sorting.

### Summary of changes

* Concept, Common: use `_sortKeyByLang` instead of `prefLabel` 
* Agent, Other: add `_sortKeyByLang` (A-Z) sort option

`query` in `store.json` is `_sortKeyByLang`, but what actually needs to be sent to the server is `_sortKeyByLang.<lang>`, which depends on the user setting. Right now `user.settings.language` is simply appended in `sort.vue`, with a default of 'sv' if language isn't set. There might very well be a better solution.

### DIFF
--- a/viewer/vue-client/src/components/search/sort.vue
+++ b/viewer/vue-client/src/components/search/sort.vue
@@ -97,7 +97,7 @@ export default {
       @change="handleSortChange">
       <option 
         v-for="(option, index) in options" 
-        :value="option.query"
+        :value="option.query.endsWith('_sortKeyByLang') ? `${option.query}.${user.settings.language || 'sv'}` : option.query"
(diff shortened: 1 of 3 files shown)


olovy (approved):
Yes!
:ship:

PR #735, opened 2021-04-16
Take filter badge label from mapping instead of stats

olovy:
## Checklist:

## Description
Get label for selected facet ("filter badge") from `mappings` instead of `stats`.
Depends on https://github.com/libris/librisxl/pull/894


andersju (approved):
:ship:

PR #738, opened 2021-04-27
Don't choke on types outside known class hierarchies

olovy:
## Checklist:

## Description
Make it possible to edit docs with mainEntity `@type` that is not a subclass of Item/Instance/Work/Agent/Concept. 

We probably want to rework some getRecordType use later.

### Tickets involved
[LXL-3531](https://jira.kb.se/browse/LXL-3531)

### Solves
Used to get an exception on e.g. DataCatalog

### Summary of changes
Main change is in `getRecordType` in vocab.js. The others are just to make it work everywhere.
* `getRecordType` returns the type passed instead of 'Other' if not found among base classes.
* In top of cards: Only show "Unspecified" (e.g. "Unspecified, Work") for Work and Instance  
* Remove sort options for  "Other", use "Common".


andersju (approved):
LGTM!

PR #747, opened 2021-05-25
Display link to login if login expired while editing

olovy:
## Checklist:

## Description
Give the user a chance to log in again if the session has expired when trying to save.
This is done by including a login link in the error message.
The link opens in an new tab, so the user has to login there and then come back to the editor tab. 
Basic solution. It would probably be nicer to stash the edited doc and do the login flow in the same tab.

### Tickets involved
[LXL-2865](https://jira.kb.se/browse/LXL-2865)

Summary of changes
- Add `sticky` mode to notifications. No timeout, only closes on click.
- Add links to notifications.
- Include login link in error notification for 401 on save

### DIFF
--- a/viewer/vue-client/src/components/search/sort.vue
+++ b/viewer/vue-client/src/components/search/sort.vue
@@ -43,7 +43,7 @@ export default {
         return sortOptions;
       }
       if (this.commonSortFallback) {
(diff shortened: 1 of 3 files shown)


andersju (approved):
LGTM!

PR #755, opened 2021-06-07
Sort directory care holding list by card heading, not id/sigel

olovy:
## Checklist:

## Description
Now that we have library names available, sorting the holding list in directory care on `@id` (library URI) gives the wrong order.
It should be sorted on what is displayed in the card heading (library name).

### Tickets involved
[LXL-3617](https://jira.kb.se/browse/LXL-3617)


andersju (approved):
:ship:

PR #757, opened 2021-06-08
Fix ghost data in 'create digital reproduction'

olovy:
## Checklist:

## Description
Data is leaking between different invocations of "Create digital reproduction".

Steps to reproduce:
* Click on "Create digital reproduction" on an instance with `responsibilityStatement`
* Click on "Create digital reproduction" on an instance without `responsibilityStatement`
* `responsibilityStatement` from the first instance will appear in the reproduction

### Tickets involved
[LXL-3480](https://jira.kb.se/browse/LXL-3480)


andersju (approved):
:ship:

PR #760, opened 2021-06-15
Fix bad shelfMarkSequences visible when linking

olovy:
Fix regresion:
ShelfMarkSequences that are inactive and owned by other libraries are
visible in the list when linking shelfMark. Because search parameters
are never set.

## Checklist:

### Tickets involved
[LXL-3628](https://jira.kb.se/browse/LXL-3628)


andersju (approved):
Tested with a couple of holds/sigels/shelfmarksequences and this fix looks good!

PR #768, opened 2021-10-07
Remove the Flask app (API proxy layer, id.kb.se client)

olovy:
## Description
Remove the Flask layer

Last remaining bits of API functionality has been moved to https://github.com/libris/librisxl (https://github.com/libris/librisxl/pull/993)
id.kb.se client has been replaced with https://github.com/libris/id.kb.se

(will merge this after devops procedures and librisxl apache instructions are ready)


olovy:
> Haven't tested this change yet but I think we need to keep `viewer/templates/maintenance.html` around somewhere (it's served directly by Apache)

Good catch!

PR #779, opened 2021-11-15
Update 'create digital reproduction' to new practice

olovy:
## Checklist:

## Description
Change "create digital reproduction" maneuver to use the new simplified form for digital reproductions.

TODO: extract and link work

### Tickets involved
[LXL-3673](https://jira.kb.se/browse/LXL-3673), [LXL-LXL-2807](https://jira.kb.se/browse/LXL-LXL-2807)

### Summary of changes
* Update template
* Only copy title from physical

### DIFF
--- a/vue-client/src/utils/record.js
+++ b/vue-client/src/utils/record.js
@@ -165,38 +165,16 @@ export function getDigitalReproductionObject(original, resources) {
       digitalReproObject.mainEntity.instanceOf = original.mainEntity.instanceOf;
     }
   }
-
-  // Copy the PrimaryPublication if there is one
-  function getPrimaryPublication(mainEntity) {
-    if (mainEntity.hasOwnProperty('publication')) {
-      for (let i = 0; i < mainEntity.publication.length; i++) {
-        if (mainEntity.publication[i]['@type'] === 'PrimaryPublication') {
-          return mainEntity.publication[i];
-        }
-      }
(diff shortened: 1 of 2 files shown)


vue-client/src/utils/record.js
```diff
@@ -165,39 +165,17 @@ export function getDigitalReproductionObject(original, resources) {
...
   ];
   for (let i = 0; i < keysToCopy.length; i++) {
     const originalValue = get(original, keysToCopy[i]);
     if (typeof originalValue !== 'undefined') {
       set(digitalReproObject, keysToCopy[i], originalValue);
     }
   }
-  // Add "indirectly identified by" with the "identified by" value from original
-  if (original.mainEntity.hasOwnProperty('identifiedBy')) {
-    digitalReproObject.mainEntity.indirectlyIdentifiedBy = original.mainEntity.identifiedBy;
```
niklasl, line 199: Are we OK with copying them all? (We may very well be; just double-checking.)

olovy: `keysToCopy` only contains `'mainEntity.hasTitle'` so I should just remove the loop

niklasl: Ah, that's true. I was only thinking about `identifiedBy` though; i.e. copying all of those (may be more than one)...

PR #783, opened 2021-12-07
Add draft for manuscript template

klngwll:
Add template for Instance type Manuscript.

### DIFF
--- a/vue-client/src/resources/json/combinedTemplates.json
+++ b/vue-client/src/resources/json/combinedTemplates.json
@@ -964,6 +964,208 @@
         }
       }
     },
+    "manuscriptdMonographText": {
+      "label": "Handskrift",
+      "description": "Handskrift monografi - RDA",
+      "value": {
+        "record": {
+          "@id": "https://id.kb.se/TEMPID",
+          "@type": "Record",
+          "mainEntity": {
+            "@id": "https://id.kb.se/TEMPID#it"
+          },
+          "bibliography": [],
+          "descriptionConventions": [
+            {
+              "@id": "https://id.kb.se/marc/Isbd"
+            },
+            {
+              "@type": "DescriptionConventions",
+              "code": "rda"
+            }
+          ],
+          "descriptionLanguage": {
+            "@id": "https://id.kb.se/language/swe"
+          },
+          "encodingLevel": "marc:AbbreviatedLevel",
+          "marc:catalogingSource": {
+            "@id": "https://id.kb.se/marc/CooperativeCatalogingProgram"
+          }
+        },
+        "mainEntity": {
+          "@id": "https://id.kb.se/TEMPID#it",
+          "@type": "Manuscript",
+          "issuanceType": "Monograph",
+          "hasTitle": [
+            {
+              "@type": "Title",
+              "mainTitle": "",
+              "subtitle": ""
+            }
+          ],
+          "responsibilityStatement": "",
+          "marc:primaryProvisionActivity": {
+            "@type": "PrimaryProduction",
+            "country": {
+              "@id": "https://id.kb.se/country/sw"
+            },
+            "marc:publicationStatus": "marc:QuestionableDate",
+            "otherYear": "",
+            "year": ""
+          },
+          "production": [
+            {
+              "@type": "Production",
+              "agent": {
+                "@type": "Agent",
+                "label": [
+                  ""
+                ]
+              },
+              "date": [
+                ""
+              ],
+              "place": [
+                {
+                  "@type": "Place",
+                  "label": [
+                    ""
+                  ]
+                }
+              ]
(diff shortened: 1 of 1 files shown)


vue-client/src/resources/json/combinedTemplates.json
```diff
@@ -964,6 +964,200 @@
...
+                "familyName": "",
+                "givenName": "",
+                "lifeSpan": ""
+              },
+              {
+                "@type": "ComplexSubject",
+                "inScheme": {
+                  "@id": "https://id.kb.se/term/sao"
+                },
+                "termComponentList": [
```
niklasl, line 1135: Untimely perhaps, but given that we're working to get rid of at least complex subjects with geographic terms, should we "discourage" them now but not putting them in templates?

vue-client/src/resources/json/combinedTemplates.json
```diff
@@ -964,6 +964,200 @@
...
+                    "@id": "https://id.kb.se/relator/author"
+                  }
+                ]
+              },
+              {
+                "@type": "Contribution",
+                "agent": null,
+                "role": [
+                  {
+                    "@id": "https://id.kb.se/relator/unspecifiedContributor"
```
niklasl, line 1107: Pointless?

vue-client/src/resources/json/combinedTemplates.json
klngwll: I know! If we can land that before or very close to the next release it we should ofc remove it, otherwise it is a big time-saver (unfortunately), my reasoning now is to maybe keep it until we can be more sure.

vue-client/src/resources/json/combinedTemplates.json
```diff
@@ -964,6 +964,200 @@
...
+                  "@type": "Place",
+                  "label": [
+                    ""
+                  ]
+                }
+              ]
+            }
+          ],
+          "carrierType": [{"@id": "https://id.kb.se/term/rda/Volume"}],
+          "baseMaterial": [{"@id": "https://id.kb.se/term/rda/material/Paper"}],
```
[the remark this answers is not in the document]
klngwll: Good catch!

vue-client/src/resources/json/combinedTemplates.json
klngwll: Hmm maybe, I need to investigate what the actual use-case would be to specify unspecified. My suspicion is that it is used when there is only one contributor and it is not the author (which usually has been inferred when missing).

klngwll: There was nothing in the instructions regarding this, so I removed it.

vue-client/src/resources/json/combinedTemplates.json
niklasl: I agree. Would you like to create a  issue/task in the Geo story for removing complex-geo-subjects for _all_ templates?

klngwll: Subtask created ✅

vue-client/src/resources/json/combinedTemplates.json
```diff
@@ -964,6 +964,196 @@
...
+            "@id": "https://id.kb.se/TEMPID#it"
+          },
+          "bibliography": [],
+          "descriptionConventions": [
+            {
+              "@id": "https://id.kb.se/marc/Isbd"
+            },
+            {
+              "@type": "DescriptionConventions",
+              "code": "rda"
```
niklasl, line 984: A link would be good (as discussed); but that is reasonably done in a separate PR for all templates.

klngwll: Yes, let's do that separate.

vue-client/src/resources/json/combinedTemplates.json
```diff
@@ -964,6 +964,196 @@
...
+                    "@id": "https://id.kb.se/relator/author"
+                  }
+                ]
+              },
+              {
+                "@type": "Contribution",
+                "agent": null,
+                "role": [
+                  {
+                    "@id": "https://id.kb.se/relator/author"
```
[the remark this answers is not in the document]
klngwll: Maybe it should be https://id.kb.se/relator/dubiousAuthor ? 😉 
It is not uncommon (lots of aggregated writings I suppose) but it is not more common than scribe, translator etc from what I can see. Let's remove and take the discussion if it pops up. Having too many prefilled values is a always risky.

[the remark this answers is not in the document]
niklasl: Sounds good! Yes, we reasonably lack more powerful editing options for adding, especially linking, things!

niklasl (approved):
LGTM!

PR #784, opened 2021-12-15
Feature/lxl 3643 in scheme as array

oBlissing:
## Checklist:

### Tickets involved
[LXL-3643](https://jira.kb.se/browse/LXL-3643)

### Solves

The UI will handle most values as arrays as often as it can, but limiting them to a length of 1 if the property only takes one value. `uriminter` wasn't able to handle this before this change.

### Summary of changes

* Make uriminter able to handle relations being in arrays


oBlissing:
Closed because a commit that wasn't intended for this PR snuck itself into it.

PR #785, opened 2021-12-15
lxl 3643 inScheme as array

oBlissing:
## Checklist:

### Tickets involved
[LXL-3643](https://jira.kb.se/browse/LXL-3643)

### Solves

The UI will handle most values as arrays as often as it can, but limiting them to a length of 1 if the property only takes one value. `uriminter` wasn't able to handle this before this change.

### Summary of changes

* Make uriminter able to handle relations being in arrays

### DIFF
--- a/vue-client/src/utils/uriminter.js
+++ b/vue-client/src/utils/uriminter.js
@@ -55,7 +55,10 @@ export default class URIMinter {
       }
       const containerMemberMap = containerRelationMap[relation];
       if (mainEntity.hasOwnProperty(relation) && mainEntity[relation] !== null) {
-        const relationId = mainEntity[relation][ID];
+        let relationId = mainEntity[relation][ID]
+        if (Array.isArray(mainEntity[relation])) {
+          relationId = mainEntity[relation][0][ID];
+        }
         const container = containerMemberMap[relationId];
         if (container) {
           if (container.administeredBy.find(it => it[ID] === library[ID])) {
--- a/vue-client/src/utils/uriminter.js
+++ b/vue-client/src/utils/uriminter.js
@@ -49,10 +49,7 @@ export default class URIMinter {
       containerRelationMap = this.containerMap[type];
       if (containerRelationMap) break;
     }
-    for (const relation in containerRelationMap) {
-      if (!containerRelationMap.hasOwnProperty(relation)) {
-        continue;
-      }
+    for (const relation of Object.keys(containerRelationMap)) {
       const containerMemberMap = containerRelationMap[relation];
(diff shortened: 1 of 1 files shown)


vue-client/src/utils/uriminter.js
```diff
@@ -55,7 +55,10 @@ export default class URIMinter {
       }
       const containerMemberMap = containerRelationMap[relation];
       if (mainEntity.hasOwnProperty(relation) && mainEntity[relation] !== null) {
-        const relationId = mainEntity[relation][ID];
+        let relationId = mainEntity[relation][ID];
+        if (Array.isArray(mainEntity[relation])) {
```
niklasl, line 59: I doubt (hope) that these won't have more than one value (it's not correct for a term to have more than one `inScheme` for instance); but if they do, this would be more resilient if it looked through all values to see if there is a matching `container`.

oBlissing: Very good point. Fixing!

oBlissing: Thinking about this I'm wondering how multiple containers should be handled. Wouldn't it make more sense to check if _all_ the entered containers matches the current sigel? Otherwise you could potentially add random containers as long as one of them is one that you are admin for.

niklasl: Good point! I think the semantics of membership in containers ([at least for us](https://github.com/libris/definitions/blob/develop/source/containers.ttl)) ought to be that something is off (even wrong) if an entity qualifies for membership in more than one container. In theory, we could keep old containers around which have minted resources according to one set of rules, but then only one container still ought to be allowed to define the URI of a new entity (which is what we use them for).

This hasn't happened yet, but I'd be inclined to extend this code later on if we device new  containers for the same scheme. Probably, the old ones would be marked with `deprecated: true` and/or `isReplacecBy` with a link to the new container, and then let this code, if need be, mint `sameAs`-URI:s using the replaced containers. But this is just theory right now.

Therefore, this code should throw an explicit (and informative) error if there is more than one container that qualifies here. Something like:
```
new Error(`Got multiple container matches: ${containerCandidates.map({it => it[ID]})join(', ')} for ${mainEntity[ID]}`)
```

oBlissing: As I see it; it should solve two problems:

* If several containers are added, use the one that matches for the URI, but keep them all in the data. This is something that probably _could_ happen.
* In the _unlikely_ event that more than one matches, throw an error and don't proceed.

olovy:
Rebased on develop

PR #793, opened 2022-03-11
Feature/flexible context handling

niklasl:
## Checklist:
- [x] I have run the unit tests. `npm test`

### Tickets involved
[LXL-3740](https://jira.kb.se/browse/LXL-3740)

### Solves

Fixes a hard dependency upon an idiomatic way of using a context array with prefixes in the first object and term definitions in the second.

Now using a simple JSON-LD context with terms and prefixes mixed will work as well (which will be the case when the KBV context is used as default).

### Summary of changes

Sets up the internal structure in `preprocessContext`.


olovy (approved):
:ship:

PR #798, opened 2022-05-11
Remove API_PATH from nuxt-app

olovy:
## Description
Remove API_PATH from nuxt-app. Instead just fetch resources based on the domain we are serving.

Still TODO: 
Remove all usages of "defaultHostPath", i.e:
* make sure all required static resources can be loaded from both libris and id
* make layout switching and index page more generic

### Tickets involved
[LXL-3738](https://jira.kb.se/browse/LXL-3738)


olovy:
vocab should always be accessed as https://id.kb.se...

PR #799, opened 2022-05-18
Bugfix/nuxt redirects

olovy:
## Description
* When nuxt-app gets a redirect from the backend for a resource it should also send a redirect to the browser.

* Download links should be constructed from record id not thing id
  * ie https://libris.kb.se/86lnm8ns2nd7c62/data.jsonld, not https://id.kb.se/term/sao/H%C3%A4star/data.jsonld

Mixed bag of other fixes (see individual commits):
* Error message if XL_SITE_CONFIG is not defined

* Make cat client work on kblocalhost.kb.se:5000/katalogisering

* Handle errors when fetching vocab

* Handle click on nuxt link with redirects

* Always render sameAs links as normal links (not nuxt links)

* Fetch vocab stuff in parallel (no big improvement)

* Make webpack dev server host configurable

### DIFF
--- a/nuxt-app/components/EntityNode.vue
+++ b/nuxt-app/components/EntityNode.vue
@@ -5,7 +5,7 @@
         {{ getItemLabel }}
       </span>
       <template v-else-if="entityData['@id']">
-        <template v-if="isInternalUri(entityData['@id'])">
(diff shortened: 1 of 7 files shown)


niklasl (approved):
Good stuff! :ship:

PR #800, opened 2022-05-20
Don't accidentally translate sigel; fix lint error

andersju:
## Checklist:

### Tickets involved
[LXL-3863](https://jira.kb.se/browse/LXL-3863)

### Solves

The sigel "No" would accidentally be translated to "Nej", because e.g. "No" and "has holding" were passed individually to the translation method, which then looked up and found "No" in the i18n file.

### Summary of changes

Don't pass the sigel to the translation method. Also covers translation of aria-label (in this case it passes label to rounded-button, which then runs the label through translatePhrase; but since we now pre-translate the string, translatePhrase (actually StringUtil.getUiPhraseByLang) doesn't find it in i18n.json and so returns it untouched. Seemed like the least messy solution)

### DIFF
--- a/vue-client/src/components/inspector/create-item-button.vue
+++ b/vue-client/src/components/inspector/create-item-button.vue
@@ -84,8 +84,10 @@ export default {
       'status',
     ]),
     tooltipText() {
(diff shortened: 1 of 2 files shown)


olovy:
:ship:

PR #801, opened 2022-05-23
Make terms with & in prefLabel be searchable and not 404

andersju:
## Checklist:

### Tickets involved
[LXL-3787](https://jira.kb.se/browse/LXL-3787)

### Solves

Terms with '&' would 404, and would also not be searchable:

* https://id.kb.se/term/saogf/Rhythm%20%26%20blues
* https://id.kb.se/find?q=Rhythm%20%26%20blues

### Summary of changes

I noticed that this worked: `curl -H "Accept: application/ld+json" "http://id.kblocalhost.kb.se:5000/https://id.kb.se/term/saogf/Rhythm%20%26%20blues"`

But this didn't: `curl -H "http://id.kblocalhost.kb.se:5000/https://id.kb.se/term/saogf/Rhythm%20%26%20blues"`

The first one is caught by `ProxyPassMatch ^/(http.*)$ http://localhost:8180/$1 nocanon`;, resulting in the request (from Apache to the backend) `GET /https://id.kb.se/term/saogf/Rhythm%20%26%20blues HTTP/1.1`.

The second one is caught by `RewriteRule ^/(.*)$ http://localhost:3000/$1 [P,L]`, resulting in the request (from Apache to nuxt-app) `GET /https:/id.kb.se/term/saogf/Rhythm%20&%20blues HTTP/1.1`

`%26` was turned into `&` (before even reaching Nuxt).

That part could be solved by changing the RewriteRule to something like
```
    RewriteCond %{THE_REQUEST} ^\S+\s+\/(.*)\s
    RewriteRule ^/(.*)$ http://localhost:3000/%1 [P,NE,L]
```
...but Nuxt, somewhere deep down, would decode the `%26` anyway before making the request to the XL backend.

So, instead: just replace `&` with `%26` where appropriate -- we already did this for `(` and `)`. Seems to work (individual pages and searching) and doesn't seem to break anything :shrug:

### DIFF
--- a/nuxt-app/components/SearchBar.vue
+++ b/nuxt-app/components/SearchBar.vue
@@ -29,6 +29,7 @@
 // import mockSuggest from '@/resources/json/mockSuggest.json';
 import { mapGetters } from 'vuex';
 import SuggestItem from '@/components/SuggestItem';
+import { encodeSpecialChars } from '../plugins/env';
 
 export default {
   data() {
@@ -111,7 +112,7 @@ export default {
     },
     async doSuggestSearch() {
       this.clearSuggest();
-      const searchPath = `find.jsonld?q=${this.keyword}&_lens=chips&_suggest=${this.settings.language}&_limit=7`;
+      const searchPath = `find.jsonld?q=${encodeSpecialChars(this.keyword)}&_lens=chips&_suggest=${this.settings.language}&_limit=7`;
       const suggestData = await fetch(`${this.activeHost()}/${searchPath}`).then(res => res.json());
       this.suggestedItems = suggestData.items;
       // const suggestData = mockSuggest;
--- a/nuxt-app/pages/_.vue
+++ b/nuxt-app/pages/_.vue
@@ -33,6 +33,7 @@ import LensMixin from '@/mixins/lens';
 import ResultItem from '@/components/ResultItem';
 
 import { translateAliasedUri } from '../plugins/env';
+import { encodeSpecialChars } from '../plugins/env';
 
 export default {
   mixins: [LensMixin],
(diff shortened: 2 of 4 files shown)


niklasl:
Good catch! In principle, we ought to only do this (replace `&` with `%26` in the path portion) _before_ any eventual queryString (i.e. after an `?`). (Technically, this is not safe either, as sub-delims are allowed in any segments, and thus differ from encoded forms (as opaque parts of the segments). But that's so rare and undesirable I don't mind us conflating them in the path space.)

In practise, I also don't think we'll ever use query strings in formal identifiers. It is unclear though, whether we might want to directly mint and pass such URI:s down to the API from the user interface without this rewriting happens though. One case might be rendering HTML for (conneg:able) search views... (Which did work in the Python/Flask-based UI layer.)  I _suppose_ we'd notice quickly enough if this fix would interfere with that though, but there's a risk of us adding a workaround to a workaround... :thinking:

andersju:
@niklasl Query strings should still work as before with this workaround - it basically does two things:

1) For individual records, e.g. http://id.kblocalhost.kb.se:5000/https://id.kb.se/term/saogf/Rhythm%20%26%20blues it ensures `&` is replaced with `%26` in the path component - the query component is not affected. So you can still do http://id.kblocalhost.kb.se:5000/https://id.kb.se/term/saogf/Rhythm%20%26%20blues?foo=bar&baz=woo, and Nuxt will see the query params (...but will anyway proceed to throw them away when talking to XL - only the path is preserved, currently)
2) For /find, it replaces `&` with `%26` in any query parameter _value_

Since & _is_ legal in the path component, this is obviouly problematic if we do have (or will have) IDs with `&` rather than `%26` (but then again we already make Nuxt forcibly replace `(` and `)` with their encoded versions before passing them on to XL, so same problem there). I guess the _best_ solution would be to have Apache pass the raw request to Nuxt, and stop Nuxt from doing things to it, but the latter seems like a black hole of pain (from my not-at-all-experienced-with-these-JS-things POV - YMMV :smile:).

niklasl:
@andersju Excellent checkup! That settles my worries given the current state of things!

niklasl (approved):
LGTM! :ship:

olovy:
Great! :ship:

PR #802, opened 2022-05-24
Add publicNote and cataloguersNote to associatedMedia

klngwll:
Return properties for digi-workflow.

[LXL-3784](https://jira.kb.se/browse/LXL-3784)


olovy (approved):
:ship:

PR #806, opened 2022-07-01
Feature/version history

lrosenstrom:
## Checklist:

## Description

View change history for a specific record.

* Basic history view where each version of the document can be inspected.
* Mark added / removed / modified (roughly) according to UX sketches. 
* Automatically expand all changed paths when selecting a specific version.
* 'Blame' functionality: clicking a field label navigates to the version where this property was last changed.

TODO:
* Icons for added/removed/modified for clarity/DOS?
* Fix tabbing
* Auto scroll to selected version when clicking a field label.
* Use same 'maskinellt' / 'icke-maskinellt' logic as in 'sammanslagning av poster'?

Depends on `origin/feature/fix-history2` for the backend.


vue-client/src/components/inspector/item-value.vue
```diff
@@ -173,7 +173,13 @@ export default {
 
 <template>
   <div class="ItemValue js-value" 
-    v-bind:class="{'is-locked': isLocked, 'unlocked': !isLocked, 'is-removed': removed}"
+    v-bind:class="{
+    'is-locked': isLocked,
+    'unlocked': !isLocked,
+    'is-removed': removed,
+    'is-diff-removed': diffRemoved && !diffAdded,
+    'is-diff-added': diffAdded && !diffRemoved,
```
olovy, line 181:
```suggestion
    'is-diff-removed': diffRemoved && !diffModified,
    'is-diff-added': diffAdded && !diffModified,
```

vue-client/src/components/inspector/version-history-changesets.vue
```diff
@@ -0,0 +1,88 @@
...
+  mounted() {
+  },
+};
+</script>
+
+<template>
+  <div class="VersionHistory-changeSets" v-if="changeSetsReversed">
+    <div class="ChangeSet" v-for="(changeSet, index) in changeSetsReversed" :key="changeSet.date" @click="selectVersion(index)" :class="{ 'selected': selectedVersion == index }">
+      <div class="ChangeSet-changeSetContainer" :class="{ 'selected': selectedVersion == index }" ref="changeSets">
+        <span class="ChangeSet-currentVersion" :class="{ 'selected': selectedVersion == index }" v-if="index == 0">Aktuell version</span>
```
olovy, line 70: i18n

olovy (changes requested):
Good work!

Things I think need to be fixed before merging this:
- Window title is "DocumentHistory"
- "Enum" properties are not translated
  - e.g "Collection → Serial" is displayed instead of "Samling -> Seriell resurs"

Things that can be improved after merging:
- The version list on the right side doesn't scroll all the way down if the window isn't tall enough.
- Version list isn't visible in mobile view
- "Show latest change"
  - I think it should show the latest _previous_ change from the currently displayed version? Now it always shows the latest change of all versions.
  - Clicking on a property that hasn't changed since the first version doesn't work.
- Try to make the switching between version more instant? Now changed values and highlights appear after a slight delay.
- It seems that linking straight to the history with `/history` in the URI doesn't work? I end up in the regular inspector view.
- I did a quick fix for modified links in non-array fields to be able to continue testing. See 8f86eb67ceba7e6054da302e2c1e2b844e127025. Display of these could be improved.

olovy (approved):
:ship: :dash: :exclamation:

PR #809, opened 2022-08-09
Feature/lxl 3885 settings email change

olovy:
## Checklist:

## Description
Use persistent user id as key for user settings. For both server side settings and local storage.
So that settings are not lost when user changes email address.

Depends on
https://github.com/libris/xl_auth/pull/250
https://github.com/libris/librisxl/pull/1136

### Tickets involved
[LXL-3883](https://jira.kb.se/browse/LXL-3883)
### Solves


andersju (approved):
Nice! LGTM!

andersju:
Actually, one thing - probably `"sjcl": "^1.0.8"` should be in `lxljs/package.json` rather than `vue-client/package.json`? Otherwise nuxt-app will fail because it can't find sjcl (since nuxt-app also depends on lxljs). `lxljs/yarn.lock` then also needs to be updated.

olovy:
> Actually, one thing - probably `"sjcl": "^1.0.8"` should be in `lxljs/package.json` rather than `vue-client/package.json`? Otherwise nuxt-app will fail because it can't find sjcl (since nuxt-app also depends on lxljs). `lxljs/yarn.lock` then also needs to be updated.

Yes! Good catch.

PR #815, opened 2022-08-18
Make error pages more user-friendly

andersju:
## Checklist:

### Tickets involved
[LXL-3887](https://jira.kb.se/browse/LXL-3887)

### Solves

Error pages (e.g. 500) should not expose internal details to visitors and shouldn't use technical jargon.

### Summary of changes

Clear headings (e.g. "There is a problem with the service" rather than "500"), possibility to have additional (and translatable) information, proper <title> (e.g. "Page not found | id.kb.se" rather than "id.kb.se").


lrosenstrom:
Sounds reasonable!

lrosenstrom (approved):
Sounds reasonable!

PR #816, opened 2022-08-22
Better error message when holding already exists

andersju:
## Checklist:

### Tickets involved
[LXL-2078](https://jira.kb.se/browse/LXL-2078)

### Solves

"Something went wrong" is not a useful error when trying to save a holding that already exists for the current sigel.

### Summary of changes

Adds a human-friendly error message if status 409 is encountered when saving.


olovy (approved):
:ship: :dash:

PR #818, opened 2022-08-30
Closeable fullscreen version list view on window shrink

lrosenstrom:
* Improve flex behavior
* Make version list go fullscreen on smaller window sizes


olovy (approved):
LGTM!
:ship:

vue-client/src/components/inspector/version-history.vue
```diff
@@ -58,6 +59,9 @@ export default {
       }
       return null;
     },
+    hideSideCol() {
```
olovy, line 62: Maybe `isSideColHidden`?

lrosenstrom: :100:

PR #819, opened 2022-09-01
Feature/handle old works

lrosenstrom:
Modify the backend response to be able to show it in a uniform way when the work is at graph[2]. 

Tested with a mocked backend response that corresponds to the old form:
```
          // Fake backend response ============================
          const instance = get(result, ['@graph', 1, 'instanceOf']);
          const id = get(result, ['@graph', 0, '@id']);
          instance['@id'] = id.concat('#work');
          set(result, ['@graph', 2], instance);
          set(result, ['@graph', 1, 'instanceOf'], { '@id': instance['@id'] });
          console.log('FAKE BACKEND RESPONSE:', JSON.stringify(result));
          // ==================================================
```

### DIFF
--- a/vue-client/src/utils/data.js
+++ b/vue-client/src/utils/data.js
@@ -164,6 +164,15 @@ export async function fetchMissingLinkedToQuoted(obj, store) {
     .catch(e => console.log(e));
 }
 
+export function moveWorkToInstance(data) {
+  const oldWork = data.work;
+  if (oldWork !== undefined) {
+    delete oldWork['@id']; // Get rid of <record-id>#work
+    data.mainEntity.instanceOf = oldWork;
+  }
+  return data;
+}
+
 // Changes XML to JSON
 // Modified version from here: http://davidwalsh.name/convert-xml-json
 export function xmlToJson(xml) {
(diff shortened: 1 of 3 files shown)


vue-client/src/utils/data.js
```diff
@@ -164,6 +164,15 @@ export async function fetchMissingLinkedToQuoted(obj, store) {
     .catch(e => console.log(e));
 }
 
+export function moveWorkToInstance(data) {
+  const oldWork = data.work;
+  if (oldWork !== undefined) {
+    delete oldWork['@id']; // Get rid of <record-id>#work
```
niklasl, line 170: Makes sense (and is "correct"); I just wonder if we want some sort of cue to show that something "happened" to the work entity (that is was embedded). We've only done this once though, so building view logic for that is arguably overkill.

niklasl (approved):
LGTM!

vue-client/src/utils/data.js
lrosenstrom: Yes, I'm not 100% sure how the diff will look for the version where embedding took place. Probably it will look like all items in instanceOf were newly added. We can take another iteration for that specific case if it's unclear/misleading.

PR #820, opened 2022-10-06
Cache vocab resources in nuxt-app

olovy:
Only fetch vocab resources once in nuxt-app and not for every (server-side rendered) request.

A further improvement would to refactor VocabUtil so that more of the preprocessing that is now done in e.g. SET_VOCAB_PROPERTIES could be done when fetching vocab.


niklasl (approved):
Excellent! :ship:

PR #822, opened 2022-10-18
Use WOFF/WOFF2 instead of TTF for fonts

andersju:
Just because this has mildly annoyed me for some time. WOFF2 [is supported](https://caniuse.com/woff2) by every modern browser. WOFF is even [supported](https://caniuse.com/woff) by ancient browsers like IE 9 (2011). 50-60% size reduction. TTF is a waste of bandwidth here.

Sourced from https://github.com/FontFaceKit/open-sans and https://github.com/FontFaceKit/roboto. I verified that they support the same languages as the TTFs we used before. (They also include some updates that we'd missed out on.)

Uncompressed font transfer size before and after:

id.kb.se, TTF: 1.82 MB
id.kb.se, WOFF2: 833 KB

lxlviewer, TTF: 370 KB
lxlviewer, WOFF2: 201 KB

Also removed some FontAwesome files that we no longer use.


olovy (approved):
Nice!

:ship:

PR #825, opened 2022-11-29
Show subclasses and -properties in the vocab view

niklasl:
Done by preprocessing the vocab data upon initial fetch.

Needs more testing:

- [x] Run the unit tests. `yarn test:unit`
- [x] Run the linter. `yarn lint`
- [x] Test the application more (having `baseClassOf` around (again?) should not interfere with the cataloguing client, but that must be verified)


olovy (approved):
Finally!
:ship:

PR #833, opened 2023-01-09
Feature/lxl 3956 runtime vars

henrikb77:
## Checklist:
- [x ] I have run the unit tests. `yarn test:unit`
- [x ] I have run the linter. `yarn lint`

## Description
As a pre-req for LXL-3956, make vue-client environment-agnostic, by injecting runtime-config at deploy-time. Hence artifact can be built once for any environment.

### Tickets involved
[LXL-3956](https://jira.kb.se/browse/LXL-3956)

### Remarks

Open for comments. There are some doubts I will comment near the changes. And maybe especially I'm concerned for how long runtimeConfig.js will be cached. Although changes will likely be rare per environment it must be possible to get clients to update the config.

### Depends on
https://github.com/libris/devops/pull/146


vue-client/public/index.html
```diff
@@ -10,11 +10,12 @@
     <meta property="og:title" content="Libris katalogisering">
     <meta property="og:site_name" content="Libris katalogisering">
     <meta property="og:description" content="Cataloguing interface for the Swedish National Union Catalogue">
-    <meta property="og:image" content="https://libris<%= process.env.VUE_APP_ENV_LABEL ? '-' : '' %><%= process.env.VUE_APP_ENV_LABEL %>.kb.se<%= BASE_URL %>opengraph.png">
+    <meta property="og:image" content="https://libris.kb.se<%= BASE_URL %>opengraph.png">
```
henrikb77, line 13: With these changes, this value cannot be known at build-time/template processing time.
Don't know how crucial it is to have the right url pointed out from test ends here. With current implementation it was sacrificed.

olovy:
LGTM!

> And maybe especially I'm concerned for how long runtimeConfig.js will be cached. 

The index page is never cached and it always points to the updated js-file versions so that shouldn't be a problem.

olovy (approved):
:ship:

PR #835, opened 2023-01-13
Bugs/lxl 4036 fix resource history route

didair:
## Checklist:

## Description
Fixes the issue with user navigating to resource history view directly and page redirecting unexpectedly.

### Tickets involved
[LXL-4036](https://jira.kb.se/browse/LXL-4036)

### Solves
Mounting Inspector view in history view

### Summary of changes
* Add support for "DocumentHistory" view when initializing resource inspector view
* Add null check for missing refs in App component
* Small formatting in documents

### DIFF
--- a/vue-client/src/App.vue
+++ b/vue-client/src/App.vue
@@ -4,23 +4,33 @@
     <EnvironmentBanner />
     <navbar-component />
     <search-bar v-if="resourcesLoaded" :class="{ 'stick-to-top': stickToTop }" />
+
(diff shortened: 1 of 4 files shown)


olovy (approved):
LGTM!

Note that the history button is still broken if the resource is accessed with an extra slash at the end. But that can be handled as a separate fix.

Example
https://libris.kb.se/katalogisering/3ld0l6kf0dk7tzv/
vs
https://libris.kb.se/katalogisering/3ld0l6kf0dk7tzv

didair:
> LGTM!
> 
> Note that the history button is still broken if the resource is accessed with an extra slash at the end. But that can be handled as a separate fix.
> 
> Example https://libris.kb.se/katalogisering/3ld0l6kf0dk7tzv/ vs https://libris.kb.se/katalogisering/3ld0l6kf0dk7tzv

I missed this when making my original change but I have now corrected this by adding a middleware in router.js that removes any trailing slashes from the url when the user navigates somewhere.

olovy (approved):
:ship:

PR #838, opened 2023-01-18
fix .env file location

henrikb77:
## Checklist:
- [ x] I have run unit tests. `yarn test`

## Description

Noticed a regression currently present on id-dev.kb.se (after moving all source to src/).
Console log says: "[Error] env.XL_SITE_CONFIG not defined (fix for development: copy .env.in to .env)"

By pointing out location of .env the problem seems to be fixed.


andersju (approved):
LGTM!

PR #839, opened 2023-01-20
Feature/lxl 4035 make history work for lang containers

lrosenstrom:
## Summary of changes
* Adds support for showing changes in language containers in the history view. 
* Refactoring of item-byLang.vue to make it more logically consistent, and easier to handle per-entry properties.
* Better caching.

### Tickets involved
[LXL-4035](https://jira.kb.se/browse/LXL-4035)


didair (approved):
LGTM 👍

PR #843, opened 2023-01-24
Don't fail when a repeatable property isn't an array

lrosenstrom:
## Checklist:
- [ x] I have run the unit tests. `yarn test:unit`
- [ x] I have run the linter. `yarn lint`

## Description
During QA-testing some users noticed that it language tags cannot be added in some cases, e.g. in PrimaryPublication->Place on https://libris-qa.kb.se/ctr9t1f798vdw4r 

### Tickets involved
[LXL-4009](https://jira.kb.se/browse/LXL-4009)

### Solves
No longer fail when a repeatable property isn't an array.

### Summary of changes

Replace check for repeatability with check if we are dealing with an array.


didair (approved):
LGTM 👍

PR #844, opened 2023-02-03
Make it possible to inject env variables at runtime

henrikb77:
## Checklist:
- [x ] I have run the unit tests. `yarn test`

## Description
*DEPENDS ON:* https://github.com/libris/devops/pull/154

### Tickets involved
[LXL-3956](https://jira.kb.se/browse/LXL-3956)

### Summary of changes

Moved build-time state from env.js. Instead inject functions and make use of publicRuntimeConfig, which can be retrieved in different ways depending of the context.

### DIFF
--- a/nuxt-app/nuxt.config.js
+++ b/nuxt-app/nuxt.config.js
@@ -1,8 +1,6 @@
 const { gitDescribeSync } = require('git-describe');
-import { defaultHostPath } from './src/plugins/env';
-const HOST_PATH = defaultHostPath();
-process.env.APP_VERSION = require('./package.json').version;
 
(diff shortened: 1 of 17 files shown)


olovy (approved):
LGTM!

niklasl (approved):
LGTM!

(Nitpick: `encodeSpecialChars` isn't (and should never be) application state dependent. It "should" be moved to where utility functions live instead. This change doesn't need to address that, and it could just leave it in `env.js` to avoid making it appear to be state-dependent. But there are probably more things in this category and could be addressed in another cleanup.)

PR #848, opened 2023-02-20
Feature/lxl 3725 lock next shelf control number

didair:
## Checklist:

## Description
Integrates same lock as in item-type in new next-shelf-control-number item.

### Tickets involved
[LXL-3725](https://jira.kb.se/browse/LXL-3725)


olovy (approved):
LGTM.

Let's fix that not all strings in lxlviewer have i18n separately.
The next time we add a lock we can refactor it to a general thing that can be added to all items.

PR #849, opened 2023-02-20
Check for conflicting ids before adding to path object in version-history

didair:
## Checklist:

### Description
Makes sure no duplicate data is displayed in history view when e.g. one genre is added and others already exists.

### Tickets involved
[LXL-3991](https://jira.kb.se/browse/LXL-3991)


lrosenstrom (approved):
:ship:

PR #850, opened 2023-02-27
Fix bug where expand-icon shows up in entity-header for non-expandabl…

lrosenstrom:
…e properties

See [LXL-4063](https://jira.kb.se/browse/LXL-4063)


didair (approved):
👍

PR #851, opened 2023-03-01
Allow identical language tags in repeatable properties

lrosenstrom:
## Checklist:

## Description
See [LXL-4005](https://jira.kb.se/browse/LXL-4005)


olovy (approved):
LGTM.

Did a quick test, seems to work fine.
One consequence is that pressing the romanization button for these fields always adds a new field but I guess that's fine.

PR #857, opened 2023-03-16
Cache bust runtimeConfig.js using app version

niklasl:
## Description

VUE_APP_VERSION contains the git tag, set via vue.config.js and injected in dist/index.html at build time. The index file has no cache set by the server, but we use this pattern since most of the built files have similar (albeit hash-based) cache-busting URL:s.

## Checklist:

### Solves
Risk of using browser-cached and stale runtime config.

### Summary of changes
Add GET-parameter with app version (ignored by apache but considered as a different URL by browsers).


andersju (approved):
LGTM!

PR #860, opened 2023-03-20
Back button -> navigate to record

lrosenstrom:
## Checklist:

### Tickets involved
[LXL-4079](https://jira.kb.se/browse/LXL-4079)

### Solves

The back button in the history view navigates the user back to the previous page visited. This is a problem when going to the history view by directly specifying the path with `/history`.
### Summary of changes

Summary of changes
The back-button always leads back to the record to which the history is associated.


olovy (approved):
LGTM!

PR #864, opened 2023-04-06
Change instruction video links from youtube to KB Play

olovy:
It is not possible to embed the whole playlist on KB Play, only individual videos. So I removed that part from the card.
Landing page is now one row with three cards and one row with one card of the same size.

I have not ripped out all the youtube embedding code in `link-card.vue`

## Checklist:

### Tickets involved
[LXL-4096](https://jira.kb.se/browse/LXL-4096)


didair (approved):
LGTM 👍

PR #867, opened 2023-05-17
Feature/lxl 3723 show domain and range includes

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-3723](https://jira.kb.se/browse/LXL-3723)

### Solves

Shows properties that can occur on classes using `domain`, `domainIncludes`, `range` and `rangeIncludes`. 

### Summary of changes

- Use maps instead of arrays of objects for `vocabClasses`and `vocabProperties` in the vuex store.
- Show allowed properties on classes.
- Remove repetition of domain/range list methods (`getDomainList` and `getRangeList` is replaced with `getPropertiesList`).

### DIFF
--- a/lxljs/vocab.js
+++ b/lxljs/vocab.js
@@ -450,8 +450,9 @@ export function getLinkedProperties(linkType, classId, vocabClasses, vocabProper
     return [];
   }
 
-  if (termObj[`${linkType}Properties`]) {
-    return termObj[`${linkType}Properties`];
+  const capitilzedLinkType = linkType[0].toUpperCase() + linkType.slice(1);
+  if (termObj[`in${capitilzedLinkType}Of`]) {
+    return termObj[`in${capitilzedLinkType}Of`];
   }
 
   const linkedProperties = Array.from(vocabProperties.values())
--- a/lxljs/vocab.js
+++ b/lxljs/vocab.js
@@ -441,10 +441,33 @@ export function getDomainList(property, vocab, context) {
   return domainList.map(item => StringUtil.getCompactUri(item, context));
 }
 
+export function getRangeList(property, vocab, context) {
+  if (property['@type'] === 'Class') {
+    return false;
+  }
+  let rangeList = [];
+  const vocabPfx = context[0]['@vocab'];
+  if (property.hasOwnProperty('range')) {
+    rangeList = rangeList.concat(property.range.map(obj => obj['@id']));
+  }
+  if (property.hasOwnProperty('rangeIncludes')) {
+    rangeList = rangeList.concat(property.rangeIncludes.map(obj => obj['@id']));
+  }
+  if (property.hasOwnProperty('subPropertyOf') && rangeList.length === 0) {
+    for (const superPropNode of property.subPropertyOf) {
+      if (superPropNode['@id'] && superPropNode['@id'].indexOf(vocabPfx) !== -1) {
+        const superProp = getTermObject(superPropNode['@id'], vocab, context);
+        if (superProp) {
+          rangeList = rangeList.concat(getRangeList(superProp, vocab, context));
+        }
+      }
+    }
+  }
+  return rangeList.map(item => StringUtil.getCompactUri(item, context));
+}
+
 export function getProperties(classId, vocabClasses, vocabProperties, context) {
-  // Get all properties which has the domain of the className
-  const props = [];
-  // console.log("Getting props for", className);
+  // Get all properties which has the domain or range of the className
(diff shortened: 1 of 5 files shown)


lxljs/vocab.js
```diff
@@ -453,22 +450,16 @@ export function getProperties(classId, vocabClasses, vocabProperties, context) {
...
+      if (getPropertiesList(prop, vocabProperties, context).includes(classId)) {
+        return true;
       }
-    }
-  });
-  termObj.allowedProperties = props;
-  return props;
+      return false;
+    });
+  termObj.allowedProperties = allowedProperties;
```
johanbissemattsson, line 461: Should `termObj` be mutated here? There seems to be several cases of mutations which makes it difficult to trace the data flow.

nuxt-app/src/store/index.js
```diff
@@ -185,21 +186,42 @@ export const mutations = {
...
     props = props.concat(VocabUtil.getTermByType('Property', vocabMap, state.vocabContext, state.settings));
     props = props.concat(VocabUtil.getTermByType('DatatypeProperty', vocabMap, state.vocabContext, state.settings));
     props = props.concat(VocabUtil.getTermByType('ObjectProperty', vocabMap, state.vocabContext, state.settings));
     props = props.concat(VocabUtil.getTermByType('owl:SymmetricProperty', vocabMap, state.vocabContext, state.settings));
-    state.vocabProperties = props;
+    const vocabProperties = new Map(props.map(entry => [entry['@id'], entry]));
+
+    state.vocabClasses.forEach((classObj) => {
+      const compactClassUri = StringUtil.getCompactUri(classObj['@id'], state.vocabContext)
+      const allowedPropertiesInClass = VocabUtil.getProperties(compactClassUri, state.vocabClasses, vocabProperties, state.vocabContext)
```
johanbissemattsson, line 222: The returned value from `const allowedPropertiesInClass = VocabUtil.getProperties(...` isn't actually used here but `getProperties` adds `allowedProperties` to the classes in the vuex-store (due to mutation). I'm thinking I should restructure this in some way but need to discuss it...

lxljs/vocab.js
niklasl: Agreed; let's discuss improvements here!

lxljs/vocab.js
```diff
@@ -453,22 +450,16 @@ export function getProperties(classId, vocabClasses, vocabProperties, context) {
...
-        getSubClassChain(domainList[i], vocabClasses, context),
-      );
-    }
-    for (const domain of domainListWithSubClasses) {
-      if (domain === classId) {
-        props.push(prop);
+
+  const allowedProperties = Array.from(vocabProperties.values())
+    .filter((prop) => {
+      if (getPropertiesList(prop, vocabProperties, context).includes(classId)) {
```
niklasl, line 456: This appears to result in the set of "allowed properties for this class" including properties who have this class as their range (which would be backwards)? Am I missing something?

johanbissemattsson: > 

Skärmavbild 2023-05-24 kl  09 03 26
Should the properties listed under _Occurs in_ in the screenshot be separated into two separate table rows (one for "Förekommande egenskaper" and one for "Förekommer i egenskaper")? Would that be correct?

lxljs/vocab.js
johanbissemattsson: I think that the main gripe I have (coming mainly from a functional programming background) is when a function used for getting data (as in `getProperties`) is also setting data elsewhere. In my world that should be separate concerns.

I'm currently in favor of manually appending the data for each class inside https://github.com/libris/lxlviewer/blob/b144ccf2f55df8e8ed8cb118ba76337f2baa8392/nuxt-app/src/store/index.js#L188-L210 if there are any related properties using `domain`/`domainIncludes` or `range`/`rangeIncludes`... and not relying on `getProperties` to populate the vocabClasses store.

lxljs/vocab.js
niklasl: Yes. the formal view for class Person would be: "in domain of": [dates and names ...], and "in range of": [subject, hasProminentFamilyMember]. The Swedish labels are not set in stone; we need to agree on formal names here (ping @klngwll @andand79 ).

johanbissemattsson:
Should I merge this as is (with new labels but without alphabetic sorting) or wait until the sorting is finished? I'm on parental leave today so I will probably not manage to fix it today. :/

olovy (approved):
LGTM

niklasl:
Was the state mutation resulting in duplication of subclasses caused by this?

PR #872, opened 2023-06-14
Feature/lxl-4161 Remove placeholder from work search string

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4161](https://jira.kb.se/browse/LXL-4161)

### Solves

Removes placeholder (e.g. `{Primär medverkan utan agent}` or `{PrimaryContribution without agent}`) from work search string.

### Summary of changes

- Remove placeholder from work search string
- Trim search string
- Find and replace phrase in all languages found in `availableUserSettings`

### DIFF
--- a/vue-client/src/components/inspector/search-window.vue
+++ b/vue-client/src/components/inspector/search-window.vue
@@ -134,7 +134,7 @@ export default {
                 const cleanedChipString = DisplayUtil.getItemLabel(this.itemInfo, this.resources, this.inspector.data.quoted, this.settings)
                   .replace(/#|_|•|\[|\]/g, ' ')
                   .replace(/  +/g, ' ')
-                  .replace(new RegExp(`s?({(.*?)${StringUtil.getUiPhraseByLang('without', this.settings.language, this.resources.i18n)}(.*?)})`), '')
+                  .replace(new RegExp(`s?({(.*?)(${this.settings.availableUserSettings.languages.map(lang => StringUtil.getUiPhraseByLang('without', lang.value, this.resources.i18n)).join('|')})(.*?)})`, 'g'), '')
                   .trim();
                 this.keyword = cleanedChipString;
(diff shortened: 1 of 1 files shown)


vue-client/src/components/inspector/search-window.vue
```diff
@@ -131,7 +131,11 @@ export default {
             this.$nextTick(() => {
               this.resetSearch();
               if (this.itemInfo !== null) {
-                const cleanedChipString = DisplayUtil.getItemLabel(this.itemInfo, this.resources, this.inspector.data.quoted, this.settings).replace(/#|_|•|\[|\]/g, ' ').replace(/  +/g, ' ');
+                const cleanedChipString = DisplayUtil.getItemLabel(this.itemInfo, this.resources, this.inspector.data.quoted, this.settings)
+                  .replace(/#|_|•|\[|\]/g, ' ')
+                  .replace(/  +/g, ' ')
+                  .replace(new RegExp(`s?({(.*?)(${this.settings.availableUserSettings.languages.map(lang => StringUtil.getUiPhraseByLang('without', lang.value, this.resources.i18n)).join('|')})(.*?)})`, 'g'), '')
```
niklasl, line 137: This will work fine in this case,  as long as the word without isn't translated into something with regexp-specific characters (which I really don't expect here).

We should in general be very careful when creating regexps like this though, since they could blow up (e.g. if they contain unbalanced parens). See [MDN docs on regexp escaping](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping) for details.

johanbissemattsson: Fix using lo-dash’s `escapeRegExp` can be found here: https://github.com/libris/lxlviewer/pull/877
(not sure if a pull request is needed for such a small change)

PR #873, opened 2023-06-14
Fix/lxl-4055 Only show separator if there are transliterations

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4055](https://jira.kb.se/browse/LXL-4055)

### Solves

Removes the leftover transliteration separator if `transliteratedKeys` is empty.

### Summary of changes

- Only show separator if there are transliterations
- Change the transliteration separator character as the previous character could be rendered differently on different devices.


olovy (approved):
:ship:

PR #874, opened 2023-06-15
Bugs/lxl 4089 remove only selected field on duplicate values

didair:
## Checklist:
- [x] I have run the linter. `yarn lint` (Failed)

## Description
If a field has multiple values and one of them is removed, the entire field previously got removed. This PR adds support for removing only one of them using the item index. See example below:

bild

### Tickets involved
[LXL-4089](https://jira.kb.se/browse/LXL-4089)


olovy (commented):
LGTM!

(just to be clear: this fix is for the case when a property that is not allowed to have multiple values still has multiple values. For instance after uploading a bad jsonld file)

olovy (approved):

PR #880, opened 2023-06-26
Set button size on ReverseRelations component

didair:
## Checklist:

## Description
Fixes button alignment issues in reverse relations component

### Tickets involved
[LXL-4216](https://jira.kb.se/browse/LXL-4216)

### DIFF
--- a/vue-client/src/components/inspector/reverse-relations.vue
+++ b/vue-client/src/components/inspector/reverse-relations.vue
@@ -275,5 +275,11 @@ export default {
   &-spinner {
     margin-bottom: 10px;
   }
+
+  button {
+    height: 32px;
+    margin-top: 0;
+    margin-bottom: 0;
+  }
 }
 </style>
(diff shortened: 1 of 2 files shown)


niklasl:
It's great to solve this! I only wonder why it started appearing in QA when it doesn't in PROD? What changed to make this additional style now necessary?

didair:
@niklasl Great question! I will have a look

didair:
The .btn class is overwriting styles from .RoundedButton in qa and not in prod. So the same issue as previously seen. How do we go about fixing this without rewriting all stylesheet?

didair:
A better solution was to revert some changes I made last week to fix eslint errors https://github.com/libris/lxlviewer/commit/e2166c5a34e9201ab3821151ddeaaaa381c453ec and silence the warnings using eslint-disable-line. I still kept the height rule in reverse-relations so the buttons have the same height.

Skärmavbild 2023-06-27 kl  12 53 46

niklasl (approved):
LGTM! Nice catching of what affects the import order!

PR #881, opened 2023-06-28
Don't save empty fields to inspector.data

lrosenstrom:
## Checklist:

### Tickets involved
[LXL-4074](https://jira.kb.se/browse/LXL-4074)

### Solves
Fixes a bug where unstyled empty labels show up when saving records with empty language taggable fields, as described in [LXL-4074](https://jira.kb.se/browse/LXL-4074). 

### Summary of changes
* Don't save to inspector.data, while editing, if field content is empty. 
* Added extra isLocked-check so that nothing can be saved be mistake when the user is out of edit mode. This possibly solves some of the issues described in the comments of [LXL-4074](https://jira.kb.se/browse/LXL-4074) --- however, I was unable to reproduce those issues locally.

### DIFF
--- a/vue-client/src/components/inspector/item-bylang.vue
+++ b/vue-client/src/components/inspector/item-bylang.vue
@@ -150,7 +150,7 @@ export default {
       return false;
     },
     update(viewObjects) {
-      if (this.isHistoryView()) {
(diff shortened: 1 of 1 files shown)


didair (approved):
LGTM 👍

PR #882, opened 2023-07-05
Feature/lxl 4190 dont automatically save extracted work when editing

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4190](https://jira.kb.se/browse/LXL-4190), [LXL-4191](https://jira.kb.se/browse/LXL-4191)

### Solves

Saves and links extracted items first when when the user saves an instance, instead of saving the extracted work as soon as the user tries to create a link (which could result in mistakes and left-over unused works as the linking between the extracted item and the original instance was done in a later step).

Saving of extracted items are batched using the vuex store state `extractItemsOnSave`, enabling saving/linking the extracted data iteratively just before saving the original instance.

As the extraction and linking is done subsequently after each other, there is still some margin for errors (e.g. if the first request is successful but not the latter) – an more optimal solution would be relying on a single atomic request to the backend, in other words something to put on the wishlist...

### Summary of changes

- Remove preview button and update copy
- Batch saving/linking of extracted items
- Store items to extract as object with path keys


johanbissemattsson:
There are some (since earlier) unhealthy code duplication between `item-local.vue` and `item-sibling.vue` so all changes made in one of them also needs to be done in the other. We should investigate if we can reduce the repetition and merge them into one component.

However, I'm still a bit lost after reading the main comment in `item-sibling.vue` so I'm not sure of the use-case for it... https://github.com/libris/lxlviewer/blob/d43823dec6fc6e522dba0f5c3e5f3407cc989569/vue-client/src/components/inspector/item-sibling.vue#L2C1-L7

niklasl (approved):
LGTM AFAICS; let's get this into DEV to test thoroughly!

(We need a separate issue for the code duplication, which _should_ be changed to parameterization instead.)

PR #890, opened 2023-08-29
Fix marcframe display for auth och hold

olovy:
Get marcframe.json (expanded) from system and not from github.

Depends on https://github.com/libris/devops/commit/69675957dfc60a5cdc1bed5ec00e3360067b6efe

### Tickets involved
[LXL-4103](https://jira.kb.se/browse/LXL-4103)

### Solves
Marcframe view was broken for auth and hold.


niklasl (approved):
LGTM! :ship:

PR #891, opened 2023-08-31
Fix 410 Gone message on id.kb.se

andersju:
### Tickets involved
[LXL-4250](https://jira.kb.se/browse/LXL-4250)

### Summary of changes

Previously the formal page of a deleted resource would have the correct HTTP status code but show the generic error "There is a problem with the service. Try again later." (example: https://libris.kb.se/tz4ttnm0251wshd#it). This adds a special case for 410 Gone (like in the cataloging client).


niklasl (approved):
LGTM! :speedboat:

PR #893, opened 2023-09-08
Feature/lxl 2513 filter roles depending on contribution parent

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-2513](https://jira.kb.se/browse/LXL-2513)

### Solves

Filters roles depending on contribution parent.

### Summary of changes
- Filter linkable items if field is a `kbv:predicate`
- Use `URLSearchParams` interface to allow duplicate keys (`or-domain.@id`). This probably needs some QA to ensure it doesn't break something else.

### DIFF
--- a/vue-client/src/components/mixins/sidesearch-mixin.vue
+++ b/vue-client/src/components/mixins/sidesearch-mixin.vue
@@ -146,6 +146,7 @@ export default {
       /**
        * If field is a kbv:predicate (e.g. role) then filter linkable items depending on field and parent type.
        * */
+      // A VocabUtil.isSubPropertyOf(field.subPropertyOf, 'predicate', ...) would be preferable here.
       if (field.subPropertyOf.find(subProp => subProp['@id'] === VocabUtil.getTermObject('predicate', this.resources.vocab, this.resources.context)['@id'])) {
         const statement = VocabUtil.getTermObject(field.domain[0]['@id'], this.resources.vocab, this.resources.context); // e.g. Contribution
         const statementOf = statement.allowedProperties.find(p => p.domain?.find(d => d['@id'] === statement['@id'])); // e.g. contributionOf
--- a/vue-client/src/components/mixins/sidesearch-mixin.vue
+++ b/vue-client/src/components/mixins/sidesearch-mixin.vue
@@ -153,7 +153,7 @@ export default {
         const statementOfRangeIncludes = statementOf.rangeIncludes[0]['@id']; // e.g. Endeavour
         const subClassesOfRangeIncludes = VocabUtil.getSubClassChain(statementOfRangeIncludes, this.resources.vocabClasses, this.resources.context); 
 
-        const fieldParentPath = [...(this.path.split('.').slice(0, -2))].join('.'); // Maybe too specific... are we always sure we need to go up two steps to find field parent path?
+        const fieldParentPath = this.path.split('.').slice(0, -2).join('.');
         const fieldParentType = get(this.inspector.data, fieldParentPath)['@type']; // e.g. Text
         const fieldParentBaseClasses = VocabUtil.getBaseClasses(
           VocabUtil.getTermObject(fieldParentType, this.resources.vocab, this.resources.context)['@id'],
--- a/vue-client/src/components/mixins/sidesearch-mixin.vue
+++ b/vue-client/src/components/mixins/sidesearch-mixin.vue
@@ -1,4 +1,5 @@
 <script>
+import { get } from 'lodash-es';
 import { mapGetters } from 'vuex';
 import * as DisplayUtil from 'lxljs/display';
 import * as VocabUtil from 'lxljs/vocab';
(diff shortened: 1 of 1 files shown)


vue-client/src/components/mixins/sidesearch-mixin.vue
```diff
@@ -136,16 +127,49 @@ export default {
...
       });
-      if (typeof this.typeArray !== 'undefined' && this.typeArray.length > 0) {
-        params['@type'] = this.typeArray;
+
+      const field = VocabUtil.getTermObject(this.fieldKey, this.resources.vocab, this.resources.context);
+
+      /**
+       * If field is a kbv:predicate (e.g. role) then filter linkable items depending on field and parent type.
+       * */
+      if (field.subPropertyOf.find(subProp => subProp['@id'] === VocabUtil.getTermObject('predicate', this.resources.vocab, this.resources.context)['@id'])) {
```
niklasl, line 149:
```suggestion
      // A VocabUtil.isSubPropertyOf(field.subPropertyOf, 'predicate', ...) would be preferable here.
      if (field.subPropertyOf.find(subProp => subProp['@id'] === VocabUtil.getTermObject('predicate', this.resources.vocab, this.resources.context)['@id'])) {
```
In the actual case here, this works. But the base property (here `predicate`) *could* be higher up in a `subPropertyOf` chain. I don't see a `VocabUtil.isSubPropertyOf` alas, so we could just note that this would be preferable in a comment. (We can factor out such method (from `getDomainList` around lines 442-450 in `vocab.js`) and use  in a separate PR.)

vue-client/src/components/mixins/sidesearch-mixin.vue
```diff
@@ -136,16 +127,49 @@ export default {
...
+        const statement = VocabUtil.getTermObject(field.domain[0]['@id'], this.resources.vocab, this.resources.context); // e.g. Contribution
+        const statementOf = statement.allowedProperties.find(p => p.domain?.find(d => d['@id'] === statement['@id'])); // e.g. contributionOf
+        const statementOfRangeIncludes = statementOf.rangeIncludes[0]['@id']; // e.g. Endeavour
+        const subClassesOfRangeIncludes = VocabUtil.getSubClassChain(statementOfRangeIncludes, this.resources.vocabClasses, this.resources.context); 
+
+        const fieldParentPath = [...(this.path.split('.').slice(0, -2))].join('.'); // Maybe too specific... are we always sure we need to go up two steps to find field parent path?
```
niklasl, line 155: I'd say yes, given that we are within a reified relation, and that relation is directly on the "parent" entity we want here... (The path string repr which needs to be split and rebuilt is more awkward; but that's another level of redesign to address.)

niklasl (approved):
This looks very good! Let's try it out! :shipit: 

(I'd like to address the comments, but just some more clear code comments in this PR is perfectly fine. The refactoring is for another day (possibly incorporating https://github.com/libris/lxlviewer/tree/feature/more-object-oriented).)

vue-client/src/components/mixins/sidesearch-mixin.vue
johanbissemattsson: Made a small simplification in https://github.com/libris/lxlviewer/pull/893/commits/6291e0573bf0e55c82ef29d01c9b3c5f267f4b14 as strings already are immutable.

I have to think some more about a less awkward solution 🤔😊

vue-client/src/components/mixins/sidesearch-mixin.vue
```diff
@@ -136,16 +127,50 @@ export default {
...
+      const field = VocabUtil.getTermObject(this.fieldKey, this.resources.vocab, this.resources.context);
+
+      /**
+       * If field is a kbv:predicate (e.g. role) then filter linkable items depending on field and parent type.
+       * */
+      // A VocabUtil.isSubPropertyOf(field.subPropertyOf, 'predicate', ...) would be preferable here.
+      if (field.subPropertyOf.find(subProp => subProp['@id'] === VocabUtil.getTermObject('predicate', this.resources.vocab, this.resources.context)['@id'])) {
+        const statement = VocabUtil.getTermObject(field.domain[0]['@id'], this.resources.vocab, this.resources.context); // e.g. Contribution
+        const statementOf = statement.allowedProperties.find(p => p.domain?.find(d => d['@id'] === statement['@id'])); // e.g. contributionOf
+        const statementOfRangeIncludes = statementOf.rangeIncludes[0]['@id']; // e.g. Endeavour
```
olovy, line 153: I think these lines should handle 
* domain/range vs domainIncludes/rangeIncludes
* empty domain/range
* number of domainIncludes/rangeIncludes > 1 (by aborting is ok)
?

But it is not important now. We can fix that while refactoring lxljs.

olovy (approved):
:ship:

vue-client/src/components/mixins/sidesearch-mixin.vue
niklasl: I'd like to talk/whiteboard this, since i think:
1.  It _has_ to be `domain` (it's the only fact that entails something; "includes" are _only_ suggestions, they don't _mean_ what this is a property of).
2. The same goes applies for `range`, but I'm not sure why we check all allowedProperties here? (We got here via `contribution` and that is why we're using this logic?)
3. Yes, abort on empty sounds correct.

olovy: > I'd like to talk/whiteboard this, since i think:
> 
>     1. It _has_ to be `domain` (it's the only fact that entails something; "includes" are _only_ suggestions, they don't _mean_ what this is a property of).

:heavy_check_mark: suspected this

> 
>     2. The same goes applies for `range`, but I'm not sure why we check all allowedProperties here? (We got here via `contribution` and that is why we're using this logic?)

I think that was what was currently available in this context.
We should add better vocab preprocessing + lxljs methods to make clearer.

> 
>     3. Yes, abort on empty sounds correct.

:heavy_check_mark:

olovy: Actually I think the last line needs to handle all cases?
Out ff the `hasStatement` subproperties, `contribution` and `seriesMembership` have `domainIncludes` while `releationship` has `domain`.

johanbissemattsson: Like this?:
```js
const subClassesOfRangeIncludes = [
  ...new Set( // spread Set to ensure unique values in array
    [...(statementOf.range || []), ...(statementOf.rangeIncludes || [])].flatMap( // iterate over both range and rangeIncludes
      rangeItem => VocabUtil.getSubClassChain(rangeItem['@id'], this.resources.vocabClasses, this.resources.context), // get subclasses of e.g. Endeavour
    ),
  ),
];
```

johanbissemattsson: > Like this?:
> 
> ```js
> const subClassesOfRanges = [
>   ...new Set( // spread Set to ensure unique values in array
>     [...(statementOf.range || []), ...(statementOf.rangeIncludes || [])].flatMap( // iterate over both range and rangeIncludes
>       rangeItem => VocabUtil.getSubClassChain(rangeItem['@id'], this.resources.vocabClasses, this.resources.context), // get subclasses of e.g. Endeavour
>     ),
>   ),
> ];
> ```

☝️Added in the latest commit.

olovy: We merge but continue discussion

PR #897, opened 2023-09-18
Fix undo history when extracting items

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4190](https://jira.kb.se/browse/LXL-4190)

### Solves

Fixes undo history when extracting items.

The `changeHistory` state currently relies on implicit changes on the values of an item, this causes problems when waiting with extracting items until save as there shouldn't be any changes other than updating the id (which doesn't exist yet as it hasn't been extracted yet)...

So we need a work-around: the `changeList` payload is replaced with a constant (`EXTRACT_ON_SAVE`) indicating that the item should be extracted on save – this constant acts as an opt-out for the normal `changeList` functionality (it skips setting the inspector data with new values).

When cancelling an extraction the related `changeHistory` item is removed (even if it isn't the latest change).

### Summary of changes

- Fix undo history when extracting items
- Remove extraction from changeHistory

### DIFF
--- a/vue-client/src/store.js
+++ b/vue-client/src/store.js
@@ -1,12 +1,14 @@
 import Vue from 'vue';
 import Vuex from 'vuex';
-import { cloneDeep, each, set, get, assign, filter, isObject } from 'lodash-es';
+import { cloneDeep, each, set, get, assign, filter, isObject, isEqual } from 'lodash-es';
 import ClientOAuth2 from 'client-oauth2';
 import * as VocabUtil from 'lxljs/vocab';
 import * as StringUtil from 'lxljs/string';
 import * as httpUtil from '@/utils/http';
 import * as User from '@/models/user';
 import settings from './settings';
+
+const EXTRACT_ON_SAVE = '__EXTRACT_ON_SAVE__';
 
 Vue.use(Vuex);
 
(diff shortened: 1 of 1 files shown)


johanbissemattsson:
This PR is ready for review again! It now also removes the related `changeHistory` when cancelling an extraction.

olovy (approved):
LGTM!

I'm still seeing the issue where the first edit causes a seemingly random additional entry to be added to the changeHistory. But that seems unrelated to this.

johanbissemattsson:
> LGTM!
> 
> I'm still seeing the issue where the first edit causes a seemingly random additional entry to be added to `changeHistory`. But that seems unrelated to this.

Do you have a link to a record where this happens (or does it happen on all records)? It is as you mentioned probably unrelated to this PR but I'm having problems reproducing the random entries 🤔

olovy:
> > LGTM!
> > I'm still seeing the issue where the first edit causes a seemingly random additional entry to be added to `changeHistory`. But that seems unrelated to this.
> 
> Do you have a link to a record where this happens (or does it happen on all records)? It is as you mentioned probably unrelated to this PR but I'm having problems reproducing the random entries 🤔

It happens locally for me in e.g. in this one. Looks like the same problem in QA.
It is part of the work_example_data dataset so you should have it locally.
https://libris-qa.kb.se/katalogisering/2kc43g1d18l7tq2

Try changing e.g `hasTitle.mainTitle`.
`mainEntity.otherPhysicalFormat[0].editionStatement` is also added to `changeHistory.`

PR #898, opened 2023-09-20
Fix rewriting of array searchparam values

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-2513](https://jira.kb.se/browse/LXL-2513)

### Solves
Fixes rewriting of array search param values ensuring: `...@type=Person,Family,Meeting,Organization,Jurisdiction,Subject,Work,Place`
is rewritten to: 
...`@type=Person&@type=Family&@type=Meeting&@type=Organization&@type=Jurisdiction&@type=Subject&@type=Work&@type=Place`

This PR reverts to old `buildQueryString` helper instead of using the native `URLSearchParams` interface.

It would be preferable to replace buildQueryString but we have to ensure all array values are rewritten in a similar fashion as in buildQueryString, before doing so. Tests would be great to have here...

### Summary of changes

- Fix rewriting of array searchparam values

### DIFF
--- a/vue-client/src/components/mixins/sidesearch-mixin.vue
+++ b/vue-client/src/components/mixins/sidesearch-mixin.vue
@@ -3,6 +3,7 @@ import { get } from 'lodash-es';
 import { mapGetters } from 'vuex';
 import * as DisplayUtil from 'lxljs/display';
 import * as VocabUtil from 'lxljs/vocab';
+import { buildQueryString } from '@/utils/http';
 
 export default {
   data() {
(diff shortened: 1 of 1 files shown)


vue-client/src/components/mixins/sidesearch-mixin.vue
```diff
@@ -127,19 +137,15 @@ export default {
       this.fetch(n);
     },
     getItems(keyword) {
-      const searchPhrase = this.getSearchPhrase(keyword);
-      const urlSearchParams = new URLSearchParams({
```
niklasl, line 131: Note that `URLSearchParams` can also be passed a sequence of pairs, or have it's ` append(key, value)` method called multiple times with the same key to handle a repeated parameter.

johanbissemattsson: Yes, I think it would be preferable to use `URLSearchParams` again (I used the `append` method before but we need to ensure that it applies to all array values). This will be the scope of a later PR.

johanbissemattsson:
Closing in favour of #899

PR #901, opened 2023-10-12
Feature/changenote settings

lrosenstrom:
## Checklist:

### Tickets involved
[LXL-4268](https://jira.kb.se/browse/LXL-4268)

### Solves

Add change note subscription settings to user settings so that the user can select which categories will trigger a change notification.

### DIFF
--- a/lxljs/string.js
+++ b/lxljs/string.js
@@ -308,3 +308,16 @@ export function getFormattedEntries(list, vocab, language, context) {
   remove(formatted, value => value === '' || value === null); // Remove empty strings
   return formatted;
 }
+
+export function getSigelLabel(sigel, len) {
+  if (!sigel.friendly_name) {
+    return sigel.code;
+  }
+
+  const sigelPart = ` (${sigel.code})`;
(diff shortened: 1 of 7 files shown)


lxljs/string.js
```diff
@@ -308,3 +308,16 @@ export function getFormattedEntries(list, vocab, language, context) {
   remove(formatted, value => value === '' || value === null); // Remove empty strings
   return formatted;
 }
+
+export function getSigelLabel(sigel, len) {
```
klngwll, line 312: It is mostly a hygiene detail. Is it possible to use a bit more generic or less libris-centric names excluding the word sigel in some of these function or var/const names? Since the sigel is by our definition the code on something and not the actual entity it describes (contrary to how we use it in natural language, the hard part here being it plays different roles and is not always a library). Are we meaning userLocation? Not a stopper by any means and should it turn out to be too timeconsuming to figure out I wouldn't argue it easier to revisit later when the model and definitions around this is clearer.

lrosenstrom: We can change it to 'collection', because that's what it is called in the user store?

PR #910, opened 2023-11-10
Feature/changenote settings only

lrosenstrom:
Change note settings only.


olovy (approved):
It's more than just the settings but LGTM...

PR #913, opened 2023-11-15
Fix creation of local entities

johanbissemattsson:
Falsy options were coerced to the string `false`, resulting in a truthy value. In vue3 `null` or `undefined` should be used to explictly remove an attribute. See https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html#coercing-false-to-false-instead-of-removing-the-attribute

## Checklist:

### Tickets involved
[LXL-4339](https://jira.kb.se/browse/LXL-4339)

### Solves

Fix creation of local entities

### Summary of changes

Fix creation of local entities by using `null` instead of `false` inside the `selectOptions` method inside `sidesearch-mixin.vue`


jesperengstrom (approved):

jesperengstrom:
Ship it!

PR #917, opened 2023-11-20
Feature/change view

lrosenstrom:
### Tickets involved
[LXL-4269](https://jira.kb.se/browse/LXL-4269)

### Solves
Let the user view and filter administrative notices / change notes.
TODO: marking a notice as handled works (by creating an administrative action on checked check-box), but removing it requires a coming backend fix. 

### Summary of changes

- New tab under directory care focused on searching among administrative notices / change notes.
- Facets in this view work as OR-facets
- Categories and collections that are checked in the user settings change note section are initialized as checked facets.

Screenshot from 2023-11-20 10-54-02

### DIFF
--- a/vue-client/src/components/search/facet-group.vue
+++ b/vue-client/src/components/search/facet-group.vue
@@ -102,7 +102,6 @@ export default {
       const self = this;
       const list = this.group.observation.map((o) => {
         let label;
-        console.log('observation', JSON.stringify(o));
         if (o.object.hasOwnProperty('@id')) {
           label = DisplayUtil.getItemLabel(
             o.object,
--- a/vue-client/src/components/search/facet-group.vue
+++ b/vue-client/src/components/search/facet-group.vue
@@ -59,7 +59,7 @@ export default {
       this.$store.dispatch('setUser', userObj);
     },
     featuredComparison(facet) {
-      if (this.group.dimension === '@reverse.itemOf.heldBy.@id') {
+      if (this.isCollectionGroup(this.group.dimension)) {
         // Featured code for '@reverse.itemOf.heldBy.@id'
         const userSigels = this.user.collections.map((item) => item.code);
         if (facet.object.hasOwnProperty('sigel')) {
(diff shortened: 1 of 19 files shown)


vue-client/src/components/search/facet-group.vue
```diff
@@ -167,11 +181,24 @@ export default {
     hasScroll() {
       return !this.revealLevels[this.currentLevel] && this.isExpanded;
     },
+    isChangeFacetGroup() {
+      return (this.group.dimension === 'category.@id' || this.group.dimension === 'concerning.@reverse.itemOf.heldBy.@id') && this.isChangeView;
+    },
```
olovy, line 186: Is this check necessary? Which facets are we filtering out?
We want `@type` here as well?

vue-client/src/components/search/facet-group.vue
```diff
@@ -237,13 +268,20 @@ export default {
             :show-iconless="false"
             v-if="group.dimension === 'instanceOf.@type' || group.dimension === '@type'"
             :type="facetItem.object['@id']" />
+          <check-box v-if="group.dimension === 'concerning.@reverse.itemOf.heldBy.@id' && isChangeView"
```
olovy, line 271: This check should be on the presence of the `_selected` property so that it works for any multi-select facet group.

vue-client/src/components/search/facet-group.vue
```diff
@@ -254,6 +292,12 @@ export default {
             :show-iconless="false"
             v-if="group.dimension === 'instanceOf.@type' || group.dimension === '@type'"
             :type="facetItem.object['@id']" />
+          <check-box v-if="isChangeFacetGroup"
```
olovy, line 295: This check should be on the presence of the _selected property so that it works for any multi-select facet group.

vue-client/src/views/ChangeNotes.vue
```diff
@@ -0,0 +1,24 @@
+<script>
+import Find from './Find.vue';
+
+export default {
+  name: 'ChangeNotes',
```
olovy, line 5: Name could be updated since they're not ChangeNotes any longer. `AdminNotices`?

olovy (approved):
See comments + we need to improve the appearance of the "Handled" button/check box.
But I think we can merge this as-is.

vue-client/src/components/search/facet-group.vue
lrosenstrom: Yes, this was introduced to explicitly hide the @type facets, but if we want to show that I can remove this.

PR #918, opened 2023-11-20
Add refined titles when extracting and allow editing

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4297](https://jira.kb.se/browse/LXL-4297), [LXL-4316](https://jira.kb.se/browse/LXL-4316)

### Solves

Copies and refines the title from the main entity if the extractable item is missing a title. This was earlier done automatically behind the scenes in a later step when the user clicks on save.

The user is now also allowed to continue editing the fields inside the extractable item (as we now only save the paths which should be extracted on save instead of storing the entire state of the extractable fields).

### Summary of changes

- Copy and refine title from the main entity if a title is missing on the extractable item.
- Remove reliance on vuex state for storing the entire state of the extractable fields.
- Shows different notifications depending on if title was copied/redefined or not.


johanbissemattsson:
I'm working on making the extraction action _undoable_ but it isn't release ready (it needs to be more thought-through) so it will sadly not be finished before the feature-stop.

kwahlin (approved):
The title copying part looks good! As for the rest I say let's merge it and then test properly in QA.

lrosenstrom (approved):
LGTM!

PR #919, opened 2023-11-21
LXL-4383: item-value update bug

jesperengstrom:
## Checklist:

### Tickets involved
[LXL-4383](https://jira.kb.se/browse/LXL-4383)

### Solves

Weird bug that causes some input fields (`item-value`) to revert to their previous value on mouseleave.
reproduce:
* create new --> 'agent' --> 'person'
* put the cursor in 'födelsedatum' and type something. Move the cursor to 'dödsdatum'. The text then disappears and re-appears after 1 sec.

### Summary of changes

Moving the cursor from the field causes the component to update (in Vue 2, it did not apparently) resetting to the prop value. But since the update function has a 1s debouncer, the current text has not yet been synced back from the store, causing the old value to be displayed for a short while.

the `@mouseover`/ `@mouseleave` in `field` is what causes the update in (some of!) the components. This is used for the `showActionButton` functionality that has been removed and does nothing. Seems to me that the easiest solution is to remove this altogether. There may even be a performance upside to stop listening to mouseover events on all the fields...

### DIFF
--- a/vue-client/src/components/inspector/entity-adder.vue
+++ b/vue-client/src/components/inspector/entity-adder.vue
@@ -40,10 +40,6 @@ export default {
     compositional: {
       default: null,
     },
-    showActionButtons: {
-      type: Boolean,
-      default: false,
-    },
     isPlaceholder: {
       type: Boolean,
       default: false,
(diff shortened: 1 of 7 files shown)


lrosenstrom (approved):
LGTM :ship:

PR #922, opened 2023-11-22
Add undoable extract actions

johanbissemattsson:
## Checklist:

### Solves

Add undoable extract actions and add label clarifying **what** is going to be extracted and **when** (on save).

### Summary of changes

- Add undoable extract actions
- Add label clarifying what is going to be extracted and when (on save).


kwahlin (approved):
Tested locally, works fine. Let's test this in QA too.

olovy (approved):

PR #923, opened 2023-11-27
Ensure highlight is removed when undoing extraction

johanbissemattsson:
## Checklist:

### Solves

Ensures the highlight is removed when undoing extraction actions (sometimes the blue highlight could be stuck which could confuse the user if the item will be extracted or not).

### Summary of changes

- Ensure highlight is removed when undoing extraction


olovy (approved):
LGTM

PR #925, opened 2023-11-30
Fix inactive ShelfMarkSequences showing up in entity adder

olovy:
## Checklist:

### Tickets involved
[LXL-4404](https://jira.kb.se/browse/LXL-4404)

### Solves
Inactive ShelfMarkSequences and ShelfMarkSequences owned by other libraries showing up when linking shelfMark.

The inactive ones should also have been blocked from linking by the client because they are `replacedBy` the new ones. But  `replacedBy` is not indexed (not in card for `ShelfMarkSequence`).

bild
 
### Summary of changes
Moved the check to where params are built now.
b9705205 was the breaking change. `getSearchPhrase` is not called anymore

### DIFF
--- a/vue-client/src/components/inspector/entity-adder.vue
+++ b/vue-client/src/components/inspector/entity-adder.vue
@@ -198,6 +198,7 @@ export default {
     translatePhrase,
     labelByLang,
     capitalize,
+    // TODO: dead code?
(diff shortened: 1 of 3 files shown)


olovy:
FYI

PR #926, opened 2023-11-30
Only fetchMissingLinkedToQuoted for links within system

olovy:
## Description
When opening a document we try to lookup all links not in quoted / embellish data.
This is useful when opening a template where we don't have the definitions for the linked things in the template loaded.
However, we shouldn't try to lookup URIs outside libris.kb.se / id.kb.se.
Only `fetchMissingLinkedToQuoted` for links within system.

### Solves
`fetchMissingLinkedToQuoted` trying to lookup things outside libris.kb.se / id.kb.se


niklasl (approved):
LGTM :100: :shipit:

PR #931, opened 2023-12-20
LXL-4421: Remove loop causing item-error to dispatch new errors

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4421](https://jira.kb.se/browse/LXL-4421)

### Solves
Fixes issue causing Vue to slow down or crash due to `Maximum recursive updates exceeded in component` while in debug mode. The `item-error.vue` component automatically dispatched unwanted `setValidation` actions in debug mode, resulting in a loop as the actions triggered new errors.

The code removed is five years old so the issue shouldn't be related to the Vue3 migration (but maybe Vue2 has a different escape hatch for recursive loops? 🤷)

### Summary of changes

- Remove loop causing item-error to dispatch new errors

### DIFF
--- a/vue-client/src/components/inspector/item-error.vue
+++ b/vue-client/src/components/inspector/item-error.vue
@@ -26,19 +26,7 @@ export default {
         delete cleanItem._uid;
       }
       return cleanItem;
-    },
-    failedValidations() {
-      const failedValidations = [];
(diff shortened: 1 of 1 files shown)


olovy (approved):
LGTM.

Tested locally. No freezes.

PR #932, opened 2023-12-21
Fix LXL-4422 bug

lrosenstrom:
## Checklist:

### Tickets involved
[LXL-4422](https://jira.kb.se/browse/LXL-4422)
### Solves

- Fixes as bug where newly input text was removed (after a short delay) when adding a new field.

### Summary of changes

Remove unspecified parameter/property on initialization of item-value.vue, item-shelf-control-number.vue and item-next-shelf-control-number.vue. The parameter is already specified as a computed in the components, not a property.


olovy (approved):
LGTM!

Tested locally, works fine.

PR #940, opened 2024-01-12
Move VS Code workspace file to app root

johanbissemattsson:
### Solves
Moves VS Code Workspace file to app root making it easier to open in VS Code using "Open Workspace from File...".

The earlier location inside the hidden .vscode folder made it difficult to find the file.

### Summary of changes

- Move VS Code Workspace file to app root


jesperengstrom (approved):
Yep, had trouble opening the workspace file too 👍

PR #942, opened 2024-01-15
Bugfix/user db sometimes not loaded

lrosenstrom:
## Checklist:

## Description
For a recently logged in user, navigating directly to the user settings view does not necessarily load the user database. Hence, editing the change message subscription section does not work and causes an error. To the user, it seems like the changes take effect, since the boxes still can be checked/unchecked, however these changes are not persistent. This is wrong and causes confusion.

### Tickets involved
--

### Summary of changes

- Load user db when user settings component is mounted
- Don't try to load the user db (and suppress the error) when logged out

### DIFF
--- a/vue-client/src/App.vue
+++ b/vue-client/src/App.vue
@@ -366,9 +366,11 @@ export default {
   mounted() {
     this.verifyUser().then(() => {
       this.$nextTick(() => {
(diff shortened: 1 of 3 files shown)


olovy (approved):
LGTM!

PR #944, opened 2024-01-19
Change path in VS Code workspace file

jesperengstrom:
### Solves

After moving the workspace file we also need to change the relative path from parent dir to current
(we vant `lxl-web` and `lxljs`, not `lxlviewer` and `lxljs`).


johanbissemattsson (approved):
LGTM! Works as intended!

PR #947, opened 2024-01-23
Add design fundament to Tailwind

jesperengstrom:
[## Description

### Tickets involved
[LXL-4431](https://jira.kb.se/browse/LXL-4431)

### Summary of changes

Add the design fundamentals from current design to the Tailwind theme. I've tried to have some priciples in mind when setting it up:

* The theme needs to be flexible and applying new colors should be possible by only swapping a file or changing a value (support white label/LOPAC feature/dark mode).
* We should follow the Tailwind pattern of styling components inline using the built in utilities. While at the same time make it as easy as possible to apply the correct styling  and make it hard(er) to divert from the design.

This means:

* I've only defined the base colors as the palette. Since shades are different levels of tranparency, i suggest we use the built in transparent modifier (`bg-primary/80`). Colors now have abstract names that refer to CSS variables located in a separate colors.css file.

* I've added a set of classes corresponding to typography sizes/variants 1-6 in design (`text-1-cond-caps`, `text-3-cond-bold` etc - the `addUtilites`function was the only way to include them in vscode autocomplete). These will save us some thime styling text I think.

* The design also contains a lot of other named classes (`bg-toggle-active` etc). Adding and maintaining these would be too much work, we should just style them inline in the components. I've cleaned out many of Tailwind's default values (in theme `fontSize`, `fontWeight`, `lineHeight`, `letterSpacing`, `borderRadius`) and replaced them with the ones present in the design to constrain our options.

Other notes:

* I'm a fan of [Tailwind config viewer](https://github.com/rogden/tailwind-config-viewer) for visualizing the theme, but sadly it doesn't work with css variables. So i made a basic theme viewer component, we can throw it away later.
* The Roboto Flex font is nice, but the downloadable version is way too big for us, just not suitable as a web font. The Google font api supports serving a subset of the font in woff2 format, but we don't want to use that. My suggestion is we let the design stabilize a bit and then somehow just include the font-styles, weights and widths we need as woff2.


lxl-web/tailwind.config.js
```diff
@@ -3,32 +3,109 @@
...
+			'3xl': ['2.5rem', theme('lineHeight.tight')],
+			'4xl': ['4rem', theme('lineHeight.tight')]
+		}),
 		fontFamily: {
-			// 'sans': ['roboto-flex', 'system-ui', ...],
+			sans: ['"Roboto Flex"', 'sans-serif'],
+			condensed: [
+				'"Roboto Flex", sans-serif',
+				{
+					fontVariationSettings: '"wdth" 60'
```
johanbissemattsson, line 40: `wdth` looks like a small typo

lxl-web/src/app.html
johanbissemattsson, line 1: I would keep this as is for the small  <sub><sup>miniscule</sup></sub> performance boost using inline styling for the initial rendering. I normally would advice against this, but think this could be an exception as `display: contents` is omnipresent on all pages.

johanbissemattsson (approved):
LGTM!

PR #950, opened 2024-01-24
Trim sibling whitespaces using a preprocessor

johanbissemattsson:
### Solves
Trims sibling whitespaces using a preprocessor which inlines all the template code in specified Svelte files (currently only `DecoratedData.svelte`).

This preprocessor should be seen as an rather blunt work-around until https://github.com/sveltejs/svelte/issues/189 is resolved.
An alternative is to manually inline the relevant code in `DecoratedData.svelte` together with a `prettier-ignore` block.

### Summary of changes

- Trim sibling whitespaces using a preprocessor


johanbissemattsson:
Replaced by #973

PR #956, opened 2024-02-08
Add site header

jesperengstrom:
### Tickets involved
[LXL-4443](https://jira.kb.se/browse/LXL-4443)

### Summary of changes

- Adds a `SiteHeader` component to the app layout. It includes the `Search` and `LangPicker` components and returns a diffferent template if user is on startpage (/).

- Modified tailwind screen sizes and layout helper class (`.container-fluid`) to align with design.
- More fine-grained css color variables to align with design naming. For example:
  -  `.text-primary` --> `var(--text-primary)` --> `var(--color-primary)` --> `var(--color-brown)`
  - `.text-secondary` --> `var(--text-primary) / 0.8` etc
 
  It lets us both change the primary color, or the text color without changing the primary color etc...

TODO: button for search. We might as well do a single component that handles all style variations.


johanbissemattsson:
Skärmavbild 2024-02-09 kl  09 10 42
Skärmavbild 2024-02-09 kl  09 10 59
The pages look strange when testing it out, I guess it is related to dark mode / `prefers-color-scheme`?

lxl-web/src/lib/assets/json/settings.json
johanbissemattsson, line 1: I'm thinking if another json file is needed here if it's only used for the service name? Couldn't we use either a static or dynamic environment variable instead (so we can skip adding passing it around using pageData)?

lxl-web/src/routes/(app)/[[lang=lang]]/+page@(app).svelte
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/+page@(app).svelte:1]: I prefer having shared layouts as long as possible to prevent unnecessary remounting of shared components. The unmounting also clears the local/internal state in the components which can have unwanted effects if the user has started interacting with elements which are present in both layouts.

This is probably not a problem right now as the search field probably _should_ be reset when navigating to the start page, but will cause issues if we for example have menu buttons for accessing saved items / or toggles for opening and closing toolbars.

But I think we can keep it as is for now as the start page will need to be revisited when the design is more settled.

jesperengstrom:
> The pages look strange when testing it out, I guess it is related to dark mode / `prefers-color-scheme`? (I have the system settings set to dark mode)

Heh, yep and it's broken after varable name change. Commented out the dark mode thing, it was meant as a reminder if we want to do it later...

lxl-web/src/lib/assets/json/settings.json
johanbissemattsson: Ah, I guess your assumption is correct... or it depends if we will support white-labeling of our client?

If it's not intended to be changed then I would probably keep "Libris" as a hard-coded string when used, until we have a need for multiple settings. Sorry for complicating it :D

lxl-web/src/routes/(app)/[[lang=lang]]/+page@(app).svelte
Comment (johanbissemattsson) [reply lxl-web/src/routes/(app)/[[lang=lang]]/+page@(app).svelte:1]: I generally prefer the later example, at least if the design allows it and the logic inside component doesn't grow too much. We can always refactor/split it into multiple components when the file becomes to big/unwield (I also generally prefer too split things first when they become a problem).

But as with everything it can _depend_ on a lot of different things (mainly on the final design). So I'm totally fine with keeping it as currently is in the PR and when the final design is settled, revisit it.

Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/+page@(app).svelte:1]: I'm 100% for not adding complexity when we don't need to. It turned out making the `SiteHeader` adaptable was much more straightforward. (Probably still stuck in Nuxt land, where a layout is mostly a template you can use and components can only return one root element)

lxl-web/src/colors.css
```diff
@@ -10,17 +10,33 @@ see https://tailwindcss.com/docs/customizing-colors#using-css-variables */
...
+
+	/* Text colors */
+	--text-primary: var(--color-primary);
+	--text-primary-inv: var(--color-white);
+	--text-link: var(--color-accent-dark);
+
+	/* Bg colors */
+	--bg-main: var(--color-subtle);
+	--bg-head: var(--color-accent-light);
+	--bg-positive: var(--color-accent-light);
```
johanbissemattsson, line 28: When is `--bg-positive` intended to be used? For active states on elements?

johanbissemattsson (approved):

johanbissemattsson:
This is probably best kept as separate issue/branch but I think `autofocus` on the search input should only be enabled when  visiting the index page (it will otherwise snatch the focus everytime the user applies a filter as the route changes).

lxl-web/src/colors.css
jesperengstrom: I interpret it as the bg of an 'active' element such as (in this case) a filter pill.
It's a figma variable, there are a bunch:
Skärmavbild 2024-02-14 kl  09 07 22

I'm not sure if we should add all these to the project. The upside is it adds flexibility (for whitelabelling etc) and easy styling if they are used often. The downside is we'll eventually get a lot of variables, harder to maintain. I've just added those i've encoutered so far. But i'm up for discussion :)

PR #957, opened 2024-02-09
Add resource popover/tooltip for previewing linked data

johanbissemattsson:
### Tickets involved
[LXL-4447](https://jira.kb.se/browse/LXL-4447)

### Solves
Add resource popover/tooltip for previewing linked data (when hovering over links and definitions). [Floating UI](https://floating-ui.com/) is used for placement of the popover.

The popovers are added by using a custom [svelte action](https://svelte.dev/docs/svelte-action) called `resourcePopover` (note that the `ResourcePopover.svelte` component which the action conditionally renders isn't meant to be used directly in the template code).

The action is currently only used in `DecoratedData.svelte` but can be used on any HTML element by simply adding `use:resourcePopover={{'@id': ADD_FNURGEL_HERE}}`

The action adds event listeners on the trigger node and will render a popover if certain conditions are fulfilled. More specifically; when the user hovers over a link or definition two cancellable promises are started at the same time and the popover will only be rendered _if_ and _when_ both promises are resolved.

**The promises / conditions are:**

1. **The user hovers over a trigger element for a set amount of time (currently 500 ms).** This delay is added to prevent obnoxious amounts of popovers when the user moves the cursor around the page. If the user leaves the trigger node the attachment is cancelled.
2. **The fetching of the required data has finished.** The fetch uses a separate timeout (currently set to 250ms) so the fetching starts before the popup appears. This timeout is also added to prevent obnoxious amounts of fetches when the user moves the cursor around the page.

If _any_ of the above is cancelled or fails then no popover will appear. 

The fetch uses a proxy api route to fetch decorated data (`api/[[lang]]/[fnurgel]`). Currently a simple HTTP cache is used but future improvements can be made using localeStorage (or preferably IndexedDB?) to save all linked data locally and  prevent unnecessary refetching. But we then need some way to update the local cache when the data is updated.

### Summary of changes

- Add Floating UI dependency
- Add API route for decorated resource
- Add resource popover action

### DIFF
--- a/lxl-web/src/lib/actions/resourcePopover/index.ts
+++ b/lxl-web/src/lib/actions/resourcePopover/index.ts
@@ -1,0 +1,115 @@
+import type { Action } from 'svelte/action';
+import type { ResourceData } from '$lib/types/ResourceData';
+import ResourcePopover from './ResourcePopover.svelte';
+import { getResourceId, getResourcePropertyStyle } from '$lib/utils/resourceData';
+
+/** Tests to do
+ * - [] Attaches popover when user hovers over trigger node (after delay)
+ * - [] Removes popover when user stops hovering over trigger node (after delay)
+ * - [] Removes popover when user blurs trigger node
+ * - [] Keeps popover open if user hovers over popover (when leaving trigger node)
+ * - [] Removes popover if user stops hovering over popover (without entering trigger node)
+ * - [] Keeps popover if user has focused child element
+ * - [] Doesn't attach popover if fetching of resource data fails
+ */
+
+/* Conditionally add popover action so it's only added when needed */
+const conditionalResourcePopover = (node: HTMLElement, value: ResourceData) => {
+	const style = getResourcePropertyStyle(value);
+	if (style && style.includes('link' || style.includes('definition'))) {
+		const id = getResourceId(value);
+		if (id) {
+			return resourcePopover(node, id);
+		}
+	}
+};
+
+const resourcePopover: Action<HTMLElement, string> = (node: HTMLElement, id: string) => {
+	const FETCH_DELAY = 250;
+	const ATTACH_DELAY = 500;
+	const REMOVE_DELAY = 200;
+	const container = document.getElementById('floating-elements-container') || document.body; // See https://atfzl.com/articles/don-t-attach-tooltips-to-document-body
+
+	let attached = false;
+	let floatingElement: ResourcePopover | null = null;
+	let cancelAttach: (reason?: unknown) => void;
+	let cancelRemove: (reason?: unknown) => void;
+	let cancelFetch: (reason?: unknown) => void;
+
+	node.addEventListener('mouseover', attachPopover);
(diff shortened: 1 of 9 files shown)


johanbissemattsson:
I have now moved the conditional logic for adding resource popovers on decorated data to the `DecoratedData.svelte` component. This enables easier re-usability of the `resourcePopover` action.

You can now just add `use:resourcePopover={ADD_RESOURCE_ID_HERE}` on _any_ HTML element if you want to allow previews of linked data.

lxl-web/src/lib/actions/resourcePopover/index.ts
```diff
@@ -0,0 +1,102 @@
...
+	const container = document.getElementById('floating-elements-container') || document.body; // See https://atfzl.com/articles/don-t-attach-tooltips-to-document-body
+
+	let attached = false;
+	let floatingElement: ResourcePopover | null = null;
+	let cancelAttach: (reason?: unknown) => void;
+	let cancelRemove: (reason?: unknown) => void;
+	let cancelFetch: (reason?: unknown) => void;
+
+	node.addEventListener('mouseover', attachPopover);
+	node.addEventListener('mouseout', removePopover);
```
jesperengstrom, line 27: Enable tooltip when using keyboard too?
```
node.addEventListener('focus', attachPopover);
node.addEventListener('blur', removePopover);
```

lxl-web/src/lib/components/DecoratedData.svelte
```diff
@@ -26,17 +21,12 @@
...
-		}
-		return undefined;
-	}
-
 	function getLink(value: ResourceData) {
-		const hints = getObjectProperty(value, '_hint');
-		if (Array.isArray(hints) && hints.includes('link')) {
-			return value?.['@id' as keyof typeof value];
+		if (getResourcePropertyStyle(value)?.includes('link')) {
+			const id = getResourceId(value);
```
jesperengstrom, line 26: There's a `relativize` util in utils/http that might be useful here...

lxl-web/src/lib/actions/resourcePopover/index.ts
```diff
@@ -0,0 +1,102 @@
...
+		removePopover();
+	}
+
+	async function getDecoratedData(id: string) {
+		await new Promise((resolve, reject) => {
+			cancelFetch = reject;
+			setTimeout(resolve, FETCH_DELAY);
+		});
+		try {
+			const resourceRes = await fetch(`/api/${id.split('/').pop()}`);
```
jesperengstrom, line 87: Include the `lang` param here when fetching?

johanbissemattsson: This is resolved in the latest commit, where the action now also requires `lang`. See example:
```
resourcePopover(node, {
  id,
  lang: getSupportedLocale($page.params.lang)
});
```
This will not be needed in Svelte v5 when we can read `$page.params.lang` directly inside the action.

jesperengstrom (approved):
Awesome, nice work!

PR #963, opened 2024-02-16
Result sorting

jesperengstrom:
### Tickets involved
[LWS-56](https://jira.kb.se/browse/LWS-56)

### Summary of changes

* Adds a `select` element with 3 options (for now, more will be added) that sort the current search result
* Display the number of `totalItems` / no hits

Extra:
* Added very basic error handling to `find/page.server` (I noticed that 'hacking' the `_sort` param caused the app to crash since we never check if the fetch was successful before trying to run xl methods on the response)
* Only show the filter panel if there are hits

### DIFF
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
@@ -37,16 +37,16 @@
 					{/if}
 				</p>
 				{#if numHits > 0}
-					<form on:change={handleSortChange}>
+					<div>
 						<label for="search-sort">Sortera efter</label>
-						<select name="search-sort">
+						<select id="search-sort" on:change={handleSortChange}>
 							{#each sortOptions as option}
 								<option value={option.value} selected={option.value === sortOrder}
 									>{option.label}</option
 								>
 							{/each}
 						</select>
-					</form>
+					</div>
 				{/if}
 			</div>
 			<ol class="flex flex-col gap-2">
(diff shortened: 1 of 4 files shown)


lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
```diff
@@ -2,18 +2,55 @@
...
+					{#if numHits && numHits > 0}
+						{numHits.toLocaleString($page.data.locale)} träffar
+					{:else}
+						Inga träffar
+					{/if}
+				</p>
+				{#if numHits > 0}
+					<div>
+						<label for="search-sort">Sortera efter</label>
+						<select id="search-sort" on:change={handleSortChange}>
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte:42]: We can add a form attribute here (e.g. `form="main-search"`) to associate the select input with the form used for searching (which enables some sort of fallback if javascript isn't available).

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#form

johanbissemattsson (approved):

johanbissemattsson:
LGTM!

lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte:42]: Yep. I guess it will then not react to any change in the select menu, but rather include the value on form submission (= search click). But it's better than nothing!

PR #964, opened 2024-02-21
Don't add an empty change note in adminmetadata

olovy:
## Description
It is obsolete as long as we don't enable automatic `ChangeObservation`s

### Tickets involved
[LXL-4454](https://jira.kb.se/browse/LXL-4454)


lrosenstrom (approved):
LGTM! :ship:

PR #965, opened 2024-02-21
Don't drop query part when navigating within CXZ message view

lrosenstrom:
Fixes a bug where the CXZ message search results turns empty when clicking Directory Care / Katalogvård. 
See more in [LXL-4444](https://jira.kb.se/browse/LXL-4444).


lrosenstrom:
Don't know if there is a better way to do this which is more isolated!

olovy (approved):
Works! :ship:

PR #968, opened 2024-02-22
Search result pagination

jesperengstrom:
### Tickets involved
[LWS-24](https://jira.kb.se/browse/LWS-24)

### Summary of changes
* Adds a component that allows the user to paginate the search result
* The pagination includes outer (conditionally displayed prev/next, first/last) navigations and an inner sequence of pages (currently 2) in relation to the current page.
* We want to reset the `_offset` on new search (?), so the `addDefaultSearchParam` condition is reversed; reset if it exists, else omit it (defaulting to 0)
* Renamed http util `relativize` -> `relativizeUrl` and extended it to accept relative url:s - removing its opening slash to keep the /lang/ href. Applied to pagination as well as facets.
* Added and updated tests

### DIFF
--- a/lxl-web/src/lib/components/DecoratedData.svelte
+++ b/lxl-web/src/lib/components/DecoratedData.svelte
@@ -9,7 +9,7 @@
 		getResourceId,
 		getPropertyValue
 	} from '$lib/utils/resourceData';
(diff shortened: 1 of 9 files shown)


johanbissemattsson (approved):
Awesome! 🙌 LGTM!

PR #969, opened 2024-02-23
Add support for picking properties from decorated data

johanbissemattsson:
### Tickets involved
[LWS-12](https://jira.kb.se/browse/LWS-12)

### Solves

Adds support for picking properties from decorated data

### Summary of changes
- Add support for picking properties from decorated data
- Add support for showing property labels
- Control label visibility using nolabel and showLabels override
- Add block prop to DecoratedData component


jesperengstrom (approved):
LGTM! 👍

PR #970, opened 2024-02-26
Fix redirect after login

jesperengstrom:
## Checklist:

### Tickets involved
[LXL-4427](https://jira.kb.se/browse/LXL-4427)

### Solves
* The problem with not properly redirecting after login.

This has to do with the app treating a user as authenticated (by `verifyUser` function resolving the promise) as long as they have a token and the call to the verify endpoint is successful, even if the token is invalid or expired. The app only stores the `lastPath` for unauthenticated users, resulting in the path not updating for users who have an old token stored.

### Summary of changes
* Check if the response from Libris login contains a user or is in fact an error message. If so, throw an error that rejects the promise.

### DIFF
--- a/vue-client/src/store.js
+++ b/vue-client/src/store.js
@@ -687,6 +687,9 @@ const store = createStore({
             headers,
             method: 'GET',
           }).then((response) => response.json()).then((result) => {
+            if (!result.user) {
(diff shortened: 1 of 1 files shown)


olovy (approved):
Nice!
:ship:

PR #972, opened 2024-02-27
Break out web stuff from display

jesperengstrom:
### Tickets involved
[LWS-59](https://jira.kb.se/browse/LWS-59)

### Solves

Breaks out lxl-web lens and formatting into a local file for easier management

To be tested together with https://github.com/libris/definitions/pull/473

### Summary of changes

* Create display-web.json
* Merge contents with display on load
* Fix so that app doesn't break if display-web is empty, for example


olovy (approved):
LGTM!

PR #973, opened 2024-02-29
Trim sibling whitespaces using a preprocessor

johanbissemattsson:
### Solves
Trims sibling whitespaces using a preprocessor which inlines the template code in specified Svelte files (currently only `DecoratedData.svelte`).

The preprocessor replaces whitespace characters (at the end of a line) and tabs which are followed by Svelte specific logic blocks (by matching their starting characters) or HTML tags.

This preprocessor should be seen as a work-around until https://github.com/sveltejs/svelte/issues/189 is resolved.

An alternative is to manually inline the relevant code in `DecoratedData.svelte` together with a `prettier-ignore` block.

### Summary of changes

- Trim sibling whitespaces using a preprocessor


olovy (approved):
Seems to work!
:ship:

PR #975, opened 2024-03-13
Add e2e tests

jesperengstrom:
### Summary of changes

* Adds some e2e tests for the index and /find routes. Mainly checking for the presence of important ui components and basic type/click interactions. Since the tests use real live data, none of them rely on the presence of any specific record, only that there _is_ data.
* Added `axe-core/playwright` for automated a11y tests on the routes (do `npm install`)

We could also add something like [playwright-lighthouse](https://github.com/abhinaba-ghosh/playwright-lighthouse) to make sure we stay above a certain performance threshold, for example.


johanbissemattsson (approved):
Awesome! 🌟 LGTM!

PR #976, opened 2024-03-15
Rework search card w. new lens

jesperengstrom:
### Tickets involved
[LWS-22](https://jira.kb.se/browse/LWS-22)

### Summary of changes
Iteration of search card given the design:
pippi
(minus year, checkbox, description, right side buttons)

* Created a new `web-card` lens that exclude many of the base card properties
* Rework the `SearchCard` component to support alternate renderings (`hasInstance`)
* New `truncate` option to `DecoratedData` that renders the first item and indicates the remainder (+ x). Not happy to grow it further, other suggestions how to achieve this are welcome :)
* New option to pass `data` directly to popover instead of fetching with id. Used here to show the remainder data (+ x) on hover. 
* Updated styles (typography variants)

### DIFF
--- a/lxl-web/src/colors.css
+++ b/lxl-web/src/colors.css
@@ -20,7 +20,7 @@ see https://tailwindcss.com/docs/customizing-colors#using-css-variables */
 	/* Text colors */
 	--text-primary: var(--color-primary);
 	--text-primary-inv: var(--color-white);
-	--text-link: var(--color-accent-dark);
+	--text-link: var(--color-primary);
 
 	/* Bg colors */
 	--bg-main: var(--color-subtle);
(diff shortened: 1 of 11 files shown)


olovy (approved):
LGTM

The card heading is blue in firefox. Other than that I have no comments on the implementation.

About the design: I'm not sure the vertical layout and boxed styling of the details works but this is a an improvement

johanbissemattsson (approved):
I’m also not particularly fond of the design/layout; and more specifically the truncation of contributors (especially when it’s already done after the first contributor). Multiple contributors can often be almost equally important in relation to each other and can also be useful for quickly differentiating between the results when searching.

I also don’t think that the search result items necessarily needs to be of equal height, or that it should be and end in itself. But If they need to be truncated, then I think it should preferably be done using CSS (e.g. with text-overflow: ellipsis`. maybe with -webkit-line-clamp` so it’s done after the second or third row.). The full list of contributors can then be shown in a popover / or using the title attribute (like e.g. Spotify does in their app).

But I think the implementation looks good, and think that it can be good to test this variant on the testing group / target audience.

jesperengstrom:
Not comfortable with merging something nobody likes. Closing.

johanbissemattsson:
Opening it and merging (even if we most likely will need to revisit it again soon).

PR #980, opened 2024-03-20
Wip/instances table

johanbissemattsson:
### Tickets involved
[LWS-31](https://jira.kb.se/browse/LWS-31)

### Solves

Adds a table of instances on the resource pages.

The implementation with the expandable rows works without Javascript (using search params and replaced history-state which enables saved state between page navigations). It also has basic keyboard functionality. Filtering by type and more elaborate styling is currently missing but shouldn't be too hard to implement.

Currently [jmespath](https://github.com/jmespath/jmespath.js) is used for querying and selecting the data in the columns but this should preferably be replaced with something more home-baked (with smaller bundle-size). Another alternative is to add more table-specifc lenses/formatting in the display.json file.

The `InstancesTable` component could also be refactored into a more generic  `DecoratedTable`/`DecoratedTreeGrid` component which can also be used on the search results page. But it is probably best saved for later...

### Summary of changes

- Add InstancesTable component


jesperengstrom:
LGTM!

(some styling remain as you say, as well as displaying the holding data in some meaningful way)

Some random thoughts:
If we follow this path (expansion state in url/built to work without js) we should probably use the same method with facets on the search result page for example (that will lead to new challenges: urls can grow really long and the api search query will have to be separated from ui state etc). The topic how much we should go out of our way to build for non-js users is worth discussing.

Another way to go about it could be to only allow one expanded instance at a time and use # anchors, allowing users to link directly to an instance on the product page (bibliotek.dk style). (But then you lose the ability to compare instances so, i dunno, it's a UX issue I guess)

jesperengstrom (approved):

johanbissemattsson:
> LGTM!
> 
> (some styling remain as you say, as well as displaying the holding data in some meaningful way)
> 
> Some random thoughts: If we follow this path (expansion state in url/built to work without js) we should probably use the same method with facets on the search result page for example (that will lead to new challenges: urls can grow really long and the api search query will have to be separated from ui state etc). The topic how much we should go out of our way to build for non-js users is worth discussing.
> 
> Another way to go about it could be to only allow one expanded instance at a time and use # anchors, allowing users to link directly to an instance on the product page (bibliotek.dk style). (But then you lose the ability to compare instances so, i dunno, it's a UX issue I guess)

Yeah, I think it sounds good to discuss it further! 

While I'm usally an advocate for almost always supporting non-js users (as long as the basic functionality works) but I'm not totally sure if this is the best implementation. An alternative that could also be worth exploring is opening the instance in an aside/pane to the right.

olovy (approved):
LGTM!

> If we follow this path (expansion state in url/built to work without js) we should probably use the same method with facets on the search result page for example (that will lead to new challenges: urls can grow really long and the api search query will have to be separated from ui state etc). The topic how much we should go out of our way to build for non-js users is worth discussing.

> Another way to go about it could be to only allow one expanded instance at a time and use # anchors, allowing users to link directly to an instance on the product page (bibliotek.dk style). (But then you lose the ability to compare instances so, i dunno, it's a UX issue I guess)

Perhaps this is a place for progressive enhancement?
* Expanding an instance sets the location/url to that instance
    * should be `@id` of that instance
* If javascript is enabled multiple instances can be kept expanded
* When there is only one instance it is expanded by default (usually an instance with a local work)


That's kind of how it works in search4. 
It is a bit subtle since the instance is just highlighted, not expanded.
Link to a work https://libris-dev2.kb.se/search4/1js6xhfpz7pvlwsw
Link to one instance of the same work: https://libris-dev2.kb.se/search4/s99l977zq37nbph8

PR #981, opened 2024-03-22
Publication year on search card

jesperengstrom:
### Tickets involved
[LWS-58](https://jira.kb.se/browse/LWS-58)

### Summary of changes

Displays instanceOf.publication.year (from-to) on search card (using the `jmespath` package to dig out the years).

Maybe these kinds of operations should be done in the backend, but i'm not entirely sure how to align it with the current data structure there 🤔 

Testing: search for 'sommar' against qa displays a few hits with their publication years indexed.

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -124,8 +124,8 @@
 							]
 						},
 						"contribution",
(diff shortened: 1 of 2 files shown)


olovy (approved):
LGTM

We shouldn't let i18n slip too much.

PR #983, opened 2024-03-26
Lxl-web cover images

lrosenstrom:
First iteration of cover images. So that we can show something tomorrow.

TODO: 
* `image` needs to be added to @reverse.instanceOf in hitlist (if we want to show instance images on work hits)
* ~~placement of image on product page~~
* ~~add instance images to instance table on product page~~
* ~~devops: handle auxd secret~~


lrosenstrom:
Would be cool to try this with QA data!

lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/+page.svelte
```diff
@@ -12,12 +13,21 @@
...
 			</h1>
 		</header>
-		<div class="overview mb-4">
-			<DecoratedData data={data.overview} block />
+		<div class="flex items-start">
+			<div>
+				<div
+					class="rounded-lg h-64 w-64 bg-contain bg-top bg-no-repeat"
+					style="background-image: url('{data.imageUris[0].imageUri}')"
+				/>
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/+page.svelte:21]: Could we use `` elements instead of using `background-image`? The former are generally better for accessibility and SEO.

A related question: is it easy to also get the width and height of the images? Width and height attributes on img elements are good to have as they prevent layout shifting when the page renders.

PR #984, opened 2024-03-26
Use _q parameter

jesperengstrom:
### Tickets involved
[LWS-40](https://jira.kb.se/browse/LWS-40)

### Solves

Uses the new `_q` parameter allowing advanced search queries.

### Summary of changes

* Use `_q` instead of `q` when searching
* Update `DisplayMappings` (search.ts) and `SearchMappings` (component) to support this.

TODO (maybe not now):
* ~~Build a nicer mapping interface~~
* Search for work by default (atm you need to ask for `type:work`)
* Hide certain params from the input field (you probably don't always want `type:work` there)
* Add labels for all properties in search mapping. Now displaying the `@id` in the absence of `labelByLang`.
* [uri #it ending issue](https://github.com/libris/librisxl/pull/1391) to be solved in FE?

### DIFF
--- a/lxl-web/src/colors.css
+++ b/lxl-web/src/colors.css
@@ -29,6 +29,7 @@ see https://tailwindcss.com/docs/customizing-colors#using-css-variables */
 	--bg-head: var(--color-accent-light);
 	--bg-positive: var(--color-accent-light);
 	--bg-positive-inv: var(--color-primary-light);
+	--bg-negative: var(--color-highlight);
 	--bg-cards: var(--color-white);
 	--bg-pill: var(--color-brown);
 }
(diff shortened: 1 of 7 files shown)


kwahlin (approved):
I've tested this and I think it solves the main issue, namely mapping the search input to the `_q` parameter and generate a visual, modifiable representation of the query and it's components.

More iterations to come of course.

The remaining questions/TODOs are also noted in https://github.com/libris/librisxl/pull/1391 and will be handled.

olovy (approved):
:ship:

kwahlin (approved):
LGTM

PR #988, opened 2024-04-05
Use details element for instances list

johanbissemattsson:
### Tickets involved
[LWS-31](https://jira.kb.se/browse/LWS-31)

### Solves

Uses [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) elements for the instances list, allowing native support for opening/closing the list rows without Javascript. It replaces the `InstancesTable` component https://github.com/libris/lxlviewer/pull/980.

The opened/closed state of the details are saved locally using the relatively new [Shallow routing](https://kit.svelte.dev/docs/shallow-routing) feature in SvelteKit, so the state will be saved if the user navigates forward and back in the history.

However, the state will not be applied if the user reloads the page (or returns from another document). See: https://kit.svelte.dev/docs/shallow-routing#caveats

Links for copying instance permalinks are also added. Anchor links will automatically be opened initially.

### Summary of changes

- Use details element for instances list
- Update label styling


olovy (changes requested):
* The labels should not be bold. We want to be able to scan the data, not the labels.
* There should be some vertical space between the properties?
* For now we are using edition/utgåva in the hit list so we should use the same terminology here: "Copy link to instance" -> "Copy link to edition"

Other than that, LGTM!

olovy:
The column layout is not responsive. Is that a later concern?

johanbissemattsson:
I have now done some minor styling adjustments and added basic responsiveness (my initial plan was to save it for a bigger design overhaul but I guess it's good to not wait to long with it) :)

jesperengstrom (approved):
LGTM!

Small note: the class `.text-subtle` used for labels is called `.text-secondary` throughout figma and is already present (and used) in the theme. 

We should really add this reworked instances list to figma, so that we have a common point of reference and can iterate from there...

johanbissemattsson:
> LGTM!
> 
> Small note: the class `.text-subtle` used for labels is called `.text-secondary` throughout figma and is already present (and used) in the theme.
> 
> We should really add this reworked instances list to figma, so that we have a common point of reference and can iterate from there...

Yeah sound good to go through the tailwind theming/styling, maybe next week?

I also have some gripes that I want to discuss, which makes the process of doing responsive styling cumbersome (I think `prettier-plugin-tailwindcss` can be at fault here...)

PR #991, opened 2024-04-11
LWS-64 show relations

jesperengstrom:
### Tickets involved
[LWS-64](https://jira.kb.se/browse/LWS-64)

### Solves

Fixes the "dead end" on product pages by displaying linked resources.

### Summary of changes

* Performs an `o` search for resources with only one instance (otherwise show instance list)
* Display the search result on the product page.

Since the /find data fetching is now needed on both the /find and /fnurgel routes, it has been moved to the layout level in a new group route (find). The /fnurgel page is now slot content for this layout. 
But because searching for relations is conditioned on the resource data (number of instances, its `@id` etc), the /fnurgel data needs to be fetched first. This means all data fetching is now performed in (find)/+layout.server

TODO:
* ~~Fix merge conflicts~~
* Stream in the search result on product pages when navigating client side (make it work...)
* Modify the facet and 'up' links so that they point to the same route
* Hide (some of) the search mapping pills
* Fix 1 instance works showing both instance list & search result (no hits)
* Fix failing test
* Fix styling


lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts
```diff
@@ -0,0 +1,136 @@
...
+
+		const overview = displayUtil.lensAndFormat(mainEntity, LxlLens.PageOverView, locale);
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const [_, overviewWithoutHasInstance] = pickProperty(overview, ['hasInstance']);
+
+		// TODO: Replace with a custom getProperty method (similar to pickProperty)
+		const instances = jmespath.search(overview, '*[].hasInstance[]');
+
+		// set condition to perform search
+		shouldFindRelations = instances.length <= 1; // correct??
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts:47]: Does this condition cover all cases where we want to `o` search?

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts
```diff
@@ -0,0 +1,136 @@
...
+
+		if (!resourceRes.ok) {
+			const err = (await resourceRes.json()) as apiError;
+			// Todo better error handling
+			throw error(err.status_code, err.status);
+		}
+
+		const resource = await resourceRes.json();
+		const mainEntity = centerOnWork(resource['mainEntity'] as FramedData);
+		resourceId = resource.mainEntity['@id']; // can we rely on this???
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts:37]: Getting the id for the `o` search like this, is it safe?

Comment (olovy) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts]: Yes!

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts
Comment (olovy) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts]: Yes.
Actually, we could remove the condition completely.
Since we are only searching for `Work` at the moment.

Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts]: But while we have the instances list, searching when there are  > 1 instances is unnecessary?

Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/+layout.server.ts]: But still, the condition prevents the fetch from even happening, which is a good thing. I suggest we leave it.

johanbissemattsson (approved):
Awesome! LGTM! ⭐

PR #995, opened 2024-04-16
LWS-63: Add holdings on resource page

johanbissemattsson:
### Tickets involved
[LWS-63](https://jira.kb.se/browse/LWS-63)

### Solves
- Adds support for a holdings modal (showing where the resource can be found) on the resource page.

### Summary of changes
- Add holdings modal on resource page
- Add ResourceImage component to prevent code duplication

### DIFF
--- a/lxl-web/src/lib/components/Modal.svelte
+++ b/lxl-web/src/lib/components/Modal.svelte
@@ -1,0 +1,92 @@
+<svelte:options accessors />
+
+<script lang="ts">
+	import { onMount } from 'svelte';
+	import { fly } from 'svelte/transition';
+	import { cubicInOut } from 'svelte/easing';
+
+	export let dialog: HTMLDialogElement | undefined = undefined;
+	export let close: ((event: Event) => void) | undefined = undefined;
+
+	let prevBodyOverflow: string | undefined = undefined;
+
+	onMount(() => {
+		disableBodyScroll();
+	});
+
+	function handleClose(event: MouseEvent | Event) {
+		// Use close method from prop if available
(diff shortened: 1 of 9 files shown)


olovy (approved):
A nice start!

I managed to get to a state where modal was closed and page not scrollable. Don't know how.

Implementation LGTM.

TODO of course:
Clicking on e.g. "Print (X libraries)" should display the correct list of libraries.
the URL could be `?holdings=Print`

jesperengstrom (approved):
Nice! ⭐ 

I only noticed there's no out animation. Maybe intentional, but when you click on the backdrop, the background color disappears, the modal is stuck for a sec, then goes away. Also the black text here is found nowhere else in the app. 

(I'm fine with leaving both for now)

johanbissemattsson:
> A nice start!
> 
> I managed to get to a state where modal was closed and page not scrollable. Don't know how.
> 
> Implementation LGTM.
> 
> TODO of course: Clicking on e.g. "Print (X libraries)" should display the correct list of libraries. the URL could be `?holdings=Print`

Yeah, noticed it myself now. Will do a follow-up fix so it always works as intended, there currently seems to be some issue with the saved page state.

> Nice! ⭐
> 
> I only noticed there's no out animation. Maybe intentional, but when you click on the backdrop, the background color disappears, the modal is stuck for a sec, then goes away. Also the black text here is found nowhere else in the app.
> 
> (I'm fine with leaving both for now)

Not intended and will probably need a workaround as Firefox doesn't support transitions on ::backdrops.

PR #1000, opened 2024-04-18
Display _i in search field

jesperengstrom:
### Solves

Search field being populated with stuff user never typed (ex * language:"https://id.kb.se/language/swe") when you use filters.

### Summary of changes

Easiest possible solution: display `_i` in search field if it exists, else use `_q`.
My suggestion is we use this as a starting point to investigate the limitations/need for an `_x=advanced` state etc.

Known quirks: 
* search for` yearPublished>2010` 
*  query remains as `_i` is still not present (the search query is not synced back from the api response - maybe is should, but that would result in re-fetching of data)
* use a filter on the side or paginate
* The query now disappears, i.e "becomes mapping pills" because `_i` now exists.

### DIFF
--- a/lxl-web/src/lib/components/Search.svelte
+++ b/lxl-web/src/lib/components/Search.svelte
@@ -7,16 +7,18 @@
 	export let placeholder: string;
 	export let autofocus: boolean = false;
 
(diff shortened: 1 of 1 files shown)


kwahlin (approved):
:ship:

PR #1002, opened 2024-04-19
Feature/ i param 2

jesperengstrom:
### Solves

* Enables user to see/edit their advanced search query by toggling the `_q` / `_i` params in search field.

### Summary of changes

* Use fetch `reddirect: manual` option to intercept api redirects. Then redirect to the same url in the frontend (in order to get the correct `_i`/`_q` combination).
* Display an edit button next to the search mapping pills _when `_i` and `_q` differ_ (i.e there is an active filter/advanced search query that can be shown in the input field). This felt like a clearer way to express this functionality than to have a 'simple/advanced search' toggle. 

**test with dev api**

~~backend fixes needed for this to work:~~
*  ~~`_x` param accepted and returned from the api (or this state will always be reset in the redirect operation)~~
*  ~~double url encoding fixed (or the url `searchParams` and the response `@id` will never match when using facet links,  resulting in a redirect loop)~~

### DIFF
--- a/lxl-web/src/lib/components/DecoratedData.svelte
+++ b/lxl-web/src/lib/components/DecoratedData.svelte
@@ -14,13 +14,7 @@
 	export let block = false;
 	export let truncate = false;
 	export let key: ResourceData | string = data;
-
-	// truncate option; use only first item as data and keep the rest for tooltip
-	let remainder: ResourceData[] | undefined;
-	if (truncate && Array.isArray(data) && data?.[0] && '@type' in data[0]) {
-		[data, ...remainder] = data;
-		truncate = false;
-	}
+	export let remainder: ResourceData | undefined = undefined;
 
 	const hiddenProperties = [
 		'@context',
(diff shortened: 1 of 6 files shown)


olovy (approved):
LGTM!

johanbissemattsson:
LGTM!

lxl-web/src/lib/components/DecoratedData.svelte
```diff
@@ -130,16 +124,30 @@
 {#key key}
 	{#if data && typeof data === 'object'}
 		{#if Array.isArray(data)}
-			{#each data as arrayItem}
+			{#if truncate && depth === 1 && data.length > 1}
+				<!-- truncate option; use only first item as data and keep the remainder for tooltip -->
+				{@const [first, ...remainder] = data}
```
jesperengstrom, line 129: @johanbissemattsson if you want, take a look at the changes made to `decoratedData` to avoid strange truncate bug. This feels a lot better, not mutating the `data`...

johanbissemattsson: This looks much better! 🙌

johanbissemattsson (approved):

PR #1004, opened 2024-04-23
Add more sort options

olovy:
## Description
Add more sort options to search component.

### Tickets involved
[LWS-113](https://jira.kb.se/browse/LWS-113)

### Summary of changes
* Add sort options
* Translate sort options

### Not in scope
* Handle sorting of different types correctly. For example the publication year option only works on Works.
* Actual publication year data includes lots of 0000 and 9999. They now come first for oldest and newest, respectively. They should be filtered out by backend.

### DIFF
--- a/lxl-web/src/lib/i18n/locales/en.js
+++ b/lxl-web/src/lib/i18n/locales/en.js
@@ -36,6 +36,14 @@ export default {
 		intendedAudience: 'Intended audience'
 	},
 	search: {},
+	sort: {
+		relevancy: 'Relevancy',
+		alphaAsc: 'A-Z',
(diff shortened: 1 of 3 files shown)


johanbissemattsson (approved):
LGTM!

jesperengstrom (approved):
Nice!

The current sort value doesn't seem to be updating on resource pages, but that's a separate thing.

🚢

olovy:
> Nice!
> 
> The current sort value doesn't seem to be updating on resource pages, but that's a separate thing.
> 
> 🚢

Yes, I noticed that. But decided it was a separate thing. Forgot to mention it here.

PR #1006, opened 2024-04-23
Add input field for filtering all filters

olovy:
## Description
Replace individual input fields in each facet group with a single input field for filtering all filters in all facet groups

### Tickets involved
[LWS-67](https://jira.kb.se/browse/LWS-67)

### Summary of changes
* Hide facet groups without matches
* Slight improvement in matching. Get hits for any word in filter. Separated by space or "--" (e.g "Sverige--Stockholm")
* Add translation for "find filter"
* It would be a bit trickier to both keep track of expanded state outside of "filtered filters" context and let the user collapse facet groups while filtering them. Now they can't be minimized while filtering and get a placeholder icon that keeps labels from jumping left instead...
    * We should also take care of facet groups being expanded by default when searching before fixing this?

### Not in scope
* Remember facet group state between pages
* Fuzzy matching of filters etc.


lxl-web/src/routes/(app)/[[lang=lang]]/(find)/FacetGroup.svelte
```diff
@@ -4,37 +4,48 @@
...
 		class="w-full text-left font-bold"
 		data-testid="facet-toggle"
 	>
 		<span class="flex items-center gap-2">
 			<span>
-				{#if expanded}
+				{#if searchPhrase}
+					<!-- FIXME -->
+					<BiDiamond class="text-icon-default" />
+				{:else if expanded}
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/FacetGroup.svelte:48]: I would prefer skipping the diamond icon for now (and instead just show the chevron down icon).

It's not the scope of this PR but I think we should soon rework the facet group list to use native [`<details>` ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) elements instead of having buttons controlling the expanded javascript state.

johanbissemattsson (approved):

jesperengstrom (approved):
Nice! ⭐

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/FacetGroup.svelte
Comment (olovy) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/FacetGroup.svelte]: Ok, replaced.
It feels a bit broken though since it is normally clickable and you still get the hand icon but now it is not clickable.

PR #1007, opened 2024-04-24
LWS-63: Fix holders by type

johanbissemattsson:
### Tickets involved
[LWS-63](https://jira.kb.se/browse/LWS-63)

### Solves

Fixes the list of unique holders (e.g. libraries) by instance type.

### Out of scope
The `+layout.server.ts` file has started growing so it would be preferable to separate it into different `+page.server.ts` files again for the different routes. We also need to fix the missing Typescript types for the resource pages...

### Summary of changes

- Add isFnurgel util method
- Fix unique holders by type
- Add count in holder list title

### DIFF
--- a/lxl-web/src/lib/utils/isFnurgel.test.ts
+++ b/lxl-web/src/lib/utils/isFnurgel.test.ts
@@ -1,0 +1,11 @@
+import { describe, it, expect } from 'vitest';
+import isFnurgel from './isFnurgel';
+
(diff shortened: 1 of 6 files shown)


olovy (approved):
Seems to work fine. 

We have to review what to show in the top of the holdings panel when all holdings for an instance type is shown. Just displaying the latest instance without any context is confusing. Probably just some data from the Work + the instance type?

PR #1009, opened 2024-04-24
Feature/instancelist chevrons

lrosenstrom:
### Tickets involved
[LWS-124](https://jira.kb.se/browse/LWS-124)

### Solves

Add chevrons for entries in instance list.

TODO: add chevron rotate animation in facet list.

### DIFF
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte
@@ -118,10 +118,10 @@
 						on:toggle={() => handleToggleDetails($page.state)}
 					>
 						<summary
-							class="flex min-h-11 gap-2 pl-1 align-middle hover:bg-pill/16"
+							class="flex min-h-11 items-center gap-2 pl-1 align-middle hover:bg-pill/16"
 							on:keydown={handleSummaryKeydown}
 						>
-							<span class="arrow flex items-center">
+							<span class="arrow">
 								<BiChevronRight />
 							</span>
 							{#each columns as columnItem}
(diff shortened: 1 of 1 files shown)


lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte
```diff
@@ -110,18 +112,18 @@
...
+							$page.url.searchParams.getAll('expanded').includes(id) ||
+							data.length === 1}
 						on:toggle={() => handleToggleDetails($page.state)}
 					>
 						<summary
-							class="flex min-h-11 gap-2 pl-2 align-middle hover:bg-pill/16"
+							class="flex min-h-11 gap-2 pl-1 align-middle hover:bg-pill/16"
 							on:keydown={handleSummaryKeydown}
 						>
+							<span class="arrow flex items-center pr-2">
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte:124]: I suggest removing the `pr-2` on the arrow (and adding space elsewhere), it becomes a padding-bottom when it rotates and results in a skewed rotation (not around centre)

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte
```diff
@@ -212,4 +214,8 @@
 	details[open] > summary {
 		@apply bg-pill/8 hover:bg-pill/16;
 	}
+
+	details[open] > summary .arrow {
+		@apply rotate-90 transition-transform;
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte:219]: this removes the transition class when the panel's not open, resulting in no closing animation.
suggestion:
```
.arrow {
  @apply rotate-0 transition-transform;
}

[open] > .arrow {
  @apply rotate-90;
}
```

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte
Comment (johanbissemattsson) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/[fnurgel=fnurgel]/InstancesList.svelte:124]: I would also add `w-fit` and `h-fit` so the span size will be equal to the size of the SVG. See https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content

`items-center` can then be moved to the parent `summary` element.

PR #1012, opened 2024-04-25
Add icons for Work types

olovy:
* Different symbols for different types in placeholder cover image
* Small type icon in the corner if there is no cover 
    * Except for type Text which is the "default"
* We will need the type spelled out as text as well. This just adds icons
* Approximate mapping of font awesome icons used in Libris katalogisering
* Will of course be obsoleted by type normalization

## TODO
* ~Breaks the placeholder image layout in search result list...~
* ~Small icon on top of cover image is not perfectly placed...~
* Use `ResourceImage` in `SearchCard`?
* alt-text

### Not in scope
* Icons in facets

### Tickets involved
[LWS-129](https://jira.kb.se/browse/LWS-129)


lrosenstrom (approved):
LGTM

PR #1013, opened 2024-04-25
LWS-68: Add year range to facets

jesperengstrom:
### Tickets involved
[LWS-68](https://jira.kb.se/browse/LWS-68)

### Solves

Add year range input to publication year facets

### Summary of changes

* Pass on the new optional ['search' key](https://github.com/libris/librisxl/pull/1419) from api to frontend.
* If it exists for a facet group, render new component `FacetRange` and display its values.
* input values are placed in the provided template link to perform a new range search.


olovy (approved):
Very nice!
Works well. :ship: 

Maybe it would be nice if the placeholder value was the min/max facet value pf the result set.

johanbissemattsson (approved):
LGTM! 🙌

jesperengstrom:
> Maybe it would be nice if the placeholder value was the min/max facet value of the result set.

Would be a nice feature, but would need some additional work (this component does not have access to the data/slices atm). A ticket maybe?

olovy:
> > Maybe it would be nice if the placeholder value was the min/max facet value of the result set.
> 
> Would be a nice feature, but would need some additional work (this component does not have access to the data/slices atm). A ticket maybe?

Yes definitely not for now.

PR #1015, opened 2024-04-26
LWS-147: Fix instance page sort order

jesperengstrom:
### Tickets involved
[LWS-147](https://jira.kb.se/browse/LWS-147)

### Solves

Fixes correct `selected` value for the sort select on instance page

**BUT**... the sorting is still not being done correctly on instance pages despite us sending the correct `_sort` param it seems. Maybe @olovy or @kwahlin can look into it?

### Summary of changes

* make `sortOrder` reactive to changes
* Style `select` somewhat
* add translation

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -46,6 +46,10 @@
 		@apply text-ellipsis rounded-md border-t border-t-primary/16 px-4 py-2 text-3-regular;
 	}
 
+	select {
+		@apply rounded-sm bg-white p-1;
+	}
+
 	.container-fluid {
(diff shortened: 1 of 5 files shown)


olovy:
> **BUT**... the sorting is still not being done correctly on instance pages despite us sending the correct `_sort` param it seems. Maybe @olovy or @kwahlin can look into it?
>

Yes it got broken by "deterministic order for wildcard search". Fixed now

olovy:
Alignment is off in Firefox. 
But opposite in Chromium if padding removed.

bild

olovy (approved):
LGTM!

See comment about text alignment.
Merging this so that sorting works

PR #1017, opened 2024-05-02
LWS-123: responsive filters

johanbissemattsson:
### Tickets involved
[LWS-123](https://jira.kb.se/browse/LWS-123)

### Solves

Adds support for showing and selecting filters on smaller screens.

TODO:
- Shallow routing / usage of `pushState` so back navigation closes the modal after opening it. This however opens a can of worms regarding what should happen when multiple filters are applied before closing the modal. We need to discuss this before implementing it.
- Horizontal scrolling of selected filters? The selected filters should probably not be wrapped as the screen real estate is really limited on mobiles. Figma sketches to discuss around would be good here.
- A _"Show X results"_ button (placed at the bottom of the modal?)
- Better no-js fallback styling (with added close button using an anchor link).
- The `+layout.svelte` has grown quite alot lately so we should probably separate the search functionality into a component, which can be reused instead. The same also applies to `+layout.server.ts` which has become a little bit unwieldy.
 
### Summary of changes

- Add support for choosing modal position
- Localize strings
- Add utility class for `ghost-btn`

### DIFF
--- a/lxl-web/src/lib/components/Modal.svelte
+++ b/lxl-web/src/lib/components/Modal.svelte
@@ -7,12 +7,15 @@
 	import { browser } from '$app/environment';
 	import { page } from '$app/stores';
 	import IconClose from '~icons/bi/x-lg';
+	import { setModalContext } from '$lib/contexts/modal';
 
 	export let dialog: HTMLDialogElement | undefined = undefined;
 	export let close: ((event: Event) => void) | undefined = undefined;
 	export let position: 'left' | 'right' = 'right';
 
 	let prevBodyOverflow: string | undefined = undefined;
+
+	setModalContext();
 
 	onMount(() => {
 		dialog?.showModal();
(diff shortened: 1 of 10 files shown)


olovy:
Nice!

One small bug: The hidden filter "Freetext: *" counts towards the number of filters so that the button shows (1) when no filter are selected.

jesperengstrom:
> Nice!
> 
> One small bug: The hidden filter "Freetext: *" counts towards the number of filters so that the button shows (1) when no filter are selected.

Yep, we went for the easiest solution to hide the 'Freetext *' with plain css. But it has some downsides - also the 'clear' button should appear after 2 or more filters, but now appears after one (since there's _actually_ 2), really redundant. Maybe we should wipe the mappings we don't want do display back in search.ts instead...

jesperengstrom (approved):
Looks good now I think!

PR #1018, opened 2024-05-03
LWS-119: Responsive search card

jesperengstrom:
### Tickets involved
[LWS-154](https://kbse.atlassian.net/browse/LWS-154)

### Solves

Search cards for smaller screens

### Summary of changes

Did not exhaust myself in reworking this, since I have a feeling search cards will be redesigned again sometime soon. But at least it looks ok now in mobile imo. 

* Under `sm`: Removed 'language' field, no 'pill' styling, tighter gaps; as in design
* I am however allowing contributor & years to have one row each (then line-clamped), I don't think it's possible to fit the card body into one single row.

* Fixed the empty 'ghost' bubbles that appear on cards (instances but no years)
* Added translations

### DIFF
--- a/lxl-web/src/lib/components/DecoratedData.svelte
+++ b/lxl-web/src/lib/components/DecoratedData.svelte
@@ -241,7 +241,7 @@
 	}
 
 	.remainder {
-		@apply ml-2 rounded-full bg-pill/8 px-2 py-1 text-secondary;
+		@apply ml-2 whitespace-nowrap rounded-full bg-pill/8 px-2 py-0.5 text-secondary;
 	}
 
 	.block {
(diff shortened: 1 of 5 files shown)


olovy (approved):
LGTM!

I agree this is a good start since we have to improve the design of the larger search cards as well.

Pet peeve: personally I think that the cards should extend all the way to sides of the screen on mobile.

Thoughts for later:
* image makes more sense on the right side on small screens?
* Align the sort control to the right?
* Use an icon instead of spelling out "sort by"? https://icons.getbootstrap.com/icons/sort-down/

jesperengstrom:
> Pet peeve: personally I think that the cards should extend all the way to sides of the screen on mobile.

Yes why not, that would save some space. Let's try it when @johanbissemattsson PR is merged (the layout markup has been changed there, I think)

johanbissemattsson:
LGTM and I also think it's good to keep it simple for now as it will probably be revamped soon.

My only reservation is the removal of the language field (I'm generally think it's good to be extra careful/restrained when removing content all-together on smaller screens).

johanbissemattsson (approved):

PR #1019, opened 2024-05-03
LWS-128: Better error handling when searching

johanbissemattsson:
### Tickets involved
[LWS-128](https://kbse.atlassian.net/browse/LWS-128)

### Solves

Better error handling when searching + 404 page with basic styling.

### Summary of changes

- Add error page
- Add not found error page
- Localize error pages
- Throw error if resource isn't found
- Add e2e tests


lxl-web/src/lib/components/Error.svelte
```diff
@@ -0,0 +1,76 @@
+<script>
+	import { page } from '$app/stores';
+	import { getSupportedLocale } from '$lib/i18n/locales';
+	import getPageTitle from '$lib/utils/getPageTitle';
+	export let showHeader = false;
+
+	const locale = getSupportedLocale($page.url.pathname.split('/')[1]);
+
+	/** Currently using hard-coded localized strings here as we cannot access the translator/t function in $page.data from the root error page)... there probably is a much cleaner way to do this? */
```
jesperengstrom, line 9: The only way I can think of is importing the translator function directly and initializing it:

```
import { getTranslator } from '$lib/i18n';
const translatorPromise = getTranslator(locale)
```

Because the function returns a promise, you have to wait for it in the template:

```
{#await translatorPromise}
{:then t}
   <span>{t('general.year')}</span>
{/await}
```
Seems to be working...

johanbissemattsson: I found an alternative way that doesn't involve async functions client-side.

If we move the i18n related `load` function to the root `+layout.ts` file then all error pages should be able to access `$page.data` (with the relevant localized data).

Added it in https://github.com/libris/lxlviewer/pull/1019/commits/19905838087f304beb0d48d95da2884740306b19

jesperengstrom: Heh, yes that was a simpler solution 😄

jesperengstrom: Now getting an inception-like situation that wasn't there before. Reproduce:
Apply a language filter. Click 'edit' btn. Change language to langage or something. Search.

Reloading the page displays the correct view for a short while.

johanbissemattsson: Oh noes! Will look into it!

johanbissemattsson: https://github.com/libris/lxlviewer/pull/1019/commits/cff90ad6a2b74106f6bd55c2701d09058ee19fc4 should resolve the issue (it now only uses the new `search` string part of the API response and keeps the current pages' pathname).

jesperengstrom (approved):
Nice! 🚢

PR #1021, opened 2024-05-06
added style for pagination

Katzlerjulia:
## Checklist:

### Tickets involved
[LXL-NNNN](https://jira.kb.se/browse/LXL-NNNN), [LXL-NNNN](https://jira.kb.se/browse/LXL-NNNN)

### Solves

Pagination is now styled
### Summary of changes

Summary of changes

added global css for buttons, and then the style of the buttons was applied on pagination

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -53,14 +53,14 @@
 	}
 
 	.button-primary {
-		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4  py-2 text-primary-inv text-4-cond-bold gradient-primary hover:gradient-secondary;
+		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4  py-2 text-primary-inv text-3-cond-bold gradient-primary hover:text-primary-inv hover:gradient-secondary;
 		box-shadow:
 			0px 1px 0px 0px rgba(82, 51, 20, 0.16),
 			0px 1px 0px 0px,
 			rgba(239, 214, 189, 0.24) inset;
 	}
 	.button-ghost {
-		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4 py-2 text-primary text-4-cond-bold;
+		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4 py-2 text-primary text-3-cond-bold;
 		border: 2px solid rgba(82, 51, 20, 0.16);
 		color: rgba(82, 51, 20, 0.8);
 	}
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -51,4 +51,22 @@
 	.container-fluid {
 		@apply mx-auto px-4 sm:px-8;
 	}
+
+	.button-primary {
+		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4  py-2 text-primary-inv text-4-cond-bold gradient-primary hover:gradient-secondary;
+		box-shadow:
+			0px 1px 0px 0px rgba(82, 51, 20, 0.16),
+			0px 1px 0px 0px,
+			rgba(239, 214, 189, 0.24) inset;
+	}
+	.button-ghost {
+		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4 py-2 text-primary text-4-cond-bold;
+		border: 2px solid rgba(82, 51, 20, 0.16);
+		color: rgba(82, 51, 20, 0.8);
+	}
+	.button-ghost:hover {
+		border: 2px solid rgba(66, 107, 84, 0.32);
+		background-color: rgba(213, 229, 223, 0.4);
+		color: #426b54;
+	}
 }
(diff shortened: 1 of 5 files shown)


lxl-web/src/app.css
johanbissemattsson, line 1: Sometimes I get a warning saying:
>  "The `button-primary` class does not exist. If `button-primary` is a custom class, make sure it is defined within a `@layer` directive."

I have to admit that I am far from an expert regarding Tailwind (so maybe @jesperengstrom has better input here?) but moving the block to `tailwind.config.ts` seems to resolve the issue for me (had to rewrite it a little bit to work):

```JS
'.button-primary': {
	'@apply flex h-10 items-center justify-center gap-2 rounded-md px-4  py-2 text-primary-inv text-4-cond-bold gradient-primary hover:gradient-secondary':
		{},
	'box-shadow':
		'0px 1px 0px 0px rgba(82, 51, 20, 0.16), 0px 1px 0px 0px, rgba(239, 214, 189, 0.24) inset;'
},
'.button-ghost': {
	'@apply flex h-10 items-center justify-center gap-2 rounded-md px-4 py-2 text-primary text-4-cond-bold':
		{},
	border: '2px solid rgba(82, 51, 20, 0.16)',
	color: 'rgba(82, 51, 20, 0.8)'
},
'.button-ghost:hover': {
	border: '2px solid rgba(66, 107, 84, 0.32)',
	'background-color': 'rgba(213, 229, 223, 0.4)',
	color: '#426b54'
}
```

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte
```diff
@@ -57,16 +64,24 @@
...
+					<a class="button-ghost" href={relativizeUrl(next?.['@id'])} aria-label="Nästa sida"
+						><BiChevronRight class="text-icon"></BiChevronRight></a
+					>
+				</li>
 			{/if}
 		</ul>
 	</nav>
 {/if}
 
 <style>
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte:79]: Adding `<style lang="postcss">`should remove the `Unknown at rule @apply` warning.

johanbissemattsson (commented):
Looks good to me! ⭐🙌

If you have the time to add localized aria-strings it would also be super appreciated! See for example https://github.com/libris/lxlviewer/blob/feature/style-pagination/lxl-web/src/lib/components/Modal.svelte#L79&L32 how it's done (the strings also needs to be added to `src/lib/i18n/locales/en.js` and `src/lib/i18n/locales/sv.js`).

lxl-web/src/app.css
jesperengstrom: 🤔 Hmmm I haven't seen that warning, and so far I've not encountered any problems with putting classes in app.css...

The question of where they belong is not entirely clear. Adding a class to the theme using `addUtilities` enables autocomplete in the template code, which is really good for small custom utility classes used often (example: text styles i.e `text-3-cond-bold`) but the syntax required here is borderline unreadable imo.

So for classes with more css and less usage (like in this case - reusable 'component' styling if you will) I root for adding them to app.css.

lxl-web/src/app.css
```diff
@@ -49,4 +49,22 @@
...
+
+	.button-primary {
+		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4  py-2 text-primary-inv text-4-cond-bold gradient-primary hover:gradient-secondary;
+		box-shadow:
+			0px 1px 0px 0px rgba(82, 51, 20, 0.16),
+			0px 1px 0px 0px,
+			rgba(239, 214, 189, 0.24) inset;
+	}
+	.button-ghost {
+		@apply flex h-10 items-center justify-center gap-2 rounded-md px-4 py-2 text-primary text-4-cond-bold;
```
jesperengstrom, line 61: I think the text style used for buttons in the design is `text-3-cond-bold`, which is reasonable, it looks a bit too big now.

jesperengstrom: Also the dots (...)
should probably have the same text style and maybe aligned to the bottom (`items-end`)?

lxl-web/src/app.css
johanbissemattsson: Hmm strange if it only affects me!

See attached screenshot:
Skärmavbild 2024-05-07 kl  10 57 27

lxl-web/src/app.css
jesperengstrom: Other than that, lookin' good! :)

See if you can resolve the comments we made, then we'll fix the merge conflict together.

lxl-web/src/app.css
johanbissemattsson: https://tailwindcss.com/docs/functions-and-directives#using-apply-with-per-component-css

> The solution to this problem is to define any custom styles you want to @apply in your components using the [plugin system](https://tailwindcss.com/docs/plugins) instead:

and later... 

> Honestly though the best solution is to just not do weird stuff like this at all. Use Tailwind’s utilities directly in your markup the way they are intended to be used, and don’t abuse the @apply feature to do things like this and you will have a much better experience.

😬

jesperengstrom: Aaaaah there it is, I was looking in the console output.

Yes, you can't `@apply` own classes that are defined outside of tailwind. But we don't need to either.
We can just add them conditionally in the template:

`<a class={p === currentPage ? 'button-primary' : 'button-ghost'}>`

lxl-web/src/app.css
Katzlerjulia: Thank you I will try to resolve this :D

jesperengstrom (approved):
Ship it! 🚢

PR #1023, opened 2024-05-08
Fix filter pill space in Firefox

jesperengstrom:
### Solves

Using `inline` instead of `inline-block` for the labels seems to remove the weird filter pill space in Firefox. Can anyone confirm?


olovy (approved):
Yes! Works fine for me.

PR #1024, opened 2024-05-10
LWS-126: Add search button etc

jesperengstrom:
### Tickets involved
[LWS-126](https://kbse.atlassian.net/browse/LWS-126)

### Summary of changes

* Add a button for the search input
* Use new button classes on existing buttons
* Extend button classes with disabled/active state, negative variant etc.
* Tailwind-ify the CSS a bit
* Add focus outline for the search input

### DIFF
--- a/lxl-web/src/lib/components/Search.svelte
+++ b/lxl-web/src/lib/components/Search.svelte
@@ -3,6 +3,7 @@
 	import { afterNavigate } from '$app/navigation';
 	import getDefaultSearchParams from '$lib/utils/addDefaultSearchParams';
 	import getSortedSearchParams from '$lib/utils/getSortedSearchParams';
+	import BiSearch from '~icons/bi/search';
 
 	export let placeholder: string;
 	export let autofocus: boolean = false;
@@ -34,11 +35,11 @@
 	}
 </script>
 
-<form class="w-full" action="find" on:submit={handleSubmit}>
+<form class="relative w-full" action="find" on:submit={handleSubmit}>
 	<!-- svelte-ignore a11y-autofocus -->
 	<input
 		id="main-search"
-		class="h-12 w-full rounded-full border-t-accent-dark/32 text-secondary sm:h-14"
+		class="h-12 w-full rounded-full pr-12 text-secondary shadow-accent-dark/32 sm:h-14 sm:pr-28"
 		type="search"
 		name="_q"
 		{placeholder}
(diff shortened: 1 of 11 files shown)


lxl-web/src/lib/components/Search.svelte
```diff
@@ -34,11 +35,13 @@
...
 
-<form class="w-full" action="find" on:submit={handleSubmit}>
+<form class="relative w-full" action="find" on:submit={handleSubmit}>
 	<!-- svelte-ignore a11y-autofocus -->
 	<input
 		id="main-search"
-		class="h-12 w-full rounded-full border-t-accent-dark/32 text-secondary sm:h-14"
+		class="h-12 w-full rounded-full pr-12 text-secondary shadow-accent-dark/32 focus:shadow-search-focus focus:outline
+			focus:outline-8 focus:outline-accent-dark/8 sm:h-14
+			sm:pr-28"
```
johanbissemattsson, line 44: I would prefer the default browser outline styling as the outlines from the Figma sketches are extremely subtle. I feel that we loose more (accessiblity) than we gain from the custom styling.

Current custom outline (which I think is too subtle):
Skärmavbild 2024-05-13 kl  13 41 18

Default browser outline in Firefox:
Skärmavbild 2024-05-13 kl  13 41 01

johanbissemattsson (commented):
LGTM (apart from the subtle search input outline)

lxl-web/src/lib/components/Search.svelte
johanbissemattsson: I'm also ok with having a shadow outline if has more contrast / is more visible.

jesperengstrom: Skärmavbild 2024-05-13 kl  14 00 36

Revisiting the Figma design now I realize the outline is probably meant as hover state (similarly, the buttons also have hover outline). Maybe a bit unconventional for an input field. I'm ok with skipping it now and going for a general, fully visible/accessible focus ring later.

jesperengstrom: Turned the contrast up a notch. Visible enough?
Skärmavbild 2024-05-14 kl  10 51 11

johanbissemattsson: Much better, I'm fine with it :)

The only problem with adding a custom outline in one place is that missing customized outlines on other elements gets more glaring/noticeable. 😀 But that should not hinder us!

olovy (approved):
LGTM! :ship:

lxl-web/src/lib/components/Search.svelte
jesperengstrom: Yep, but that's the beta feel :)

PR #1025, opened 2024-05-13
LWS-171: Fix label visibility

johanbissemattsson:
### Tickets involved
[LWS-171](https://kbse.atlassian.net/browse/LWS-171)

### Solves

Fixes the visiblity of decorated data labels when navigating (I accidentally introduced the bug in https://github.com/libris/lxlviewer/commit/999dc779c859fe99a9712d108d5f531cfed707d1 when fixing duplicate content when closing modals). 

I have now added tests for both cases to catch eventual similar regressions in the future.

### Summary of changes

- Use data as value in key block
- Add tests
- Add workaround to remove duplicate content when closing modals


olovy (approved):
Tested. Seems to work fine!

:ship:

PR #1028, opened 2024-05-15
LWS-116: Info text & footer

jesperengstrom:
### Tickets involved
[LWS-116](https://kbse.atlassian.net/browse/LWS-116)

### Summary of changes

* Add 'beta' to service name
* remove slogan on landing page
* Add info text on landing page
* Add footer with some links


olovy (approved):
LGTM!

I think the footer should never drift above the bottom of the window but we can fix that later

bild

jesperengstrom:
> I think the footer should always stay at the bottom of the window but we can fix that later

Yikes! I had to zoom out to get that look.

johanbissemattsson:
> I think the footer should always stay at the bottom of the window but we can fix that later

This should be resolved in the latest commit.

johanbissemattsson (approved):

PR #1029, opened 2024-05-15
LWS-120: Add translations

jesperengstrom:
### Tickets involved
[LWS-120](https://kbse.atlassian.net/browse/LWS-120)

### Summary of changes

Went through all the components/pages in search of untranslated strings. Found only these, so I guess we're up to speed.


lrosenstrom:
Nice!

PR #1031, opened 2024-05-16
Add external link icon and make text smaller

lrosenstrom:
### Tickets involved
[LWS-141](https://kbse.atlassian.net/browse/LWS-141)

### Solves

Add icon to external links. Suggested form:

Screenshot from 2024-05-16 11-00-09

We could do this in a more general way by e.g. supporting icons in `fresnel:contentLast` etc, by mapping an icon name to a general Svelte component (icon) in a similar way as in` getTypeIcon.ts`  but not sure if it is worth it? 

### TODO:
* Popover: "Opens in new tab" ???

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -510,7 +510,7 @@
 			"@id": "MediaObject-format",
 			"@type": "fresnel:Format",
 			"fresnel:classFormatDomain": ["MediaObject"],
-			"fresnel:resourceStyle": ["ext-link", "uriToId()", "text-large", "block"]
+			"fresnel:resourceStyle": ["ext-link", "uriToId()"]
 		},
 		"Document-format": {
 			"@id": "Document-format",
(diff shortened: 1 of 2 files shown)


johanbissemattsson (approved):
LGTM! And I like how it works (no need to add a more general way to add icons, atleast for now).

Regarding the to-do: I think just adding a simple native `title` attribute is sufficient.

jesperengstrom (approved):
LGTM!

olovy:
Nice!

* There is no spacing between `MediaObject`s. The previous layout where they always had there own line (`block`) is more clear?
* I still think MediaObject links should be slightly heavier than other properties? We want a "call to action" for full text/image.

bild

lrosenstrom:
(1) I'll revisit the layout for several media objects!
(2) Maybe there is another way to emphasize the links... I think the large text looked a bit off.

lrosenstrom:
Couldn't figure a better way than [0c33432](https://github.com/libris/lxlviewer/pull/1031/commits/0c334324f6c3349c4b56bd35401ed986ee0bd723), which is kind of a hack, to add a row break after the icon only of _MediaObjects_. Other suggestions are welcome. 
It looks like this:
Screenshot from 2024-05-20 17-39-44

olovy (approved):
LGTM.

Works as intended.
We can improve the implementation later.

PR #1032, opened 2024-05-16
LWS-138: styling tweaks to filters

jesperengstrom:
### Tickets involved
[LWS-138](https://kbse.atlassian.net/browse/LWS-138)

### Summary of changes

Applied some additional styling to the content in the filter panel.
* Slightly more highlighted and repositioned totalItems for better readability (?)
* More visible toggle btn

Still doesn't look at all readable some cases, but if truncation isn't an option, we really should give the panel (back) some more width. 

Skärmavbild 2024-05-16 kl  16 26 56

### DIFF
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/(find)/FacetGroup.svelte
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/(find)/FacetGroup.svelte
@@ -59,11 +59,11 @@
 			<!-- facet range inputs; hide in filter search results -->
 			<FacetRange search={group.search} />
 		{/if}
-		<ol class="mt-2 max-h-[437px] overflow-y-auto" data-testid="facet-list">
+		<ol class="mt-2 max-h-[437px] overflow-y-auto overflow-x-clip" data-testid="facet-list">
 			{#each shownFacets as facet (facet.view['@id'])}
-				<li class="pl-6">
+				<li class="mb-[0.3rem]">
 					<a
-						class="facet-link flex items-end justify-between gap-2 no-underline"
+						class="facet-link flex items-end justify-between gap-2 pl-6 no-underline"
 						href={facet.view['@id']}
 					>
 						<span class="flex items-baseline">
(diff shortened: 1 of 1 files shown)


johanbissemattsson:
I think it looks nice with the background behind the hits count! Not a fan of the enormous _show more/show less_ buttons. I think it takes too much focus compared to the rest.

Some minor comments:
- More margin/padding between the facet items are needed. Not much more – just a teeny bit more so they are more distinctly separated.

- I'm inclided to favour not having any line-breaks in the facets together with `text-overflow: ellipsis;`; the ellipsis should preferebly be in the middle of the string as the most important separating parts is often in the beginning and at the end. This is sadly [not supported natively in CSS](https://github.com/w3c/csswg-drafts/issues/3937) so a work-around is needed. So it is probably best done in a later PR.

olovy:
* Agree that a teeny bit of margin between items could work for separating them. 
* Could skip the left padding of items and use the whole width? (Perhaps nicer with the chevron to the right then but that is a bigger change?). See e.g. finna.fi
* Agree that show more should be same size or smaller than the items

0.3 rem, no left padding:
bild

finna
Skärmbild från 2024-05-17 10-11-49

jesperengstrom:
Ok,
* Reverted the button
* Added separating margin between items

> Could skip the left padding of items and use the whole width?

I think we then lose the visual hierarchy that is really good for orientation?

> I'm inclided to favour not having any line-breaks in the facets together with text-overflow: ellipsis; 

Libris Kat has a neat way of ellipsing each part of the name individually. Can't do that now since it's a single string (used for searching), but maybe we can have it both ways later on somehow...

Skärmavbild 2024-05-17 kl  11 53 02

I still think the filter panel should be allowed 320px width as before (on bigger screens maybe), that would in most cases mean one line break instead of 2, which makes a big difference...

johanbissemattsson:
> Libris Kat has a neat way of ellipsing each part of the name individually. Can't do that now since it's a single string (used for searching), but maybe we can have it both ways later on somehow...

Ah that's also a nice solution!
 
> I still think the filter panel should be allowed 320px width as before (on bigger screens maybe), that would in most cases mean one line break instead of 2, which makes a big difference...

Feel free to make changes to https://github.com/libris/lxlviewer/blob/develop/lxl-web/tailwind.config.js#L117 but it's very much a balacing act  as it also affects what works and doesn't works in other places (e.g. the spacing between the resource images and the rest of the contents on the resource page). Alas, I think 320px is too wide for medium sized screens (see attached screenshot).

I think it's good to look at it together next design meeting wednesday!

Skärmavbild 2024-05-17 kl  13 53 04

johanbissemattsson:
Something like this could maybe work:
```
gridTemplateColumns: {
    find: 'minmax(min(240px, 100%), min(400px, 16.67%)) 5fr'
},
```

johanbissemattsson (approved):

olovy (approved):
:ship:

PR #1035, opened 2024-05-21
LWS-117: Add help page

johanbissemattsson:
### Tickets involved
[LWS-117](https://kbse.atlassian.net/browse/LWS-117)

### Solves

Adds a help page with content from https://kungliga-biblioteket.app.box.com/notes/1524440108812?s=vhly6db1m0vuzsgx03vog8paefsu18vv

### Summary of changes

- Add a help page
- Add [mdsvex](https://github.com/pngwn/MDsveX) (used for showing markdown content)
- Add MarkdownContent component with layout wrapping mdsvex document


jesperengstrom:
> could LibrisXL be used here?

I've heard that being suggested before. But could we use markdown or similar as input?

> I would prefer having more globally shared styles for simple typographic elements

How about just doing

```
p {
   @apply mb-2;
}
```

in app.css? And then manually override it in the app when needed?

jesperengstrom (approved):
LGTM!

johanbissemattsson:
The PR now uses [mdsvex](https://github.com/pngwn/MDsveX) to render markdown content on the help page (the localized markdown files (currently `en.md` and `sv.md` are placed together with the `+page.svelte` file.

Currently the content for both locales is imported in the `+page.svelte` for the sake of simplicity; it would however be preferable to use some kind of dynamic import/fetch so only the current language is sent to the client.

jesperengstrom (approved):
LGTM!
(I guess the english text is a todo)

Thought: would it be easier for non-dev editors of these files if they are located in a folder called Content or similar? There won't be an obvious place to look for the front page text for example...

johanbissemattsson:
Merging as-is for now; it can improved upon later (preferably with some kind of simple CMS editing capability)

PR #1040, opened 2024-05-22
Display pagination range. e.g. '11-20 of 42 hits'

olovy:
## Description
As designed:
bild

Example:
bild

### Tickets involved
[LWS-179](https://jira.kb.se/browse/LWS-179)


jesperengstrom (approved):
LGTM!

My only remark is that it's maybe a bit odd that the from/to numbers aren't localized/formatted when the total is.

PR #1042, opened 2024-05-23
Use totalItemsByRelation when possible

andersju:
Since https://github.com/libris/librisxl/pull/1439 we store the total number of incoming links per type in each ES document. This vue-client change uses that data (if present), avoiding unnecessary requests. Backwards-compatible with older indices.

For a typical cataloguing client search result page this reduces the number of `/find` requests by 20.


olovy (approved):
LGTM!

PR #1045, opened 2024-05-27
Feature/lws 165 add subject codes

lrosenstrom:
Show `inScheme` `code` for facets (sao, saogf, gmgpc/swe etc). The background is that `code` is not included in the response from the backend, hence this special handling. Also, there is currently no way in the Fresnel/`display-web.json`-logic to present something differently _only in the facets_. This may be fixed with sublenses in the future.

### TODO: 
* Some marc terms render an empty string (when missing` @type`?)

### DIFF
--- a/lxl-web/src/lib/utils/http.ts
+++ b/lxl-web/src/lib/utils/http.ts
@@ -24,5 +24,8 @@ export function getUriSlug(url: string | undefined) {
 	if (!url) {
 		return '';
 	}
-	return new URL(url).pathname.split('/').pop();
+	const urlArray = new URL(url).pathname.split('/');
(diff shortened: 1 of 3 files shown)


olovy (approved):
Nice! LGTM

lrosenstrom:
Regarding the TODO --- this seems to be a general issue affecting id.kb.se and libris.kb.se/katalogisering. See e.g. 
https://libris.kb.se/katalogisering/bb53hpwfd2kwx07c --> "• marc"
vs. 
https://libris.kb.se/katalogisering/b96vgpwfd0j3qgrg  --> ""
I will ignore this.

PR #1049, opened 2024-05-28
Feature/lws 152 redesigned search cards

johanbissemattsson:
### Tickets involved
[LWS-152](https://kbse.atlassian.net/browse/LWS-152)

### Solves

Redesign of the search card items showing more relevant information (e.g. more than one contributor) in semantically separated card sections (header, body, and footer).

The content is placed in separate sections using the new fresnel lenses `web-card-header-extra` and `web-card-footer` (together with the previously used derived lens`LxlLens.CardBody`).

**After:**
Skärmavbild 2024-05-28 kl  10 26 48

**Before:**
Skärmavbild 2024-05-28 kl  10 27 21

### Summary of changes

- Split search card information into sections using new lenses
- Add new search card layout using container queries for responsive design

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -110,9 +110,3 @@
 		@apply hover:bg-negative hover:text-secondary focus:bg-negative focus:text-secondary;
 	}
 }
-
-@layer utilities {
-	.search-card {
-		@apply flex  border-b border-b-primary/16 bg-cards p-4 md:rounded-md md:p-6;
-	}
-}
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -113,6 +113,6 @@
 
 @layer utilities {
 	.search-card {
-		@apply flex gap-4 overflow-hidden border-b border-b-primary/16 bg-cards p-4 sm:gap-8 md:rounded-md md:p-6;
+		@apply flex  border-b border-b-primary/16 bg-cards p-4 md:rounded-md md:p-6;
 	}
 }
--- a/lxl-web/src/lib/assets/img/placeholder.svg
+++ b/lxl-web/src/lib/assets/img/placeholder.svg
@@ -1,3 +1,3 @@
-<svg width="100" height="150" viewBox="0 0 100 150" fill="none" xmlns="http://www.w3.org/2000/svg">
-<rect width="100" height="150" fill="rgba(82, 51, 20, 0.04)"/>
+<svg width="100" height="100" viewBox="0 0 100 100" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect width="100" height="100" fill="rgba(82, 51, 20, 0.04)"/>
 </svg>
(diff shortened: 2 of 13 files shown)


lrosenstrom:
Looks good! It's worth pointing out that the "before"-version might look more vertically compact, but this is not the case when cover images are present:
Screenshot from 2024-05-28 11-20-02
So it's good that we utilize the empty space that is available.

jesperengstrom:
Skärmavbild 2024-05-28 kl  11 59 06
Skärmavbild 2024-05-28 kl  11 59 19

Any reason though for the radically smaller thumbs when there's plenty of room?

lrosenstrom (approved):
IMO make thumbnail same size as before, then LGTM!

johanbissemattsson:
> Skärmavbild 2024-05-28 kl 11 59 06 Skärmavbild 2024-05-28 kl 11 59 19
> 
> Any reason though for the radically smaller thumbs when there's plenty of room?

The main reason was that the images should have roughly equal layout height/width independently of the aspect ratio of the image (the images can be be of different heights and width but they must fit in an imaginary square with the size of the grid column).

Allowing tall images to fill the entire width of the grid column would enlargen the search card in cases when the text content is short.

We can of course continue to tweak this.

olovy (commented):
* The card heading has lost font-color. I think we should keep the old one.
* We could try to have a smaller top padding inside the cards. Can leave this kind of tweaking for later

olovy:
* Agree that the thumbs could be slightly larger
* Also agree that it is good that the thumbs don't grow the cards vertically.
    * See the case with only one line each of heading, body and footer
    * Maybe the images could extend below the text towards the bottom of the card without growing the card?
* I think the placement of thumbs looks slightly off. Try to center them inside the space (A)? 

bild

olovy:
> * Agree that the thumbs could be slightly larger
> 
>     * Also agree that it is good that the thumbs don't grow the cards vertically.
>       
>       * See the case with only one line each of heading, body and footer
>       * Maybe the images could extend below the text towards the bottom of the card without growing the card?
> 
>     * I think the placement of thumbs looks slightly off. Try to center them inside the space (A)?
> 
> 
> bild

Or rather just use the full width (A). Shrink it if thumbs get too tall.

johanbissemattsson:
> * The card heading has lost font-color. I think we should keep the old one.

This is fixed now.

johanbissemattsson:
I've made some minor adjustments regarding the image sizing (which I have a feeling will not be enough 😅).

I'm totally up for further tweaking it but I would rather keep it in a subsequent PR (I think we should be restrained in doing quick fixes regarding image sizes now as we have to account for all edge-cases regarding image aspect ratios and content lengths).

olovy:
Agree that we can improve the image sizing in a subsequent PR.
I think we should optimize a bit more for the "common book" case, i.e. portrait aspect ratio,

jesperengstrom (approved):

PR #1057, opened 2024-05-29
LWS-313: Yet another pagination bugfix

jesperengstrom:
### Tickets involved
[LWS-131](https://kbse.atlassian.net/browse/LWS-131)

### Solves

Paginating the relation search does not work in some cases.

Because: simply appending `_offset` to the link causes a new double-fetch (introduced with the predicate tabs) that ultimately uses the predicate `view` to get the data. That link excludes any applied `_offset`, fetching the first 10 hits again.

By using a backend-made link as template for all pagination links, we avoid the problem (and double fetch) since they include the `_p` param.

### Summary of changes

* Fix pagination links
* Tweak pagination template
* Mobile look tweaks

### DIFF
--- a/lxl-web/tailwind.config.js
+++ b/lxl-web/tailwind.config.js
@@ -129,7 +129,7 @@ export default {
 	},
 	plugins: [
 		/** @type {import('tailwindcss/types/config').PluginCreator} */
-		({ addUtilities }) => {
+		({ addUtilities, addVariant }) => {
 			addUtilities({
 				// typography 1-6 variants in design
 				'.text-1-regular': {
@@ -192,6 +192,7 @@ export default {
 					'@apply p-4 sm:px-6': {}
 				}
 			});
+			addVariant('mobile-only', "@media screen and (max-width: theme('screens.sm'))");
 		}
 	]
 };
--- a/lxl-web/tailwind.config.js
+++ b/lxl-web/tailwind.config.js
@@ -129,7 +129,7 @@ export default {
 	},
 	plugins: [
 		/** @type {import('tailwindcss/types/config').PluginCreator} */
-		({ addUtilities, addVariant }) => {
+		({ addUtilities }) => {
 			addUtilities({
 				// typography 1-6 variants in design
 				'.text-1-regular': {
@@ -192,7 +192,6 @@ export default {
 					'@apply p-4 sm:px-6': {}
 				}
 			});
-			addVariant('mobile-only', "@media screen and (max-width: theme('screens.sm'))");
 		}
 	]
 };
--- a/lxl-web/src/lib/components/find/Pagination.svelte
+++ b/lxl-web/src/lib/components/find/Pagination.svelte
@@ -33,7 +33,6 @@
 </script>
 
 {#if data.items.length > 0 && totalItems > itemsPerPage}
-	{console.log(pageSequence)}
 	<nav aria-label={$page.data.t('search.pagination')} data-testid="pagination">
 		<ul class="my-4 flex justify-center gap-1.5 sm:gap-2">
 			<!-- prev -->
--- a/lxl-web/src/lib/components/find/Pagination.svelte
+++ b/lxl-web/src/lib/components/find/Pagination.svelte
@@ -36,63 +36,58 @@
 
 {#if data.items.length > 0 && totalItems > itemsPerPage}
 	<nav aria-label="paginering" data-testid="pagination">
-		<ul class="my-4 flex justify-center">
-			<!-- prev and first -->
+		<ul class="my-4 flex justify-center gap-1.5 sm:gap-2">
+			<!-- prev -->
 			{#if !isFirstPage || itemOffset > 0}
 				<li>
 					<a
-						class="button-ghost mx-2 w-11 p-0"
+						class="button-ghost"
 						href={getOffsetLink(itemOffset - itemsPerPage)}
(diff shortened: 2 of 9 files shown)


lxl-web/tailwind.config.js
```diff
@@ -192,6 +192,7 @@ export default {
 					'@apply p-4 sm:px-6': {}
 				}
 			});
+			addVariant('mobile-only', "@media screen and (max-width: theme('screens.sm'))");
```
johanbissemattsson, line 195: Do we need an extra utility class for mobile-only when we already have the possiblity to have classes as mobile first together with `sm:classes-here-for-larger-breakpoints` for larger breakpoints?

olovy (approved):
LGTM!

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte
```diff
@@ -95,3 +95,9 @@
 		</ul>
 	</nav>
 {/if}
+
+<style lang="postcss">
+	nav li > a {
+		@apply mobile-only:h-8 mobile-only:px-2.5;
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte:101]: I think the buttons should be minimum size of 44px to adhere to: https://www.w3.org/WAI/WCAG21/Understanding/target-size.html.

I also prefer showing as many pages as possible (maybe with an upper limit of 10 instead of the current 3 in `sequenceSize`) so I would propose adding:
`
	nav li > a {
		@apply min-h-11 min-w-11 !px-2;
	}
` together with `flex-wrap` + `overflow-hidden` on a wrapping container around the pagination items (not the first and last as they should always be visible).

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte
```diff
@@ -30,16 +30,14 @@
...
 		let o = offset < 0 ? 0 : offset;
-		const params = new URLSearchParams($page.url.searchParams.toString());
-		params.set('_offset', o.toString());
-		return `${$page.url.pathname}?${params.toString()}`;
+		return `${first['@id']}&_offset=${o}`;
 	}
 </script>
 
 {#if data.items.length > 0 && totalItems > itemsPerPage}
 	<nav aria-label="paginering" data-testid="pagination">
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte:38]: Localized string for aria-label is needed here.

lxl-web/tailwind.config.js
johanbissemattsson: `max-sm:classes-here` is another already existing alternative https://tailwindcss.com/docs/responsive-design#targeting-a-single-breakpoint

johanbissemattsson:
There is still some reactivity-issues regarding the number of pagination items.

To reproduce:

1. Go to for example ttp://localhost:5173/find?_i=astrid&_q=astrid+hasInstanceType:Tactile&_limit=10
2. Add filters so there is few pages in the pagination (try adding the subject Vardagsliv). Do a new search in the site header. The old number of sections items are still visible.

jesperengstrom:
> There is still some reactivity issues regarding the number of pagination items.

Good call. `sequenceSize` being limited due to fewer hits was reactive, but not the reverse; when number of pages increased again. 

Extended the sequence to 8, let's see what happens when moving on to the styling issues...

lxl-web/tailwind.config.js
jesperengstrom: Yes, `mobile-only` was ugly. Wanted to modify the button classes without having to then re-apply the original styling for the sm-screens and up.

But going for bigger buttons now it's not needed anyway.

lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte
Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte]: Makes sense. Not really sure how you mean by the flex-wrap thing though, surely not this?

Skärmavbild 2024-05-30 kl  14 13 43

Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte]: You mean wrapping and hiding the middle (sequence) section? The current page would sometimes be in the hidden part...

Comment (johanbissemattsson) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte]: Ah, my proposed solution is probably not complete and needs some modification.

But in my head it would make sense to always keep the first, last and current (missed that one) pages visible. The number of items in the sequence around the current/first/last pages should be dynamic depending on the screen size (with an upper limit). The flex-wrap needs to be combined with overflow hidden to prevent items, and probably some containers.

It can be saved for a later PR if it requires too much reworking.

Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/(find)/Pagination.svelte]: Maybe it can be done in some clever way using pure CSS, but it eludes me atm. For example, the truncation (...) will have to appear when there is overflowing content.

Sort of reverted to the way it was, i.e showing the 'fixed' buttons for small screens. Looks better than before though, + bugs fixed. Should we settle for the time being?

jesperengstrom:
Added:
* `previous` from api that somehow got lost
* Remaining reactivity issues, should really be ok now ;)
* Center `pageSequence` around `currentPage`

Suggest we save mobile improvements for later

PR #1059, opened 2024-05-31
LWS-143: Related hits search

jesperengstrom:
### Tickets involved
[LWS-143](https://kbse.atlassian.net/browse/LWS-143)

### Solves

Adds a search input in the toolbar section for the related search result

### Summary of changes

* Added component `SearchRelated` that searches in the context of the current `_o` and `_p`.
* Zero results for the related search is no longer hidden, as it can now happen as a result of an active search.
* Some modifications to 'main search' to stay independent of related search.

Limits/todos:

* Searching completely clears any existing filters, due to frontends inability to modify `_q` param in a safe way. If we need to keep the filters, maybe template links (such as in year range) could be a solution.
* Layout/placement of the search bar for different widths is debatable (lack of design in Figma).

### DIFF
--- a/lxl-web/src/lib/components/Search.svelte
+++ b/lxl-web/src/lib/components/Search.svelte
@@ -9,9 +9,11 @@
 	export let autofocus: boolean = false;
 
 	$: showAdvanced = $page.url.searchParams.get('_x') === 'advanced';
-	let q = showAdvanced
-		? $page.url.searchParams.get('_q')?.trim()
-		: $page.url.searchParams.get('_i')?.trim();
+	let q = $page.params.fnurgel
+		? '' //don't reflect related search on resource pages
+		: showAdvanced
+			? $page.url.searchParams.get('_q')?.trim()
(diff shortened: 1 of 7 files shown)


johanbissemattsson (commented):
I think the front-end implementation looks good but I'm a little unsure about the user experience as the search doesn't always give relevant hits.

An example:

1. Go to http://localhost:5173/75kmmbnr2xl0fn5
2. Try searching for `kam` in the related search input
3. No results are found (I would atleast expect _Kamrater på en irrande planet_ to appear)

johanbissemattsson:
> I'm not sure everyone would expect that. As it is now we don't do automatically truncate terms anywhere. We can discuss that. Try searching for "kam*"

Aha! That explains it. It's out of scope of this PR but I think it's good to discuss it further in another context.

olovy (approved):
We'll merge this as is and postpone fixing the cleared filters a bit.

PR #1061, opened 2024-06-03
Alternative to external link icon

lrosenstrom:
### Tickets involved
[LWS-194](https://kbse.atlassian.net/browse/LWS-194)

### Summary of changes

The implementation for showing an icon after external links was severely shoehorned into the decorated data logic, with hard coded types etc to make it work. A fix for LWS-194 conforming to this implementation would require even more special handling, and felt strained.
This is an attempt at a more simple solution, that should work for properties regardless of their current presentation/style. 

It looks like this:
Screenshot from 2024-06-03 15-18-49

Screenshot from 2024-06-03 15-19-12

Inspired by top bar at https://iconify.design/docs/usage/svg/unplugin/:
Screenshot from 2024-06-03 15-36-55

Could maybe be fine tuned with a slight vertical raise?

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -527,7 +527,7 @@
 			"fresnel:resourceStyle": ["link"]
 		},
 		"seeAlso-format": {
-			"@id": "genreForm-format",
+			"@id": "seeAlso-format",
 			"@type": "fresnel:Format",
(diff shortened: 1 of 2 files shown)


olovy (approved):
LGTM!

I think the old icon was a bit better.
This implementation is much better.

lrosenstrom:
I agree that the icons looks a bit better (and I guess, is more of a standard)! 
(After discussion, we decided the new alternative is good enough and ready to be merged)

PR #1064, opened 2024-06-11
Fix failing tests

johanbissemattsson:
### Solves

Fixes failing tests

### Summary of changes

- Use test resource available on both QA and dev
- Fix missing testid
- Add screen-readable page title on find page
- Add aria-label on filter search input
- Add aria-label to filters navigation
- Change test query to sidestep issue with search result items without titles (see [LWS-206](https://kbse.atlassian.net/browse/LWS-206))
- Add aria label on remove filter link


jesperengstrom (approved):
Nice!

PR #1078, opened 2024-06-28
LWS-83: Show image attribution and usage policy

johanbissemattsson:
### Tickets involved
[LWS-83](https://jira.kb.se/browse/LWS-83)

### Solves

Show image attribution and usage/access policy.

Some examples:
[With attribution and usage/access policy](http://localhost:5173/dbqtzg3x4ndpm32)
[With usage/access policy](http://localhost:5173/qn245xp82zwbrfm)

### Summary of changes

- Show image attribution and usage policy
- Replace resourcePopover with more generic popover action


johanbissemattsson:
Can we rely on the typings of `publisher` and `usageAndAccessPolicy`? (see: https://github.com/libris/lxlviewer/pull/1078/files#diff-7c770cc88656a5e0c33ad447e43a633e446ed2f9cbb92480091ae3217d0584baR13-R19)

lxl-web/src/lib/components/DecoratedData.svelte
```diff
@@ -184,7 +186,9 @@
 					/>
 					{#if remainder && Array.isArray(remainder)}
 						<span
-							use:resourcePopover={{ data: remainder, lang: getSupportedLocale($page.params.lang) }}
+							use:popover={{
```
jesperengstrom, line 189: ...Similarly, I'd say the `remainder` stuff in DecoratedData is purely a leftover from the previous card. It should probably just be removed. Maybe not in scope of this PR...

jesperengstrom (approved):
Nice! LGTM ✨

lxl-web/src/lib/components/DecoratedData.svelte
jesperengstrom: ...if we don't decide to use it again to prevent [this](https://beta.libris-qa.kb.se/find?_q=musikerminnen+svenska+musikers&_limit=20&_offset=0&_sort=&_i=musikerminnen+svenska+musikers&_x=advanced)

johanbissemattsson: Yeah, I was split between keeping or removing `remainder` (as we will need some support for truncating the decorated data) but if the list of items is too long it still becomes an issue even with the current remainder implementation tooltip (see attached screenshot).

Skärmavbild 2024-07-08 kl  11 39 37

So, I think I'm currently in favour of keeping it as is for the sake of the scope of this PR. Or should we remove it altogether to later reimplement it? 🤷

olovy:
Did some small tweaks
* display image publisher link if available
* fix the check for nielsen images
* display a notice on [cropped images](https://wiki.creativecommons.org/wiki/Recommended_practices_for_attribution#This_is_a_great_attribution_for_an_image_you_modified_slightly)

TODO
I think we should include a a link to the wikimedia canonical/landing page for an image. [Example](https://commons.wikimedia.org/wiki/File:Matilda_S%C3%B6dergran_in_2019.jpg). But we don't have that available in the image data yet. We might have do redesign the image info area a bit for this.

Merging this as is.

PR #1080, opened 2024-07-05
LWS-209: Add spinner

jesperengstrom:
## Description
Adds a loading indicator

### Tickets involved
[LWS-209](https://kbse.atlassian.net/browse/LWS-209)

### Solves
No indication of anything happening when for example searching.

### Summary of changes

* Added a progress bar (rather than a spinner) using the [nProgress](https://www.npmjs.com/package/nprogress) package. It runs on every navigation but i've added a small delay to avoid showing it on instant navigations. This can of course be tweaked further.
* With this in place, I tried removing the streaming of search results on resource pages. This would otherwise require a _second_ spinner to replace the 'loading' text. I think this works well - with the indicator in place, it's quite satisfying to get the whole page at once.
* Made some minor related adjustments to the `load` function on the resource page. The `getRelated()` / `await getRelated()` call was done as the last thing before returning, not optimal when everything else is ready. We want to do it as soon as possible (i.e when there's a `resourceId`), but not wait for it, blocking subsequent operations. So we now make the call early and store the promise in a variable, awaiting it at the very end.

### DIFF
--- a/lxl-web/package-lock.json
+++ b/lxl-web/package-lock.json
@@ -32,6 +32,7 @@
 				"lxljs": "file:../lxljs",
 				"magic-string": "^0.30.7",
 				"mdsvex": "^0.11.2",
+				"nprogress": "^0.2.0",
 				"postcss": "^8.4.38",
 				"postcss-import": "^16.1.0",
 				"prettier": "^3.3.2",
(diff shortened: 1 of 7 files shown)


johanbissemattsson (approved):
LGTM! Awesome –  I agree that it's feels much more responsive and satisfying to navigate around now! 🌟 

Good call with the `getRelated` promise 👍 We could maybe also achieve the same effect by implementing multiple parallel promises in (or near) the return (instead of having to declare them as variables beforehand) but I think it's best done if we can do it together while pair programming.

olovy (approved):
LGTM
Agree that it feels more responsive!

Merging this.

PR #1082, opened 2024-08-05
Block id.kb.se and libris.kb.se datasets from being edited

olovy:
Do a simple hardcoded check for dataset URI prefixes for now. Until we have decided on an how to specify this in backend. See for example EntityContainers for Concepts.

### Tickets involved
[LXL-4254](https://jira.kb.se/browse/LXL-4254)

### Solves
Prevents accidental editing of language, library and other definitions from externally defined datasets.
There is still no access control mechanism for these in backend. This does not guard against malicious changes.


andersju (approved):
LGTM

PR #1084, opened 2024-08-08
LWS-219: Fix resource image placement on Safari

jesperengstrom:
### Tickets involved
[LWS-219](https://kbse.atlassian.net/browse/LWS-219)

### Solves

Weird placement of resource images in Safari. (Not just mobile, this is desktop Safari:)

Skärmavbild 2024-08-08 kl  15 29 05

### Summary of changes

* Clearly, Safari doesn't respect the `max-height` of an element with `display:table` (introduced with the `figcaption` stuff). Using only `height` works - it doesn't seem to break anything else...
* Setting `w-full` on the image container allows the image to align to the page left margin on md+ screens. This seems to be desired since it has `justify-center md:justify-start`.

### DIFF
--- a/lxl-web/src/lib/components/ResourceImage.svelte
+++ b/lxl-web/src/lib/components/ResourceImage.svelte
@@ -24,7 +24,7 @@
 </script>
 
 {#if image && thumb}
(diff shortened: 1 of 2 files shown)


olovy (approved):
LGTM

PR #1085, opened 2024-08-09
LWS-201/203: Fix pill labels

jesperengstrom:
### Tickets involved
[LWS-201](https://kbse.atlassian.net/browse/LWS-201)
[LWS-203](https://kbse.atlassian.net/browse/LWS-203)

### Solves

* Firefox chip gap
* Missing labels for ISBN & ISSN

### Summary of changes

Giving up on trying to fix both lowercase labels & weird Firefox pill gap with pure CSS, we now do it "for real" in the FE-backend.

In the process I discovered that the label is sometimes found in `property.label` (duh!) which we never check before falling back on the `@id`. This fixes both examples (ISBN and ISSN) in LWS-203. Don't know if you've seen more cases @poacherone?

### DIFF
--- a/lxl-web/src/lib/utils/search.ts
+++ b/lxl-web/src/lib/utils/search.ts
@@ -82,7 +82,7 @@ function displayMappings(
 					display: displayUtil.lensAndFormat(property, LensType.Chip, locale),
 					label: m.alias
 						? translate(`facet.${m.alias}`)
-						: m.property?.labelByLang?.[locale] || m.property?.['@id'] || 'no label', // lensandformat?
+						: capitalize(m.property?.labelByLang?.[locale] || m.property?.label) || m.property?.['@id'] || 'No label', // lensandformat?
 					operator,
 					...('up' in m && { up: replacePath(m.up as Link, usePath) })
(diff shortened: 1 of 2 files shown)


lxl-web/src/lib/utils/search.ts
```diff
@@ -82,7 +82,7 @@ function displayMappings(
 					display: displayUtil.lensAndFormat(property, LensType.Chip, locale),
 					label: m.alias
 						? translate(`facet.${m.alias}`)
-						: m.property?.labelByLang?.[locale] || m.property?.['@id'] || 'no label', // lensandformat?
+						: capitalize(m.property?.labelByLang?.[locale] || m.property?.label) || m.property?.['@id'] || 'No label', // lensandformat?
```
olovy, line 85: Yes we should fix `lensAndFormat` so that it can handle this instead!

olovy (approved):
LGTM!

`lensAndFormat` / display config should be fixed so that it can handle the property label.

But let's merge this as is and move on.

poacherone:
ISBN and ISSN were the ones I had seen @jesperengstrom.

PR #1086, opened 2024-08-12
LXL-4457: Keep sort options when doing new searches

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4457](https://kbse.atlassian.net/browse/LXL-4457)

### Solves

Keeps sort options when doing new searches from the search result page.

This works by reusing old sort query param for new searches. Note that this doesn't involve the local storage functionality (which currently only works if the user is signed in), so the last selected sort option isn't persisted if an unauthenticated user navigates to the start page and enters a new search query. 

### Summary of changes

- Use _sort value from route as preferred sort
- Check if sort value from route query param is available for active search type

### DIFF
--- a/vue-client/src/components/search/search-form.vue
+++ b/vue-client/src/components/search/search-form.vue
@@ -236,7 +236,9 @@ export default {
     prefSort() {
       if (
         this.$route.query?._sort
-        && this.settings.sortOptions[this.activeSearchType]?.find((sortOption) => sortOption.query === this.$route.query._sort)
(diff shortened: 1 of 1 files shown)


johanbissemattsson:
> All the sort options seems to stick except `_sortKeyByLang` (A-Ö, Ö-A), which keeps resetting. How come?

This should be resolved now by using the `includes()` method instead of a strict equality check to allow localized `_sortKeyByLang` (e.g. `_sortKeyByLang.sv`)

PR #1087, opened 2024-08-14
Fix display of user's own holding in hitlist

olovy:
Was broken by the early return in c4434a0.
We can always find the holdings in `@reverse/itemOf` for each Instance in the search result. 
So remove the code that does an additional query for the holding.

## Checklist:

### Tickets involved
[LXL-4541](https://kbse.atlassian.net/browse/LXL-4541)

### DIFF
--- a/vue-client/src/components/inspector/reverse-relations.vue
+++ b/vue-client/src/components/inspector/reverse-relations.vue
@@ -112,11 +112,19 @@ export default {
       return this.myHolding !== null;
     },
     myHolding() {
+      const isLink = (o) => Object.keys(o).length === 1 && Object.keys(o).includes('@id');
+
       if (this.user.isLoggedIn) {
         // Check if my sigel has holding
         const libraryUri = this.user.getActiveLibraryUri();
         const holdings = get(this.mainEntity, ['@reverse', 'itemOf'], []);
-        const myHolding = holdings.find((h) => get(h, ['heldBy', '@id']) === libraryUri);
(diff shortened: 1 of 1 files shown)


vue-client/src/components/inspector/reverse-relations.vue
```diff
@@ -112,11 +96,6 @@ export default {
           });
       }, timeoutLength);
     },
-    gotoHolding() {
-      const locationParts = this.myHolding.split('/');
-      const fnurgel = locationParts[locationParts.length - 1];
-      this.$router.push({ path: `/${fnurgel}` });
-    },
```
olovy, line 119: dead code

jesperengstrom:
In the search results: Looking good.

On the instance itself, kind of the opposite... Holding displaying correctly on dev:
Skärmavbild 2024-08-14 kl  15 48 08

But not on this branch:
Skärmavbild 2024-08-14 kl  15 48 57

`myHolding` returning `null`, do we need to look for it somewhere else? 🤔

olovy:
> On the instance itself,  ...  `myHolding` returning `null`, do we need to look for it somewhere else? 🤔

Good catch!

The data is framed in the search result but not when opening a record in the inspector.
So then we have to look in quoted (embellished) instead.

jesperengstrom (approved):
LGTM!

olovy:
A further improvement would be to display the holding link directly without the spinner when viewing the record.

PR #1088, opened 2024-08-14
LXL-4485: Add date & sigel to cxz message comment

jesperengstrom:
## Checklist:

### Tickets involved
[LXL-4485](https://kbse.atlassian.net/browse/LXL-4485)
### Solves

Pre-fill cxz comment with sigel & current date for better orientation when listing messages.

### Summary of changes

Add the template text to `language-entry` textarea on init if conditions are met: It's an empty `comment` whose mainEntity type is subclass of `AdministrativeAction`.

### DIFF
--- a/vue-client/src/components/care/create-message.vue
+++ b/vue-client/src/components/care/create-message.vue
@@ -1,8 +1,9 @@
 <script>
 import { mapGetters } from 'vuex';
 import { sortBy, set, pick } from 'lodash-es';
-import CreationCard from "@/components/create/creation-card.vue";
-import * as RecordUtil from "@/utils/record";
+import CreationCard from '@/components/create/creation-card.vue';
+import { formatDate } from '@/utils/datetime';
+import * as RecordUtil from '@/utils/record';
 import * as StringUtil from 'lxljs/string';
 
 export default {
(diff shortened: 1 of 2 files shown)


johanbissemattsson:
I am not so familiar with CXZ but should the template also be inserted when user replies to another message by adding a comment (by pressing the plus sign)?

See attached screenshot:
 
Skärmavbild 2024-08-15 kl  13 22 06

I think it looks good to me otherwise!

olovy:
> I am not so familiar with CXZ but should the template also be inserted when user replies to another message (by adding a comment/pressing the plus icon?)

@jesperengstrom @johanbissemattsson 
Yes, it should be inserted for every comment.
(Only for entity types `InquiryAction` and `ChangeNotice` = subclass of `AdministrativeAction`)

jesperengstrom:
> @jesperengstrom @johanbissemattsson Yes, it should be inserted for every comment. (Only for entity types `InquiryAction` and `ChangeNotice` i.e. subclass of `AdministrativeAction`)

Ah ok. Then we need to go for a different solution, to dynamically add it to the fields. Will give it another try.

jesperengstrom:
New try. See updated description.

olovy (approved):
LGTM!

The template text gets its own "step" in the undo history but I think that is fine?
We don't want to complicate this too much.

jesperengstrom:
> The template text gets its own "step" in the undo history but I think that is fine?

Yes, maybe its fine?
We could also do `this.$refs.textarea.innerText = this.templateText` (hacky) or build an option to bypass history (complicated)

olovy:
> > The template text gets its own "step" in the undo history but I think that is fine?
> 
> Yes, maybe it's fine? We could also do `this.$refs.textarea.innerText = this.templateText` (hacky) or build an option to bypass history (complicated)

Yes, let's merge this as is

PR #1089, opened 2024-08-16
LXL-4539: handle multiple types

jesperengstrom:
## Checklist:

### Tickets involved
[LXL-4539](https://kbse.atlassian.net/browse/LXL-4539)

### Solves

* Result list breaking when a hit has multiple `@type`s (locally - in production showing empty boxes:)
Skärmavbild 2024-08-16 kl  11 41 12

* [local](http://localhost:8080/katalogisering/search/libris?q=*&@type=Concept&_limit=20&_sort=_sortKeyByLang.sv&_spell=true&and-inScheme.@id=https://id.kb.se/marc&_offset=0)
[dev](https://libris-dev.kb.se/katalogisering/search/libris?q=*&@type=Concept&_limit=20&_sort=_sortKeyByLang.sv&_spell=true&and-inScheme.@id=https://id.kb.se/marc&_offset=0)

* Inspector breaking when `mainEntity` has multiple `@type`s (locally, in production some information is just missing)
[local](http://localhost:8080/katalogisering/9h0hfnvdc4crddn5)
[dev](https://libris-dev.kb.se/katalogisering/9h0hfnvdc4crddn5)

### Summary of changes

* In the result list, the culprit is the `DisplayUtil.getDisplayObject` passing the type array to `getLabelByLang`. A few lines up we actually save the _first_ type in a variable, so let's use that:
https://github.com/libris/lxlviewer/blob/3f4125b3c203100701529b5f6c8191e4b8d5e86f/lxljs/display.js#L418

* In the inspector, `item-type.vue` gets the array as its `field-value`, which it doesn't like. But the component is actually looped out, so we can just use the current `value` instead.
* In logged in mode, the toolbar relies heavily on `mainEntity['@type']` being a string to determine what can be edited etc. Not exactly sure what consequences this has, but the breaking error is fixed by passing the first type to `VocabUtil.getBaseClasses`.

...The whole app seems to be built upon the assumption of a single `@type`, so I don't know if these records cause trouble in places I didn't think of. Prop-check still warns that `entityType` cannot be an array. Should we fix it because it can, or can it?

### DIFF
--- a/vue-client/src/components/inspector/toolbar.vue
+++ b/vue-client/src/components/inspector/toolbar.vue
@@ -294,14 +294,12 @@ export default {
       this.showAdminInfoDetails = !this.showAdminInfoDetails;
     },
     isSubClassOf(type) {
-      const mainEntityType = this.inspector.data.mainEntity['@type'];
-      const baseClasses = VocabUtil.getBaseClasses(
-        // If more than one type, choose the first
-        Array.isArray(mainEntityType) ? mainEntityType[0] : mainEntityType,
+      return VocabUtil.isSubClassOf(
+        this.inspector.data.mainEntity['@type'],
+        type,
         this.resources.vocab,
         this.resources.context,
-      ).map((id) => StringUtil.getCompactUri(id, this.resources.context));
-      return baseClasses.indexOf(type) > -1;
+      );
     },
     isInReadOnlyDataset(record) {
       // TODO: get from backend
--- a/vue-client/src/components/inspector/toolbar.vue
+++ b/vue-client/src/components/inspector/toolbar.vue
@@ -294,8 +294,10 @@ export default {
       this.showAdminInfoDetails = !this.showAdminInfoDetails;
     },
     isSubClassOf(type) {
+      const mainEntityType = this.inspector.data.mainEntity['@type'];
       const baseClasses = VocabUtil.getBaseClasses(
-        this.inspector.data.mainEntity['@type'],
+        // If more than one type, choose the first
+        Array.isArray(mainEntityType) ? mainEntityType[0] : mainEntityType,
         this.resources.vocab,
         this.resources.context,
       ).map((id) => StringUtil.getCompactUri(id, this.resources.context));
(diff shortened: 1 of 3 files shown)


johanbissemattsson:
Great work! 🌟 LGTM!

> ...The whole app seems to be built upon the assumption of a single @type, so I don't know if these records cause trouble in places I didn't think of. Prop-check still warns that entityType cannot be an array. Should we fix it because it can, or can it?

I'm also unsure of this so it could maybe be relevant with further discussions around this. E.g. should `isSubClassOf` return an aggregate of all types or just the first (as in the PR)?

johanbissemattsson (approved):

vue-client/src/components/inspector/toolbar.vue
```diff
@@ -294,8 +294,10 @@ export default {
...
     },
     isSubClassOf(type) {
+      const mainEntityType = this.inspector.data.mainEntity['@type'];
       const baseClasses = VocabUtil.getBaseClasses(
-        this.inspector.data.mainEntity['@type'],
+        // If more than one type, choose the first
+        Array.isArray(mainEntityType) ? mainEntityType[0] : mainEntityType,
         this.resources.vocab,
         this.resources.context,
       ).map((id) => StringUtil.getCompactUri(id, this.resources.context));
```
olovy, line 303: This can be replaced by VocabUtil.isSubClassOf (which also happens to handle multiple types)

olovy (commented):
LGTM.

See suggestion about using VocabUtil.isSubClassOf.

It would be best if multiple types were handled consistently everywhere and displayed. 
But the important thing right now is to fix errors breaking the interface.

Multiple types should normally only appear in some vocab classes.
But they sometimes appear in broken records, e.g, `instanceOf.hasPart` here
https://libris-qa.kb.se/katalogisering/j1trg6xv00bw18m

vue-client/src/components/inspector/toolbar.vue
jesperengstrom: Ok! Now the local method only passes the argument and returns the result from `VocabUtil.isSubClassOf`. Seems to work fine...

olovy (approved):

PR #1073, opened 2024-06-25
Feature/lxl 4504 cataloguing items

lrosenstrom:
### Tickets involved

[LXL-4504](https://jira.kb.se/browse/LXL-4504)

### Solves

Support for creating and displaying of IDs when cataloguing items*.

### Summary of changes

- Make it possible to add an ID to a copy by clicking the "vertical ellipsis" icon next to the add button. The client creates a temporary ID that is replaced by a backend-generated ID on save
- Display IDs in the upper right corner of local items as long as some conditions are fulfilled (is a `hasComponent` entry and has an `@id`)
- Support for showing "ID-labels" as added or removed in the history view

### TODO: 
- When trying to link to a copy by adding e.g. a `relatedTo` field and searching for the copy's ID in the right search panel, the only hits are the instance and item containing the copy. How should this work? Are backend changes needed?
- Show reverse relations for copies, i.e. a (smaller than the current standard?) button, displaying the number of links to the copy, and the entities containing the links when clicking the button.  

\* "Exemplar" in Swedish.

PR #1091, opened 2024-08-19
Add support for linked SingleItem in holdings

olovy:
## Checklist:

## Description
Add support for `SingleItem`s (sv: exemplar) in holdings. 
Initially to be used for pinpointing `itemUsed` in digital reproductions.

An Item in `hasComponent` can be converted to a `SingleItem`.
It looks similiar to when extracting linked works.
Skärmbild från 2024-08-19 17-52-22
Skärmbild från 2024-08-19 17-53-27

The single items are always stored in their own records and linked from the the holding record.
Note the id in the top right corner.
Skärmbild från 2024-08-19 17-53-46

But they can be edited in place as if the were part of the same record.
Skärmbild från 2024-08-19 17-54-07

There are separate options for removing the linked entity or just the link. 
Reused the :x: for removing links. Maybe needs to be more obvious.
Skärmbild från 2024-08-19 17-54-53
Skärmbild från 2024-08-19 17-55-16

There can of course be many linked items and mixed with local items.
Skärmbild från 2024-08-19 17-56-32

The feature is currently only available for sigel S.

Depends on https://github.com/libris/definitions/pull/494
Depends on https://github.com/libris/librisxl/pull/1467

### Tickets involved
[LXL-4504](https://kbse.atlassian.net/browse/LXL-4504)

### Summary of changes
* Add `settings.extractableMappedTypes` that specifies which types can be extracted and which type they should be converted to. That is, `Item` can be extracted to its own document and should then be converted to `'SingleItem`. This is a bit clumsy but does the job for now. 
* When loading a record/document, other documents that should be inlined are moved from quoted to the main inspector data.  
    * This is done for linked things in properties with category `compositional` (only `hasComponent`...). 
        * TODO? Should be `compositional` + `integral`
    * Their ETags are fetched and stored for later use.
    * We count on embellish data to include the whole docs and don't fetch the doc data separately.
* When saving, inlined documents are extracted and updated before the main document in separate requests.
* Inlined records that should be deleted are deleted after the main doc is saved (this might fail...).

* 81208b0c7b5d2f575b0b6bd95f8d8a8d1eed25cd breaks out the id-pill used in the top right corner of entity-summary to its own component. Taken from https://github.com/libris/lxlviewer/pull/1073
* bcf023b19f09e5f9205de6d74c34adc6bb714b8e contains the main changes.
* The other commits just remove obsolete and dead code that I encountered.

### Not in scope
There is no way to move (link) a SingleItem to another holding record. 

# TODO
* Check MARC21 conversion
* Don't include inlined documents in "Download as JSON-LD" (?)
* Duplicating or copying a SingleItem creates a local SingleItem. Is this desired?
* Handle deletion of the whole holding?
* Future: have a way to update multiple records atomically.


vue-client/src/views/Inspector.vue
```diff
@@ -821,9 +859,46 @@ export default {
...
+      if (this.inlinedIds.includes(obj['@graph'][1]['@id'])) {
+        // we only handle one level of inlined docs so far
+        return;
+      }
+
+      const inlined = RecordUtil.extractInlinedData(obj['@graph'][1], this.inlinedIds, this.resources);
+      for (const id of this.inlinedIds) {
+        if (inlined[id]) {
+          // FIXME
+          // eslint-disable-next-line no-await-in-loop
```
johanbissemattsson, line 878: This could probably be rewritten to  `for await (const id of this.inlinedIds) {...` to fix the `no-await-in-loop` issue but I think it works for now!

olovy: See https://github.com/libris/lxlviewer/pull/1091/commits/eb47f1ee30c2117ce3c90d551b513aa9c4fd1d6a

johanbissemattsson: Even better! 👌

vue-client/src/components/inspector/item-local.vue
```diff
@@ -559,7 +603,7 @@ export default {
         />
 
         <entity-action
-          v-if="inspector.status.editing && !isLocked"
+          v-if="inspector.status.editing && !isLocked && !isInlinedRecord && !isExtracting"
```
olovy, line 606: `!isExtracting` is a workaround for a bug in the old "extract" code.
I you start extracting an entity and the delete it you get an error when saving.

It is still possible to trigger:
* Add a new Item in has component
* Click "Extract Single item" / "Extrahera Exemplar"
* Click Undo (removing the added item)
* Save
* Error!

jesperengstrom:
LGTM!

Must say I have little knowledge of this part of the code, but all looks reasonable. Nice cleaning job too.

Have tried creating/extracting, editing, viewing (both inline and on its own) and unlinking/deleting `singleItem`s. Seems to work as expected.

The only thing I noticed is an alert saying I have unsaved changes when I try to navigate away in edit mode, even if I save and try again. Using the green button does it. Maybe it’s always been like this.

jesperengstrom (approved):

PR #1093, opened 2024-08-22
LWS-225: improve display of libraries in holdings panel

jesperengstrom:
### Tickets involved
[LWS-225](https://kbse.atlassian.net/browse/LWS-225)

### Solves

Display additional (distinguishing) library information in the holdings panel.

This could really be as easy as printing the `qualifier` alongside the name. But I tried a more cumbersome approach 😆 
To stay consistent with how we do it elsewhere (libraries in the facets, for example) we now display decorated data here.

### Summary of changes

* `lensAndFormat` the libraries inside the functions that prepare data for the holdings panel: `holdingsByInstanceId` and `holdersByType`. 
* Break out new util method `getHoldingsByType` in /utils. 
* New types and some checks to get rid of ts-warnings in `/fnurgel/page.svelte`. 
* The decorated labels are ellipsed, but displayed multi-line when opened. 

* We now also use `DecoratedData` component in facets instead of printing the rendered `str`. This should give us more fine-grained control over the styling?

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -653,6 +653,30 @@
 				"fresnel:contentFirst": ""
 			}
 		},
+		"Collection-qualifier-format": {
+			"@id": "Collection-qualifier-format",
+			"@type": "fresnel:Format",
+			"fresnel:classFormatDomain": ["Collection"],
+			"fresnel:propertyFormatDomain": ["qualifier"],
+			"fresnel:propertyFormat": {
+				"fresnel:contentBefore": ", ",
+				"fresnel:contentFirst": "",
+				"fresnel:contentAfter": "",
+				"fresnel:contentLast": ""
+			}
+		},
+		"Collection-sigel-format": {
+			"@id": "Collection-sigel-format",
+			"@type": "fresnel:Format",
+			"fresnel:classFormatDomain": ["Collection"],
+			"fresnel:propertyFormatDomain": ["sigel"],
+			"fresnel:propertyFormat": {
+				"fresnel:contentBefore": " (",
+				"fresnel:contentFirst": "(",
+				"fresnel:contentAfter": ")",
+				"fresnel:contentLast": ")"
+			}
+		},
 		"default-separators": {
 			"@id": "default-separators",
 			"@type": "fresnel:Format",
--- a/lxl-web/src/lib/components/find/FacetGroup.svelte
+++ b/lxl-web/src/lib/components/find/FacetGroup.svelte
@@ -4,6 +4,8 @@
 	import { type FacetGroup } from '$lib/types/search';
 	import BiChevronRight from '~icons/bi/chevron-right';
 	import FacetRange from './FacetRange.svelte';
+	import DecoratedData from '../DecoratedData.svelte';
+	import { ShowLabelsOptions } from '$lib/types/decoratedData';
 
 	export let group: FacetGroup;
 	export let locale: LocaleCode;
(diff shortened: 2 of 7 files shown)


olovy (approved):
LGTM!

Perhaps something like this to start with?
```
		"Collection-qualifier-format": {
			"@id": "Collection-qualifier-format",
			"@type": "fresnel:Format",
			"fresnel:classFormatDomain": ["Collection"],
			"fresnel:propertyFormatDomain": ["qualifier"],
			"fresnel:propertyFormat": {
				"fresnel:contentBefore": ", ",
				"fresnel:contentFirst": "",
				"fresnel:contentAfter": "",
				"fresnel:contentLast": ""
			}
		},
		"Collection-sigel-format": {
			"@id": "Collection-sigel-format",
			"@type": "fresnel:Format",
			"fresnel:classFormatDomain": ["Collection"],
			"fresnel:propertyFormatDomain": ["sigel"],
			"fresnel:propertyFormat": {
				"fresnel:contentBefore": " (",
				"fresnel:contentFirst": "(",
				"fresnel:contentAfter": ")",
				"fresnel:contentLast": ")"
			}
		},
```

It would be nice if sigel was a bit lighter but couldn't get it to work right away with `propertyStyle`

jesperengstrom:
If `"fresnel:propertyStyle": ["someClass"]` is set in this case, it's added to `_style` as expected, but then ignored by the  `DecoratedData` component. It applies classes for objects with a `@type` (`data-type`), but not in other cases (i.e `data-property`).

It works by simply adding
```class={getStyleClasses(data)}```

...but there must be some reason for this distinction @johanbissemattsson ?

olovy:
I think it is a bit unclear if `fresnel:propertyStyle` should apply to contentBefore/After.
I think it should?

Styling the sigel in the same way as e.g. inScheme (sao etc) might be nice and consistent?
And avoids the issue for now

Skärmbild från 2024-08-22 17-03-02

jesperengstrom:
>I think it is a bit unclear if fresnel:propertyStyle should apply to contentBefore/After.
>I think it should? Otherwise the parentheses get a different style in this case.

Yeah, it's the same thing with `role` that always bugged me a bit

Skärmavbild 2024-08-22 kl  17 28 02

But having the parentheses underlined looks weird... Don't know if there's a universal solution to this 🤔

olovy:
> > I think it is a bit unclear if fresnel:propertyStyle should apply to contentBefore/After.
> > I think it should? Otherwise the parentheses get a different style in this case.
> 
> Yeah, it's the same thing with `role` that always bugged me a bit
> Skärmavbild 2024-08-22 kl 17 28 02
> 
> But having the parentheses underlined looks weird... Don't know if there's a universal solution to this 🤔

I think valueStyle should be used to set the underline and propertyStyle to set the size/color of the whole role parenthesis

johanbissemattsson:
Awesome! LGTM! :star2:

> > But having the parentheses underlined looks weird... Don't know if there's a universal solution to this 🤔
> 
> I think valueStyle should be used to set the underline and propertyStyle to set the size/color of the whole role parenthesis

I think this is the closest to a sensible solution, so I also think `fresnel:propertyStyle` _should_ preferably also be applied to `contentBefore`/`contentAfter`.

johanbissemattsson (approved):

jesperengstrom:
> Styling the sigel in the same way as inScheme (sao etc) might be nice and consistent?

A bit reluctant to do that, since (I just now discovered) the contrast is too low and violates WCAG (it's not automatically detected since they're all collapsed on page load). 

Skärmavbild 2024-08-23 kl  11 04 46

Should I continue to pursue better styling control through formatters, or is it becoming out of scope?

jesperengstrom:
Merging this now. Will do a separate PR of some changes I made trying to improve the _style thing. We can try it out there.

olovy:
> A bit reluctant to do that, since (I just now discovered) the contrast is too low and violates WCAG

I suspected it did!

PR #1094, opened 2024-08-23
LXL-4542: Update vue-client dependencies

johanbissemattsson:
## Checklist:

### Tickets involved
[LXL-4542](https://kbse.atlassian.net/browse/LXL-4542)

### Solves

Bumps dependencies (upgrades both patch and minor versions as well as selected major versions).

**The exceptions are:**
- `bootstrap` which is kept on `3.4.1` as there are **_LOTS_** of breaking changes if we are to upgrade to a later major version. There is sadly a Cross-Site Scripting (XSS) vulnerability in Bootstrap `3.X.X` but it only seems to affect the [carousel component](https://github.com/advisories/GHSA-9mvj-f7w8-pvh2) (which we don't seem to use?)
- `combokeys` which still uses an exotic dependency (which is actually ahead of the version released on npm?)
- `kungbib-styles` which has some breaking changes in the latest version (e.g. missing logo). We should think about removing the dependency altogether as the styling of `vue-client` has in many ways departed from the styling in kungbib-styles. This is therefor probably best kept as a separate PR.

**There are still some vulnerabilities when running `yarn audit` which we should address soon:**
- Moderate severity in peer dependency `tough-cookie` used by `client-oauth2` (where the later hasn't been updated to include the fix)
- 3 low severeity vulnerabilities in peer dependency `elliptic` used by `crypto-browserify` (where the later hasn't been updated to include the fixes). This need  some further investigation but I think we could remove `crypto-browserify` altogether in favour of the [Web Crypto API](https://www.w3.org/TR/WebCryptoAPI/).
- Moderate severity vulnerability in Bootstrap (see above)

### Summary of changes
- Bump `vue`, `vite` and `@vitejs/plugin-vue`
- Remove unused `vue-cli-plugin-style-resources-loader` dependency
- Fix deprecate glob option `"as"`
- Bump `start-server-and-test` causing moderate audit issue
- Bump dependencies to latest patch version
- Bump dependencies to latest minor version
- Bump `@vue/eslint-config-airbnb` to latest major version
- Bump `@vueuse/core` and `@vueuse/components` to latest major versions
- Bump `autosize` to latest major version
- Bump `eslint` to latest major version
- Bump `eslint-plugin-cypress` to latest major version
- Bump `happy-dom`to latest major version
- Upgrade `marked` and add `dompurify` to sanitize output
- Upgrade `vitest` to latest major version
- Upgrade `floating-vue` to latest major version
- Configure ESM as default


johanbissemattsson:
I have now added a new ticket revolving the moderate severity vulnerability in client-oauth2: https://kbse.atlassian.net/browse/LXL-4544

jesperengstrom:
Nice job! ⭐ Everythings seems to be working.
However, I get at load of these in the console (but not on develop). Maybe worth investigating, if it's something caused by the Vue upgrade?

Skärmavbild 2024-08-23 kl  15 00 00

johanbissemattsson:
> Nice job! ⭐ Everythings seems to be working. However, I get at load of these in the console (but not on develop). Maybe worth investigating, if it's something caused by the Vue upgrade?
> Skärmavbild 2024-08-23 kl 15 00 00

This should be resolved in https://github.com/libris/lxlviewer/pull/1094/commits/38c25fe62e9b01673cc2c4e4af188df55e4e23d5

jesperengstrom (approved):
Warnings gone, looking good :)

PR #1096, opened 2024-08-23
Fix fnurgel being displayed in id-pill instead of id.kb.se URI

olovy:
## Description
Fix fnurgel being displayed in card top corner instead of id.kb.se URI.
This affected search results and popup cards for links.

QA
thumb-Clipboard - 23 augusti 2024 kl  14_21

PROD
thumb-Clipboard - 23 augusti 2024 kl  14_22

Also: 
* Fix id-pill overflow on long URI:s (in the same way as before it was extracted to a component)
* Naming things: displayed id is not always a fnurgel

### Tickets involved
[LXL-4545](https://kbse.atlassian.net/browse/LXL-4545)

### DIFF
--- a/vue-client/src/components/shared/entity-summary.vue
+++ b/vue-client/src/components/shared/entity-summary.vue
@@ -306,7 +306,7 @@ export default {
       </div>
       <id-pill
         v-if="!excludeComponents.includes('id')"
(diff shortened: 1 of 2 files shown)


lrosenstrom (approved):
LGTM!

PR #1097, opened 2024-08-27
LWS-210: Filter libraries

jesperengstrom:
### Tickets involved
[LWS-210](https://kbse.atlassian.net/browse/LWS-210)

### Solves

Adds an input to find libraries in the holdings list. 

Works/looks very similar to the search in the facet panel, with some differences:
* The search string is used to create a filtered list, rather than sent to the individual components. This felt like a cleaner way to do it + we can display 'no hits'. Trade-of is components are unmounted/mounted on search, meaning we have to perform the api call again if re-displayed, re-expanded.
* The search is matched anywhere in the string (`indexOf`) to allow searching for sigel, for example.

* ATM the search string remains in the input if panel is closed and re-opened. Don't know if that's good or bad?

### Summary of changes
* Add the rendered (searchable) string to `DecoratedHolder` obj
* Add reactive variables `displayedHolders` / `filteredHolders`
* Add input

### DIFF
--- a/lxl-web/src/lib/i18n/locales/en.js
+++ b/lxl-web/src/lib/i18n/locales/en.js
@@ -131,6 +131,7 @@ export default {
 		availableAt: 'Available at',
 		library: 'library',
 		libraries: 'libraries',
+		findLibrary: 'Find library',
 		findAtYourNearestLibrary: 'Find it at your nearest library',
 		location: 'Location',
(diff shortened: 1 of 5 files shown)


olovy (approved):
LGTM

Seems to work fine

johanbissemattsson (approved):
LGTM!  🙌

PR #1098, opened 2024-08-29
LWS-218: Display spelling suggestions

jesperengstrom:
### Tickets involved
[LWS-218](https://kbse.atlassian.net/browse/LWS-218)

### Solves

Displays spelling suggestions next to the search result. (It's an array, but I've never seen more than one?) 
Printing the `label` instead of the `labelHtml` since ESLint discourages the use of `{@html}`.

Also, as by suggestion from @johanbissemattsson, rewriting the link to `spell=false` to prevent getting _another_ suggestion after accepting the first one.

We need to observe if the functionality is at all useful with 'related search' on resource pages. You often get suggestions that result in zero hits there (I guess it doesn't take into account the implicit o-filter...)

Skärmavbild 2024-08-29 kl  15 10 32

### Summary of changes

* Display spelling suggestion
* Add type `SpellingSuggestion`
* Make `SearchRelated` update `_i` param on navigation
* Relation search layout adjustments
* Fix failing tests

### DIFF
--- a/lxl-web/src/lib/components/find/SearchRelated.svelte
+++ b/lxl-web/src/lib/components/find/SearchRelated.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
-	import BiSearch from '~icons/bi/search';
+	import { afterNavigate } from '$app/navigation';
 	import { page } from '$app/stores';
-	import type { Link } from '$lib/utils/xl';
+	import type { Link } from '$lib/types/xl';
+	import BiSearch from '~icons/bi/search';
 
 	export let view: Link;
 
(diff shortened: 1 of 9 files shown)


olovy:
> (It's an array, but I've never seen more than one?) 
Yup, for now it's always zero or one

> Printing the `label` instead of the `labelHtml` since ESLint discourages the use of `{@html}`.
The advantage of using `labelHTML` is that it highlights the part of the phrase that has been corrected.
For example: "sara idman" -> "sara _lidman_"

> We need to observe if the functionality is at all useful with 'related search' on resource pages. You often get suggestions that result in zero hits there (I guess it doesn't take into account the implicit o-filter...)
Good point. It always gives suggestions based on the whole index regardless of any filters.

jesperengstrom:
> The advantage of using labelHTML is that it highlights the part of the phrase that has been corrected.
For example: "sara idman" -> "sara lidman"

Ah, missed that feature. Proposing the use of `{@html}` in this case since the api is a trusted source & elastic clearly sanitizing the input.

PR #1102, opened 2024-09-06
Fix [object object] for popovers

jesperengstrom:
### Solves

Checking that the `_label` is a string to prevent printing `[object object]`, currently displaying in the `role` popovers.

example [before](https://beta.libris-dev.kb.se/ztt50fk21529dw1p) / [after](http://localhost:5173/ztt50fk21529dw1p)

---
A bit more in-depth...
Data looking like:

```
"domain": {
   "@id": "https://id.kb.se/vocab/Work"
},
"_label": {
   "@id": "http://www.w3.org/2000/01/rdf-schema#domain",
   "labelByLang": {
      "en": "Domain",
      "sv": "Förekommer på"
   }
}
```

`_label` not being flattened to "Förekommer på", is that a bug? In the next recursion, `domain` content is hidden, since it contains an `@id`.

In any which case, I think the string check makes sense...

### DIFF
--- a/lxl-web/src/lib/components/DecoratedData.svelte
+++ b/lxl-web/src/lib/components/DecoratedData.svelte
@@ -222,7 +222,7 @@
 						data-property={propertyName}
 						class={getStyleClasses(data)}
 					>
-						{#if shouldShowLabels()}
+						{#if shouldShowLabels() && typeof data._label === 'string'}
 							<svelte:element this={block ? 'div' : 'span'}>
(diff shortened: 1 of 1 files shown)


johanbissemattsson (approved):
LGTM

> _label not being flattened to "Förekommer på", is that a bug? In the next recursion, domain content is hidden, since it contains an `@id`.

Yes, this sounds lika a bug in LXLJS and should be resolved there. I have created an issue for it: [LWS-240](https://kbse.atlassian.net/browse/LWS-240).

PR #1103, opened 2024-09-13
LWS-244: Filter sort

jesperengstrom:
### Tickets involved
[LWS-244](https://kbse.atlassian.net/browse/LWS-244)
### Solves

Allowing the user to sort individual filter groups as well as some general changes to facet/filter section. 

### Summary of changes

Each filter group (with the exception of `other`) now gets a sort select element on expansion. The sorting is done in the client, and is persisted after navigation/reload by setting a cookie using the `js-cookie` package. By using cookies (and not localStorage) the setting is available during SSR, which is probably also needed with future user settings until we save them in a db. 

The cookie is retrieved in the hooks `handle` function and passed to the frontend in the server `load` functions. Quite the roundtrip, but i've not come up with a more elegant way to do it...

#### Other changes

* Use the `details`/`summary` elements for the filter panel (in line with rest of the app) to allow expanding the sections without js enabled.
* Added an automatic axe a11y check when filter sections are expanded. This detected contrast violations for the `discriminator` text which i've made darker.
* Replaced the ☑ char with icons for better visibility
* Some other styling tweaks
* Added tests for sort functionality


olovy (approved):
Nice! LGTM!

-------

Some things I would like see in a later PR:
* Suitable default sort order for each facet group (should maybe be a backend config?) 

Things to discuss later:
* Some/top/all? facet groups should be expanded by default?
* Some facet groups might work better as top N + alphabetically, e.g. Language: Swedish, English, German, Afrikaans, Akkandian, Albanian...

jesperengstrom:
> Some things I would like see in a later PR:
> 
> * Suitable default sort order for each facet group (should maybe be a backend config?)

I actually had this implemented ('special' sorting for groups was a var in the constants file) but it was then removed in discussion with Ebou since users can achieve the same thing if they want to. Don't know if that is a correct conclusion. 

Let's merge this for testing and see what can be improved.

PR #1106, opened 2024-09-19
LWS-243: Facet limit info

jesperengstrom:
### Tickets involved
[LWS-243](https://kbse.atlassian.net/browse/LWS-243)

### Solves

Add info text + popover when number of facets in a group reaches its limit + more tweaks

### Summary of changes

* Add limit notification (can be seen for subject or genre/form with a wide search, when clicking 'show more')

~~The limits are set in [the apps file](https://github.com/libris/definitions/blob/develop/source/apps.jsonld) and the current solution has simply copied this to the frontend. Not ideal since sync:ing that will be easily forgotten. Possible other solutions is to a) fetch the file somehow b) Get the limit/limit reached with the api response?~~

* 'No search result' text displayed when searching in filters. This was sort of implemented but not working.

#### Proposing the following additional changes:

To improve readability in filter panel (open for discussion on these 😄 )

* Facet items ellipses after one line instead of breaking + `title` attribute to display in browser tooltip
* Reducing the font size on the filter items 16 -> 14px to fit more text (not on small screens).
* Removing the sigel from the library/bibliography chips. Save some space + means nothing for most people?

### DIFF
--- a/lxl-web/src/colors.css
+++ b/lxl-web/src/colors.css
@@ -11,7 +11,7 @@ see https://tailwindcss.com/docs/customizing-colors#using-css-variables */
 	--color-beige: 247 246 242;
 	--color-pink: 239 214 189;
 	--color-dark-beige: 240 238 233;
-	--color-red: 202 85 85;
+	--color-red: 187 54 54;
 
 	/* Fundament */
 	--color-primary: var(--color-brown);
--- a/lxl-web/src/colors.css
+++ b/lxl-web/src/colors.css
@@ -11,6 +11,7 @@ see https://tailwindcss.com/docs/customizing-colors#using-css-variables */
 	--color-beige: 247 246 242;
 	--color-pink: 239 214 189;
 	--color-dark-beige: 240 238 233;
+	--color-red: 202 85 85;
 
 	/* Fundament */
 	--color-primary: var(--color-brown);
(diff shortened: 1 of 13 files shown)


olovy:
> The limits are set in [the apps file](https://github.com/libris/definitions/blob/develop/source/apps.jsonld) and the current solution has simply copied this to the frontend. Not ideal since sync:ing that will be easily forgotten. Possible other solutions is to a) fetch the file somehow b) Get the limit/limit reached with the api response?

I think we could add the existing [maxItems](https://id.kb.se/vocab/maxItems) to [Slice](https://id.kb.se/vocab/Slice). What do you say @kwahlin ?

kwahlin:
Sounds like an easy thing to do. I can look into it.

olovy:
Nice!
I think the styling changes are an improvement.

The popover doesn't work when the filter panel is modal on smaller screens.
I think it ends up underneath.

Should the limit notification maybe be visible when searching within filters?

jesperengstrom:
> The popover doesn't work when the filter panel is modal on smaller screens. I think it ends up underneath.
> Should the limit notification maybe be visible when searching within filters?

Yep, will fix these + implement new api changes (thanks @kwahlin!)

jesperengstrom:
Fixed:

* Use new api property `maxItems`
* Place popover over modal

(This caused a lot of headache, couldn't get the modal and popover to become part of the same stacking context no matter what, to be able to control them with z-index. Workaround here is a new option to popver, to attach it to the parent element of the target node. This attaches the popover inside the modal, allowing it to be placed in front)

* Improve a11y

(The tooltip needs to be triggered/accessed using the keyboard, and you can't place `tabindex` on a plain `div`. Which means you can really only use tooltips for interacive controls. I've now turned the ⓘ-icon into a useless button that does that. But the difficulty of applying good semantics to this component is kind of telling, we shouldn't really use tooltips for 'footnotes' to text).

olovy (approved):
LGTM! :ship:

PR #1111, opened 2024-09-25
Rename nuxt-app -> idkbse, vue-client -> cat

andersju:
More descriptive directory names. https://kbse.atlassian.net/browse/LXL-4565

(Will change in devops etc before merging)


olovy (approved):
:joy_cat: :ship:

PR #1116, opened 2024-10-02
Feature/bulk change tool additions

lrosenstrom:
* Fix bug where linked entities were shown as external links + their preview card was broken.
* Close side panel when switching between steps (an open side panel made the application crash occasionally).
* Default expanded entities in bulk change view
* Don't scroll items in the preview into view if they are considered recently added


olovy (approved):
LGTM!

PR #1118, opened 2024-10-02
Feature/bulkchanges continued

lrosenstrom:
* Make keyboard shortcut for opening field adder work
* Turn off animations on step change
* Clear undo history on step change
* Smoother scroll into view (same as the rest of the cat client)
* Set preview step to active on save if totalHits > 0
* Warn the user when backing to step 1 from step 2
* Confirm run modal + make it possible to run 'loud'
* Avoid empty toolbar
* Add bulk change as an additional search type + make the word fit


cataloging/src/components/care/bulk-changes.vue
```diff
@@ -646,14 +692,70 @@ export default {
...
+        <div class="Modal-header">
+          <header>
+            {{ translatePhrase('Confirm run') }}
+          </header>
+        </div>
+      </template>
+      <template #modal-body>
+        <div class="Modal-body">
+          <p>
+            {{ translatePhrase('Exportera ändrade poster (uppdatera ändringsdatum)') }}&nbsp
```
olovy, line 740: translate?

cataloging/src/components/care/bulk-changes.vue
```diff
@@ -646,14 +692,70 @@ export default {
...
+            <input
+              :checked="shouldExportAffected"
+              type="checkbox"
+              @change="toggleExportAffected()"/>
+          </p>
+          <div class="Modal-buttonContainer">
+            <button class="btn btn-primary btn--md" @click="doRun()">
+              {{ translatePhrase('Kör') }}</button>
```
olovy, line 748: translate?

olovy, line 748:
```suggestion
              {{ translatePhrase('Run') }}</button>
```

olovy (approved):
:ship:

olovy (commented):
I think it should use the same wording as in the confirmation dialog when showing loud/silent on a finished bulk change.
Something with "Exportera ändrade poster (uppdatera ändringsdatum)"

PR #1121, opened 2024-10-03
LWS-166: Show reproduction instance chip

jesperengstrom:
### Tickets involved
[LWS-166](https://kbse.atlassian.net/browse/LWS-166)

### Solves

`reproductionOf` / `hasReproduction` on resource page show instance cards, which becomes messy.
Don't know if more properties need a similar fix?

before / after [link](http://localhost:5173/6nsmclp44mn98b3w)

Skärmavbild 2024-10-03 kl  13 59 17

/

Skärmavbild 2024-10-03 kl  13 59 44

Displaying `hasReproduction` as block elements

Skärmavbild 2024-10-03 kl  14 02 26

### Summary of changes

Adding another hardcoded exception to the lens selector. Not super happy about that, maybe this 'sublens' thing is the long term solution?

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -716,6 +716,15 @@
 				"fresnel:contentLast": ")"
 			}
 		},
+		"hasReproduction-format": {
+			"@id": "hasReproduction-format",
+			"@type": "fresnel:Format",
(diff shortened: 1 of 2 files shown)


olovy (approved):
LGTM!

> maybe this 'sublens' thing is the long term solution?

Yes, I think it is soon time to investigate that. 
It would also give us the option to use specialized context-specific lenses.

PR #1122, opened 2024-10-04
LWS-100: Add matomo

jesperengstrom:
### Tickets involved
[LWS-100](https://kbse.atlassian.net/browse/LWS-100)

### Solves

Add Matomo to application. Add the following to .env:

```
PUBLIC_MATOMO_URL=https://analytics.kb.se
PUBLIC_MATOMO_ID=88
```

### Summary of changes

* Add `Matomo` component placed on root layout level
* Attach script to head & save tracker in a store, allowing us to track custom events throughout the app. 
* Track page views on load & navigation
* Track sort filters event (just to try out this feature)
* Add CSP exception for analytics


jesperengstrom:
The POST requests to analytics look fine, however no visits are showing on the libris-dev account 🤔.  Maybe a setting somewhere ignoring visits from outside beta-dev. My suggestion is we wait and see when deployed.

jesperengstrom:
@johanbissemattsson in regards to our discussion: using the store with context now.

jesperengstrom:
> The POST requests to analytics look fine, however no visits are showing on the libris-dev account 🤔. Maybe a setting somewhere ignoring visits from outside beta-dev. My suggestion is we wait and see when deployed.

Yes, it was an account setting. Gained admin access and visits are logged from localhost now.

lxl-web/src/lib/components/Matomo.svelte
```diff
@@ -0,0 +1,51 @@
...
+	afterNavigate(async ({ to, from, type }) => {
+		if (type === 'enter' || !$tracker) {
+			await initMatomo();
+		} else {
+			if (from?.url.href) {
+				$tracker.setReferrerUrl(from.url.href);
+			}
+			if (to?.url.href) {
+				$tracker.setCustomUrl(to.url.href);
+			}
```
johanbissemattsson, line 40: Is it a problem that these lines won't be called for the initial navigation (where `initMatamo` is called instead?) Should they also be called after the initialization of the tracker?

lxl-web/src/lib/components/find/FacetGroup.svelte
```diff
@@ -17,6 +19,9 @@
 	import BiCheckSquareFill from '~icons/bi/check-square-fill';
 	import BiSquare from '~icons/bi/square';
 	import BiInfo from '~icons/bi/info-circle';
+	import type { Writable } from 'svelte/store';
+
+	const matomoTracker: Writable<MatomoTracker> = getContext('matomo');
```
johanbissemattsson, line 24: I think it would be preferable to implement a `getTracker` util function (which in turn uses `getContext('matomo')` internally) – to mitigate misspellings of the matomo context string and remove the need for importing the `Writable` and `MatomoTracker` types each time the tracker is used.

lxl-web/src/lib/components/Matomo.svelte
jesperengstrom: I would't say is a problem. In the case of `type === 'enter'` (i.e initial page load) there is no `from.url` so we can't set that manually. (I hope Matomo tracks referrals from external pages automatically, but let's see)

In the case of `setCustomUrl`, it's a way to make sure it's the new url that gets tracked, not the old. In the case of `enter` (`initMatomo`), this is not a problem, it detects the current url correctly.

lxl-web/src/lib/components/find/FacetGroup.svelte
jesperengstrom: Good suggestion. Hope the latest commit does it.

johanbissemattsson (approved):

PR #1124, opened 2024-10-15
Use unspecified type as default in bulk change view

lrosenstrom:
--** Soft type matching **--

This uses a hard coded dummy "Any" type. Maybe it should be` "@type" : "Unspecified"` instead? Maybe it should be a real type, so that the correct label is presented automatically?

TODO: fix "Any" type label in side panel.

IN THE FUTURE: move 'Add local entity' button from property-adder.vue to field.vue?

### DIFF
--- a/cataloging/src/components/care/bulk-changes.vue
+++ b/cataloging/src/components/care/bulk-changes.vue
@@ -692,10 +692,8 @@ export default {
         </div>
       </div>
       <div>
-
-
 <!--        SPECIFICATION-->
-<!--        <pre>{{this.currentBulkChange}}</pre>-->
+<!--        <pre>{{ this.currentBulkChange }}</pre>-->
 <!--        ENTITY FORM-->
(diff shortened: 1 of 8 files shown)


niklasl:
Use [`Resource`](https://id.kb.se/vocab/Resource) (the top class)?

lrosenstrom:
We have discussed using Resource, but came to the conclusion that it's better to use something else. To be honest I don't remember the exact reason but e.g. when adding a local entity where the Type dropdown is present, it should be possible to select either Unspecified or one of specific classes that is associated with the property in question. So e.g. adding a local entity under instanceOf should be unspecified by default but possible to easily select e.g. "Text" etc. Using resource would clutter the dropdown.
Also the property adder does not present all vocab properties out of the box when using Resource, (but I'm not sure if this is a bug). So it seems the getAllVocabProperties addition is needed either way.

PR #1126, opened 2024-10-16
LWS-231: Cookie consent

jesperengstrom:
### Tickets involved
[LWS-231](https://kbse.atlassian.net/browse/LWS-231)

### Solves

Adds cookie consent functionality using [vanilla-cookieconsent](https://cookieconsent.orestbida.com/) package.

Using the same implementation as Tidningar, i.e if user opts out of analytical cookies, we don't track them _at all_.

To test:
* No post requests to matomo.php are being made until user accepts all cookies.
* Post requests to matomo.php are made when user consents to all cookies.
* Requests to matomo.php stops when user redacts their consent (using the settings link in the footer) and matomo cookies (`_pk*`) are cleared.

### Summary of changes

Add cookie package, cookie component & translation files.

### DIFF
--- a/lxl-web/package-lock.json
+++ b/lxl-web/package-lock.json
@@ -8,7 +8,8 @@
 			"name": "lxl-web",
 			"version": "0.0.1",
 			"dependencies": {
-				"dotenv": "^16.4.5"
+				"dotenv": "^16.4.5",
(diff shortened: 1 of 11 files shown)


johanbissemattsson (approved):
LGTM! Seems to work as intended! 🌟

PR #1127, opened 2024-10-18
Matomo tracker load fix

jesperengstrom:
### Solves

I've discovered a bug in the Matomo implementation. Sometimes `initMatomo` is called before the matomo.js script has loaded, causing the whole thing to fail silently and nothing is logged (until a page reload - when the script is probably cached). Try it out by opening a fresh incognito tab and visit beta-dev a couple of times.

### Summary of changes

We need to `setContext` during component initialization, but we can update it with the tracker when the script is ready. Here  using an `on:load` handler. Also doing a `trackPageView` in the init function so we don't miss to track the first page load.


johanbissemattsson (approved):
LGTM!

PR #1130, opened 2024-10-24
Unlock itemOf and heldBy when editing bulkchanges

lrosenstrom:
### Solves

Unlock itemOf and heldBy when editing bulkchanges

Screenshot from 2024-10-24 15-14-47


kwahlin (approved):
:+1:

PR #1131, opened 2024-10-25
New qualifier decorator

jesperengstrom:
* Decorates a valid qualifier key:value pair.
(I noticed there are now multiple qualifier regex:es around the app, maybe we should look into that...)
* Fixes a small annoyance where the expanded search closes when selecting text and releasing the mouse outside of the box.
* Fixes a bug inserting more and more whitespace before a qualifier.
(you can see it by typing `key:value` and then before it, `df df df df df df d df df` 
* Fix broken unit tests + add one


johanbissemattsson (approved):
LGTM! 🌟

PR #1132, opened 2024-10-28
feat(lxl-web): clickable Identifier uri, e.g. ISNI

olovy:
## Description
Display `uri` in `Identifier` as external link.

Example:
bild

bild

### Summary of changes
* Add `uri` to `Identifier-chips`
* Use `uriToId()` to make uri a "fake `@id`" for Identifier
* Only show ext-link arrow on a elements

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -62,7 +62,7 @@
 		@apply mx-auto px-4 sm:px-8;
 	}
 
-	.ext-link::after {
+	a.ext-link::after {
 		content: '\2009↗';
 		@apply align-[10%] text-icon;
 	}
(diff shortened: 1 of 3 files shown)


olovy:
> LGTM!
> 
> > Otherwise the ext-link arrow shows up on things that are not links.
> 
> Or we could change the selector to
> 
> `a.ext-link::after { ... }`
> 
> To make sure it's actually a link? (decoratedData-component checks the presence of `.ext-link` + `@id` when deciding if it should render an `<a>` or `<span>`)

Yes!

niklasl:
Would it be simpler and/or more uniform to cue on the [`uri`](https://id.kb.se/vocab/uri) property having the range `xsd:anyURI`? (This was proposed for the cataloguing client, but I think it still just checks if the value starts with `https?://` (which is brittle)).

PR #1133, opened 2024-10-28
Supersearch: all results link

jesperengstrom:
Display a link in the suggest box to the full search (same as hitting enter).

Visible when there's a value in the input and we don't display "no hits". But that's up for discussion.


johanbissemattsson (approved):
LGTM!

PR #1139, opened 2024-10-31
Add functionality for deleting and creating records in bulk

lrosenstrom:
Make it possible to create and run bulk jobs with other specification types. 

For `bulk:changeSpec` now support `@type`
* 'bulk:Delete'
* 'bulk:Create'


olovy (approved):
:ship:

PR #1144, opened 2024-11-04
LWS-259: Upgrade lxl-web to Svelte 5

johanbissemattsson:
### Tickets involved
[LWS-259](https://kbse.atlassian.net/browse/LWS-259)

### Solves

Upgrades `lxl-web` to Svelte 5. No components are migrated to the new [runes syntax](https://svelte.dev/docs/svelte/what-are-runes) to lessen the scope of this PR.

### Summary of changes

- Upgrade to Svelte 5
- Use `estree-walker` as `walk` is no longer exported from `svelte-compiler`
- Resolve warning regarding self-closing HTML tags for non-void elements
- Resolve `mdsvex` layout relative path problem with monorepo and VSCode
- Fix missing popovers after Svelte upgrade


jesperengstrom:
Getting a load of these in the console relating to the popover component.
Skärmavbild 2024-11-04 kl  17 10 30

Other than that - seems to be working as before!

PR #1145, opened 2024-11-05
LWS-258: Implement supersearch in lxl-web

johanbissemattsson:
### Tickets involved
[LWS-258](https://jira.kb.se/browse/LWS-258)

### Solves

Implements the supersearch component (from `packages/supersearch`) in `lxl-web`. The development of the component is still it's early phase so there's not much to see yet but it's good to be able to test the implementation early on.

The component isn't shown by default – enable it by setting the environment variable `PUBLIC_USE_SUPERSEARCH` to `true` (see the [.env.example](https://github.com/libris/lxlviewer/blob/5ea3546808a96a18a1023911fd514e1a9e0dfc6c/lxl-web/.env.example#L9) file).

Note that changes to `supersearch` are **NOT** automatically rebuilt or repackaged – it's required to run `npm run package`  beforehand to see the changes.  `svelte-package` however accepts [`-w/--watch`](https://svelte.dev/docs/kit/packaging#Options) as an option which watches files in `src/lib` for changes and rebuilds the package.

### Summary of changes
- Implement supersearch
- Add `npm run package` to `prepare` script


jesperengstrom (approved):
LGTM!

PR #1150, opened 2024-11-08
LWS-266: Add Codemirror lang

jesperengstrom:
### Tickets involved
[LWS-266](https://kbse.atlassian.net/browse/LWS-266)

### Summary of changes

Just a skeleton implementation of a Codemirror lang, installed as a dependency and passed to Supersearch as a prop.
The grammar of course needs to be written from scratch.

Should we run the tests as part of Github workflows?


johanbissemattsson:
https://github.com/libris/lxlviewer/pull/1151 adds automatic code formatting and linting so it adheres to the normalized code formatting in `lxl-web` and the `supersearch` package. Should we merge it into this branch?

jesperengstrom:
Yes!

packages/supersearch/src/lib/components/SuperSearch.svelte
```diff
@@ -1,12 +1,19 @@
...
 	import { placeholder } from '@codemirror/view';
+	import { type LRLanguage } from '@codemirror/language';
 
 	let value = $state('');
+	let { language }: { language?: LRLanguage } = $props();
+
+	const extensions = [placeholder('Search')];
+	if (language) {
+		extensions.push(language);
+	}
```
johanbissemattsson, line 12: It's a matter of taste but I prefer conditionally adding the extensions inline [using the spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#conditionally_adding_properties_to_an_object), so you don’t have to be as aware of the code around in the same extent: 

```js
	const extensions = [
                placeholder('Search'),
		...(language ? [language] : [])
	]
```

For me it becomes clearer where all the initially assigned properties come from without having to look through the whole file (vs a quick glance at the assignment) – it isn't a problem right now but can be when the list of extensions grow.

The conditional spread will also play along better with the usage of [compartments](https://codemirror.net/examples/config/#compartments) for dynamic partial reconfiguration, as the original list of extension should be treated more or less as an constant (_it’s not the extensions that changes, it’s the compartments_) to prevent the costly reconfiguration of all extensions.

packages/codemirror-lang-lxlquery/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    include: ['test/test.js']
```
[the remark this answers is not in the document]
jesperengstrom: Yes! Renamed test file and converted to .ts also.

packages/supersearch/src/lib/components/SuperSearch.svelte
jesperengstrom: Good point, fixed now.

johanbissemattsson (approved):
LGTM! 🌟

PR #1152, opened 2024-11-11
Add placeholder compartment/prop

johanbissemattsson:
### Solves

Add placeholder compartment/prop for [partial dynamic configuration](https://codemirror.net/examples/config/#dynamic-configuration) of the placeholder extension.

It acts of an example of how to use CodeMirrors [compartments](https://codemirror.net/examples/config/#compartments).

I'm a little bit torn/unsure if we should treat `editorView` as a [bindable](https://svelte.dev/docs/svelte/$bindable) prop, or if we should expose the current editor view using an event handler function (e.g. `onmount` from the `CodeMirror.svelte` component combined with a `handleMountCodeMirror` function inside `SuperSearch.svelte`). 

The bindable approach will result in fewer lines of code but we probably don't want the bindable prop for it's two-way binding abilites, as we should generally be careful to mutate the CodeMirror related data directly (dispatching changes is totally fine!). At least that's the case when working with history.

### Summary of changes

- Add placeholder compartment/prop

### DIFF
--- a/packages/supersearch/src/lib/components/SuperSearch.svelte
+++ b/packages/supersearch/src/lib/components/SuperSearch.svelte
@@ -1,16 +1,39 @@
 <script lang="ts">
 	import CodeMirror, { type ChangeCodeMirrorEvent } from '$lib/components/CodeMirror.svelte';
-	import { placeholder } from '@codemirror/view';
+	import { EditorView, placeholder as placeholderExtension } from '@codemirror/view';
+	import { Compartment } from '@codemirror/state';
 	import { type LRLanguage } from '@codemirror/language';
 
-	let value = $state('');
-	let { language }: { language?: LRLanguage } = $props();
+	interface Props {
+		language?: LRLanguage;
+		placeholder?: string;
+	}
 
-	const extensions = [placeholder('Search'), ...(language ? [language] : [])];
+	let { language, placeholder = '' }: Props = $props();
+
+	let value = $state('');
+	let editorView: EditorView | undefined = $state();
+
+	let placeholderCompartment = new Compartment();
+	let prevPlaceholder = placeholder;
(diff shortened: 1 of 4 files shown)


packages/supersearch/src/lib/components/SuperSearch.svelte
```diff
@@ -1,12 +1,34 @@
...
 	function handleChangeCodeMirror(event: ChangeCodeMirrorEvent) {
 		value = event.value;
 	}
+
+	$effect(() => {
+		if (placeholder !== prevPlaceholder) {
+			editorView?.dispatch({
+				effects: placeholderCompartment.reconfigure(placeholderExtension(placeholder))
+			});
+			prevPlaceholder = placeholder;
```
johanbissemattsson, line 29: It's a little bit cumbersome having to add an equality check on `placeholder` to prevent the effect from firing on mount – but I cannot think of a better way right now... suggestions are appreciated! :)

johanbissemattsson:
It's a little bit cumbersome having to add an equality check on placeholder to prevent the effect from firing on mount (similar checks will probably be needed on all compartment effects) – but I cannot think of a better way right now... suggestions are appreciated! :)

packages/supersearch/src/lib/components/SuperSearch.svelte
```diff
@@ -1,12 +1,34 @@
...
+		}
```
johanbissemattsson, line 30: It's a little bit cumbersome having to add an equality check on placeholder to prevent the effect from firing on mount (similar checks will probably be needed on all compartment effects) – but I cannot think of a better way right now... suggestions are appreciated! :)

jesperengstrom (approved):
LGTM!

PR #1155, opened 2024-11-12
LWS-260: Codemirror extension for submitting forms on enter key

johanbissemattsson:
### Tickets involved
[LWS-260](https://kbse.atlassian.net/browse/LWS-260)

### Solves

Adds an CodeMirror extension that submits form elements (either the closest or by specified id using the `form` attribute) on enter keypresses. We need the extension to bypass the default behaviour of adding a new line on Enter keypresses – if a form is found, it should be submitted instead.

### Summary of changes

- Add CodeMirror extension for submitting closest form on enter keypress
- Add tests
- Add hidden textarea inside component for `name` and `form` attributes


jesperengstrom:
Nice!

(See inline comment)

packages/supersearch/src/lib/extensions/submitFormOnEnterKey.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param {string} form Optional id of the `<form>` element with which the form control should be associated with (equivalent with
+ * the [form attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form) on HTML Input elements).
+ */
+
+const submitFormOnEnterKey = (form?: string) => {
+	const submitForm = (editorView: EditorView) => {
+		const formElement = form ? document.getElementById(form) : editorView.dom?.closest('form');
+
+		if (formElement && formElement instanceof HTMLFormElement) {
+			formElement.submit();
```
jesperengstrom, line 16: You notice a difference now with the `supersearch` flag/on off. With supersearch, the progress indicator is missing (`navigating` clearly not firing) and you can submit an empty form (`on:submit` handler not running). 

"form.submit() method is similar, but not identical to, activating a form's submit button"

`form.requestSubmit()` should do the trick?

see https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit

johanbissemattsson: Good call! This should be resolved in https://github.com/libris/lxlviewer/pull/1155/commits/058d454138214540d9717d512b1cb0bdb310b1cc

jesperengstrom (approved):
:ship:!

PR #1159, opened 2024-11-13
LWS-264: Lxlquery grammar and highlighting

jesperengstrom:
### Tickets involved
[LWS-264](https://kbse.atlassian.net/browse/LWS-264)

### Summary of changes

* Add rudimentary language rules to the grammar file
* Add tests for the grammar
* Parse the language with custom lezer `Tags` for certain nodes
* Add codemirror extension that adds CSS classes when encountering those tags
* Add a css file to style those classes, achieving syntax highlighting

try some searches from the help page:
`pippi långstrump språk:(engelska OR franska) ÅR>2002]`
`språk:svenska ÅR>2010 ÅR<=2024 ämne:"sao:Drakar"`
`träd* bibliografi:"sigel:DST" NOT typ:Text`

Or a 'special' qualifier
`includeEplikt`

Notes:
* The (somewhat) straighforward way of highlighting using nodes -> tags -> classes/styles seems to be the cheapest (compared to iterating through the syntax tree adding decorators - even though we'll be doing that later on) but is limited in that it does not produce nested structures, only an inline element for the most specific exported node in a given place. On the other hand, I don't know if we want/need highlighting at all. But for now I think it's nice for evaluating/debugging purposes. The styling itself is not carefully done :)

* The highlighter extension now belongs to lxlweb. Since it's language specific, it doesn't belong to supersearch and is not a core part of the language either. It imports stuff from codemirror/lezer that lxlweb does not directly depend on, don't know if that's a problem?

(Another solution could be to let the language package export the Highlighter, import it in lxlweb and pass it as a special prop (how many can we handle?) to supersearch that then turns it into an extension.)


lxl-web/src/lib/utils/LxlQueryHighlight.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const highlighter = tagHighlighter([
+  { tag: t.Qualifier, class: 'qualifier' },
+  { tag: t.QualifierKey, class: 'qualifierkey' },
+  { tag: t.QualifierValue, class: 'qualifiervalue' },
+  { tag: t.EqualOperator, class: 'equaloperator' },
+  { tag: t.CompareOperator, class: 'compareoperator' },
+  { tag: t.BooleanOperator, class: 'booleanoperator' },
+  { tag: t.BooleanQuery, class: 'booleanquery' },
+  { tag: t.Wildcard, class: 'wildcard' }
```
johanbissemattsson, line 13: I think some kind of separation between the words is good for legibility. Sorry for nitpicking!

It's up for discussion but I have a slight preference for kebab-case in CSS classes (e.g. as on [MDN](https://developer.mozilla.org/)) – mainly to separate it from the default camelCase convention in JavaScript.

johanbissemattsson (commented):
🌟🌟🌟 I think it would be good to go through the grammar together with the rest of the team (can be done after the PR is merged) but I think it looks really good!

lxl-web/src/lib/utils/LxlQueryHighlight.ts
jesperengstrom: > I have a slight preference for kebab-case in CSS classes

Mee too, somehow it got lost here!

lxl-web/src/lib/utils/LxlQueryHighlight.ts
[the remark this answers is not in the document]
johanbissemattsson: I'm in favour of the first alternative. LGTM!

PR #1160, opened 2024-11-13
fix(lxlweb): prevent error when missing matomo-id

jesperengstrom:
### Solves

As a dev you might want to comment out `PUBLIC_MATOMO_ID` so your lousy test searches don't end up in analytics. But doing so result in console errors related to the cookie consent component. This allows us to gracefully disable the tracker.

### Summary of changes

* Add checks before running methods on `$matomoTracker`

### DIFF
--- a/lxl-web/src/lib/components/CookieConsent.svelte
+++ b/lxl-web/src/lib/components/CookieConsent.svelte
@@ -35,17 +35,15 @@
 			}
 		},
 		onConsent: ({ cookie }) => {
-			if ($matomoTracker && cookie.categories.includes('analytics')) {
-				$matomoTracker.rememberConsentGiven();
+			if (cookie.categories.includes('analytics')) {
+				$matomoTracker?.rememberConsentGiven();
 			}
 		},
 		onChange: ({ cookie }) => {
-			if ($matomoTracker) {
-				if (cookie.categories.includes('analytics')) {
-					$matomoTracker.rememberConsentGiven();
-				} else {
-					$matomoTracker.forgetConsentGiven();
-				}
+			if (cookie.categories.includes('analytics')) {
+				$matomoTracker?.rememberConsentGiven();
+			} else {
+				$matomoTracker?.forgetConsentGiven();
 			}
 		},
 		language: {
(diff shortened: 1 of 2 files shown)


lxl-web/src/lib/components/CookieConsent.svelte
```diff
@@ -35,15 +35,17 @@
...
 		onChange: ({ cookie }) => {
-			if (cookie.categories.includes('analytics')) {
-				$matomoTracker.rememberConsentGiven();
-			} else {
-				$matomoTracker.forgetConsentGiven();
+			if ($matomoTracker) {
+				if (cookie.categories.includes('analytics')) {
+					$matomoTracker.rememberConsentGiven();
+				} else {
+					$matomoTracker.forgetConsentGiven();
```
johanbissemattsson, line 47: Not much of a difference but it feels like optional chaining could be used here instead of explicitly checking for `$matomoTracker` (e.g. `$matomoTracker?.rememberConsentGiven();`.

lxl-web/src/lib/contexts/matomo.ts
```diff
@@ -30,6 +30,8 @@ export function setMatomoTracker() {
 	if (initializedMatomoTracker) {
 		tracker.set(initializedMatomoTracker);
 		console.info('Matomo tracker set');
+	} else {
+		console.info('Matomo is disabled');
```
johanbissemattsson, line 34: I think the console info log could have a slightly more specific message (e.g. `Missing Matomo id`) telling why it is disabled.

johanbissemattsson (approved):

johanbissemattsson (approved):
LGTM!

lxl-web/src/lib/components/CookieConsent.svelte
jesperengstrom: Duh. Of course...

lxl-web/src/lib/contexts/matomo.ts
jesperengstrom: > Or maybe no console log at all when it's not set (to prevent having too much information to sift out)?

This will hopefully not happen in production?

> Users could maybe also become unnecessarily annoyed by the Matomo tracker set message

Users don't look at the console?

But ok, this was more of a debugger thing when Matomo implementation was shaky. Seems to be working now, so I don't feel very strongly about it either way :)

PR #1165, opened 2024-11-14
chore: Run husky and lint-staged from git repo root

johanbissemattsson:
### Solves

Runs [husky](https://typicode.github.io/husky/) and [lint-staged](https://github.com/lint-staged/lint-staged) from the git repo root instead of juggling with different command paths for the different projects. Usage of husky is removed during CI (as it's not needed there).

### Summary of changes

- Run husky and lint-staged from git repo root
- Prevent usage of husky in CI


jesperengstrom (approved):
Nice, `lint-staged` runs now when committing something in supersearch or lxlquerylang.  
🚢

PR #1167, opened 2024-11-18
LXL-4560: use vue-matomo

jesperengstrom:
### Tickets involved
[LXL-4560](https://kbse.atlassian.net/browse/LXL-4560)

### Solves

Adds [vue-matomo](https://www.npmjs.com/package/vue-matomo) package. It's a small dependency allowing us to integrate Matomo in a more 'vue-way' instead of injecting the script ourselves. It also tracks page views automatically on navigation using the router, something we'd have to do manually otherwise (and has been missing so far, only first page load is tracked now).

### Summary of changes
- Add & implemement vue-matomo
- Remove old code


johanbissemattsson (approved):
LGTM!

PR #1171, opened 2024-11-19
feat(supersearch): Improved handling of double-click and triple-click for text selection

johanbissemattsson:
### Tickets involved
[LWS-267](https://kbse.atlassian.net/browse/LWS-267)

### Solves
Improves the handling of text-selection when double-clicking and triple-clicking on text in the collapsed editor.

Double-clicks results in the selection of [the word at the given position](https://codemirror.net/docs/ref/#state.EditorState.wordAt) while triple-clicks selects the entire line.

### Summary of changes

- Improve handling of double-click and triple-clicks for text selection


johanbissemattsson:
I think we should keep this PR unmerged for now to keep the code as simple as possible (If the expanded search text field is placed at the same place as the collapsed, text-selection by double-clicking and triple-clicking will work mostly natively out-of-the-box).

We can merge it if/when we decide that we need to support cases where the expanded search text box is placed somewhere else.

PR #1172, opened 2024-11-19
LWS-263: Supersearch pills

jesperengstrom:
## Description

An early draft to discuss if this could be a way to pursue or will end up in a world of pain.

I've created a plugin with widget decorations that iterates the node tree and, upon finding a `qualifier`, breaks down its parts and passes them to a svelte component (pretty much like before). 

The new thing here is that the component essentially steals the focus from codeMirror and lets us type the value from there, and then updates it back to cm.

This aims to solve for example: 
* typing inside pills
* Typing Astrid Lindgren but treating it as "Astrid Lindgren" (adding quotes that we can later hide)
* Prevent editing values that are display labels
* etc

... maybe this can all be done within codemirror, but would propbaby require many different decorations etc for different parts.

... Also, the term 'qualifier' is now being hardcoded into supersearch. Maybe it's reasonable that we require other languages to use this term?

### DIFF
--- a/packages/codemirror-lang-lxlquery/src/syntax.grammar
+++ b/packages/codemirror-lang-lxlquery/src/syntax.grammar
@@ -12,8 +12,8 @@ term {
 Group { "(" term* ")" } 
 
 BooleanQuery { 
-  (freetext | Qualifier) BooleanOperator (freetext | Qualifier) 
-  (BooleanOperator (freetext | Qualifier))+?
+  (freetext | Qualifier) booleanOperator (freetext | Qualifier) 
+  (booleanOperator (freetext | Qualifier))+?
 }
 
 Qualifier {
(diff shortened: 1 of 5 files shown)


johanbissemattsson:
Interesting approach! I have to think about it some more!

If we can get achieve the expected user experience behaviours (e.g. allowing the cursor to jump left _out_ of the decoration when pressing left _inside_ the escaped decoration) then it could absolutely work.

>... maybe this can all be done within codemirror, but would propbaby require many different decorations etc for different parts.

Would we really need so many different decorations?

I'm thinking we could maybe (perhaps with some wishful thinking) get by with only two decorations/components for the qualifiers: `QualifierKey` (always atomic) and `QualiferValue` (sometimes atomic, sometimes not, depending on the value). They could be pretty flexible using props (as the passed data should be enough to decide how they should be rendered)?

Or would we also need a specific decorator for atomic qualifier values (e.g. `QualifierLinkedValue` or something similar... 🤔

jesperengstrom:
Replaced by https://github.com/libris/lxlviewer/pull/1176

PR #1179, opened 2024-11-27
LWS-263: Supersearch pills adjustments/suggestions

johanbissemattsson:
### Tickets involved
[LWS-263](https://kbse.atlassian.net/browse/LWS-263)

### Solves

Some smaller suggestions/additions to the supersearch pills. I tried doing some changes to the syntax grammar but we have to discuss the best way forward. Grammar tests are currently failing due to my grammar changes...

### Summary of changes

- Keep extension and extension-specific components together (in the same folder).
- Export qualifier plugin from supersearch package. This can also help with easier event handling in the implementations.
- Rename qualifier plugin (to clarify to the connection to the Libris XL query language)
- Remove sveltekit specific import (removes the need to add sveltekit as a peer dependency)
- Add props typing
- Add lxlquery language and qualifier plugin to demo app
- Move insert quotes transaction filter to separate file (as the qualifier file already is quite long)
- Add label prop to qualifier key
- Capitalize lezer rules for qualifiers (this needs to be discussed, which is the best way forward?)
- Use visibleRanges to only render widgets when visible
- Prepend qualifier classes with lxl


jesperengstrom (approved):
LGTM!

PR #1180, opened 2024-11-28
Fix helpdocs heading ID generation

andersju:
The fragment links on the help pages stopped working because `marked` moved heading ID generation to a separate package: https://www.npmjs.com/package/marked-gfm-heading-id

https://kbse.atlassian.net/browse/LXL-4589


olovy (approved):
:ship:

PR #1181, opened 2024-11-28
fix(cataloging): Fix fragment link scroll issue in Firefox

andersju:
In Firefox if you go to e.g. https://libris-dev.kb.se/katalogisering/help/workflow-work you have to click on "Verkstyp" twice for it to actually scroll there. Not an issue in Chrome. This seems to fix it.


olovy (approved):
:ship: :ship: :ship: :ship: :ship: 
`¯\_(ツ)_/¯`

PR #1183, opened 2024-11-28
fix(supersearch): Allow deletion of inserted quotes

jesperengstrom:
### Solves

Quotes are inserted when user has typed a qualifier and the value is still empty. But weirdness happens when user is trying to delete the quotes and we try to add them again. This allows deletion by only adding them on input.

(Maybe in the future we hide the quotes from the user altogether. But for now, this seems less buggy)


johanbissemattsson (approved):
Awesome! It works much better now! 🚢

PR #1186, opened 2024-12-09
LWS-274: labels for Supersearch

jesperengstrom:
### Tickets involved
[LWS-274](https://kbse.atlassian.net/browse/LWS-274)

### Solves

Add labels to qualifiers by passing in a callback function to the `lxlQualifierPlugin`. The function is called with any found qualifier `key`/`value` pair, and should return their labels and an optional remove ('up') link. If found, a widget decoration (`QualifierComponent`) will replace them and an atomic range will be created. 

In the lxlweb implementation (no implementation for the Supersearch demo route so far), the callback function iterates `search.mappings` of `$page.data` and autosuggest fetches, suggest labels taking precedence. Page data is needed to have labels ready when page loads, suggest mappings are needed to get labels instantly when typing.

As you'll notice, at this point this has quite a few limitations and bugs left to fix. I think cooperation and iteration will be the best way to proceed from here.

### Summary of changes
* Svelte 5-migration of `lxlweb/Search` component
* Update codemirror state on lxlweb navigation, needed for facet filters to work
* Adjust response from server.ts in `/find` and `/api/lang/supersearch` (needs improvement)
* Add `getLabelsFromMapping` util function
* Refactor `lxlQualifier` plugin, now using one single `QualifierComponent` widget + additional mark decorations.

#### Notes on the above
* ~~Syncing Codemirror with url searchParams is now done using `afterNavigate` which is a svelte/kit import and should be done differently. Suggestions are welcome. Dunno if it's even part of this PR, should be added soon anyhow...~~
* The fact that we are not looping out the contents of `search.mapping` but rather matching a piece of the `_q` string with a single `DisplayMapping` object _may_ require further additions to the search api. I've not come up with a way to decisively match the two using today's response. Current implementation is hacky and resulting in `rdf:type` not getting labels, for example.
* In the case of a valid key together with an 'unlinked' value (`publicationYear:2001`, `contributor:"Astrid"`), the value gets a special styling with no remove link and continues to be editable. One of many things we'll have to try out and discuss.

#### bugs/limitations
* Because of the matching problem described above, two qualifiers/filters of the same kind now gets the same label. Maybe it can be completely solved in the frontend, maybe not. Will look into it some more.
* ~~Because of the nature of the current implementation + suggestion debounce, labels from suggest mapping will not magically appear when resolved, but rather when the function is called the next time. This usually means the labels appear only after typing the first letter of the value, which is not ideal.~~
* ~~Labels can also disappear after appearing, usually because suggestion gets an empty response. Maybe we'll have to save the last successful response to avoid this.~~
* Qualifiers following an invalid qualifier will never get their label because of the error response caused by the invalid property. Unless we choose a completely different approach, this will only be fixed by new error handling done in the search API.

#### Todos
* ~~Write tests~~
* ~~Document Supersearch api/prop changes~~


johanbissemattsson:
I think this looks very promising! I will do a more formal review but first a quick reply regarding:
> Syncing Codemirror with url searchParams is now done using afterNavigate which is a svelte/kit import and should be done differently. Suggestions are welcome. Dunno if it's even part of this PR, should be added soon anyhow...

One way of achieving this is can be by modifying the [$effect function which handles value changes inside SuperSearch.svelte](https://github.com/libris/lxlviewer/blob/20e33b23353b5f1fd98b06e8f94736613c2098fc/packages/supersearch/src/lib/components/SuperSearch.svelte#L60-L64).

If we dispatch changes when the value differs from the current document state in the editor views, then changes to to URL should also be reflected in the editors.

```js

	$effect(() => {
		if (value) {
			search.debouncedFetchData(value);
		}

		if (value !== collapsedEditorView?.state.doc.toString()) {
			collapsedEditorView?.dispatch({
				changes: {
					from: 0,
					to: collapsedEditorView.state.doc.length,
					insert: value
				}
			});
			expandedEditorView?.dispatch({
				changes: {
					from: 0,
					to: expandedEditorView.state.doc.length,
					insert: value
				}
			});
		}
	});
```

It's a little bit unfortunate that we need to dispatch the changes to both the expanded and collapsed editor in the example above, and it would also probably be preferable to use the [reset function](https://github.com/libris/lxlviewer/blob/20e33b23353b5f1fd98b06e8f94736613c2098fc/packages/supersearch/src/lib/components/CodeMirror.svelte#L72C2-L78C3) to ensure the editor state is resetted between navigations. We probably need to do some minor rewiring as `collapsedEditorView` and `expandedEditorView` is only bound to the editor view,s not the component itself (which has the exported reset function).

jesperengstrom:
Searchparam sync without afterNavigate should in be fixed in https://github.com/libris/lxlviewer/pull/1186/commits/12766964c05af407001135b36c59995191c662b7

jesperengstrom:
> Because of the nature of the current implementation + suggestion debounce, labels from suggest mapping will not magically appear when resolved, but rather when the function is called the next time. This usually means the labels appear only after typing the first letter of the value, which is not ideal.

I've (hopfully) fixed this in https://github.com/libris/lxlviewer/pull/1186/commits/e8407227932e4309e2b9d98ee0fd78fbe7a73acb

By using an `$effect` in Supersearch.svelte, we can monitor when search data comes back, passing this notification on to the plugin using `StateEffect` and trigger a re-rendering of decorations.

The same approach could be used to pass on a loading state, etc to the editors.

I've experimented with commenting out [this line](https://github.com/libris/lxlviewer/blob/e8407227932e4309e2b9d98ee0fd78fbe7a73acb/packages/supersearch/src/lib/extensions/lxlQualifierPlugin/index.ts#L157), i.e. only adding decorations (looking for labels etc) when new data is available, not on every keystroke. It should be great for performance. But to look good, instant styling of qualifiers must come from the Tag highlighter extension included in the language instead. Will work on it some more...

lxl-web/src/lib/components/Search.svelte
```diff
@@ -54,10 +66,23 @@
...
+		// to not hardcode lxl-needs into useSearchRequest
+		// could also be new separate callback
+		suggestMapping = data?.mapping;
+		return data;
+	}
+
+	function getLabels(key: string, value?: string) {
+		let pageMapping = $page.data.searchResult?.mapping;
+		return getLabelFromMappings(key, value, pageMapping, suggestMapping);
+	}
```
johanbissemattsson, line 81: This could be rewritten into a derived state:
```js
let derivedLxlQualifierPlugin = $derived.by(() => {
	function getLabels(key: string, value?: string) {
		let pageMapping = $page.data.searchResult?.mapping;
		return getLabelFromMappings(key, value, pageMapping, suggestMapping);
	}
	return lxlQualifierPlugin((getLabels));
});
```
Which can then be used further down in the file (where the extensions are set on the `<SuperSearch>` component):
```extensions={[derivedLxlQualifierPlugin]}```


It results in two more lines of code but can maybe clarify the scoping of `getLabels` and `lxlQualifierPlugin`.

What do you think?

File: packages/supersearch/src/lib/components/SuperSearch.svelte [hunk 2]
```diff
@@ -63,6 +63,16 @@
 		}
 	});
 
+	const searchStatus = StateEffect.define<{ message: string }>({});
+
+	$effect(() => {
+		if (search.data) {
+			const effects = { effects: searchStatus.of({ message: 'new_data' }) };
```
johanbissemattsson, line 70: I'm thinking we maybe could clarify what `new_data` is? E.g. `new_response_data` or `new_mapping_data` or something else? But `new_data` is also fine...

johanbissemattsson (commented):
Awesome work! 🙌

I think it looks really good (especially the changes where getQualifiers is now only triggered on new data)! It should help the performance quite alot 🌟

lxl-web/src/lib/components/Search.svelte
```diff
@@ -54,10 +66,23 @@
...
 		return undefined;
 	}
+
+	function handleTransform(data) {
+		// hijacking this function to set get autosuggest labels
+		// to not hardcode lxl-needs into useSearchRequest
+		// could also be new separate callback
+		suggestMapping = data?.mapping;
+		return data;
+	}
```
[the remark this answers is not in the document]
jesperengstrom: 👍 
Removed the comment.

packages/supersearch/src/lib/components/SuperSearch.svelte
jesperengstrom: Agreed. Also a bit shaky with the arbitrary string. In https://github.com/libris/lxlviewer/pull/1186/commits/39d57bbc63a892e6cac0a308516366e7ab4bb0dc I export a constant for messages, so we don't fail because of spelling errors in one place.

jesperengstrom:
> I think it looks really good (especially the changes where getQualifiers is now only triggered on new data)! It should help the performance quite alot 🌟

Nice, however I'm beginning to think this does not cut it. There is a noticeable lag from the widgets when you type now. Removing a widget by backspace feels kind of buggy because of this. Maybe there's some middle ground here, updates should maybe be triggered by _some_ input after all. Maybe this does not have to be 100% resolved at this point...

lxl-web/src/lib/components/Search.svelte
jesperengstrom: Nice, looks cleaner!
Is the double parenthethis in the return statement intentional?

jesperengstrom: Applied this ☝️

johanbissemattsson: > Is the double parentheses in the return statement intentional?

Oups, not at all! 🙃

packages/supersearch/src/lib/components/SuperSearch.svelte
jesperengstrom: Double effects should be fixed in https://github.com/libris/lxlviewer/pull/1186/commits/8f8cb5a83126155e92d9786e5dc6543b190a2905

johanbissemattsson (approved):
LGTM! 🚢

PR #1195, opened 2024-12-20
fix(lxlweb): Continue showing error page for _invalid when not using Supersearch

jesperengstrom:
### Solves

Frontend tests were failing because of API changes added in https://github.com/libris/librisxl/pull/1541
Basically the lack of `@id` for an invalid property caused the 500 error page to show instead of expected 400.

However, when fixing this and using 'old search' the resulting page for an invalid query is now this:
Skärmavbild 2024-12-20 kl  10 08 56

We figured as long as supersearch is not fully implemented, users still want some kind of feedback when using invalid queries. Therefore manually (temporarily) showing the same error page as before. This also makes tests pass.

ping @olovy @kwahlin 

### Summary of changes

* Add interface for `InvalidProperty`
* Add `@id` check in search.ts
* In `search.ts`, when encountering an `_invalid` property, show error page with same message as before.

### DIFF
--- a/lxl-web/src/lib/utils/search.ts
+++ b/lxl-web/src/lib/utils/search.ts
@@ -96,8 +96,7 @@ export function displayMappings(
 				// when encountering an invalid property in order to provide feedback.
 				// TODO remove this when Supersearch is fully implemented.
 				const useSuperSearch = env?.PUBLIC_USE_SUPERSEARCH === 'true';
-				const _propertyType = m.property?.['@type'];
-				if (!useSuperSearch && _propertyType === '_Invalid') {
+				if (!useSuperSearch && m.property?.['@type'] === '_Invalid') {
 					error(400, {
 						message: `Invalid query, please check the documentation. Unrecognized property alias: ${m.property?.label ?? ''}`
 					});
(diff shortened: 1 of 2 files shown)


lxl-web/src/lib/utils/search.ts
```diff
@@ -90,6 +92,17 @@ export function displayMappings(
 			const operator = _hasOperator(m);
 
 			if ('property' in m && operator) {
+				// Mock old behaviour for 'classic' search GUI, i.e show error page
+				// when encountering an invalid property in order to provide feedback.
+				// TODO remove this when Supersearch is fully implemented.
+				const useSuperSearch = env?.PUBLIC_USE_SUPERSEARCH === 'true';
+				const _propertyType = m.property?.['@type'];
+				if (!useSuperSearch && _propertyType === '_Invalid') {
```
johanbissemattsson, line 100: I would skip the const declaration on the line above and just write:
```js
if (!useSuperSearch && m.property?.['@type'] === '_Invalid') {
```
(as `_propertyType` isn't used anywhere else).

Looks good otherwise!

johanbissemattsson (approved):
LGTM! (see comment)

lxl-web/src/lib/utils/search.ts
jesperengstrom: Sure, was trying to fix the typescript complaint and somehow it stayed in there. Fixed.

PR #1196, opened 2024-12-20
fix(supersearch): Implement the _Invalid type (LWS-287)

jesperengstrom:
[LWS-287](https://kbse.atlassian.net/browse/LWS-287)

### Solves

With https://github.com/libris/librisxl/pull/1541, we now have a way to decisively say a property is invalid, and can have the red squiggly line to mean that, not just that no label was found for it.

This can so far only be tested with dev backend.

### Summary of changes

* Forward `@type: '_Invalid'` as a new key `invalid` to the frontend. Update `DisplayMapping` type accordingly.
* In the `GetLabelFunction`, match `invalid` for a certain key and return an invalid boolean. 
* Add a mark decoration if `invalid`.
* Update css somewhat.

### DIFF
--- a/lxl-web/src/lib/styles/lxlquery.css
+++ b/lxl-web/src/lib/styles/lxlquery.css
@@ -29,8 +29,9 @@
 }
 
 .invalid > .lxl-qualifier-key {
-	text-decoration: underline 2px solid;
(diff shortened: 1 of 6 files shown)


johanbissemattsson (approved):
Awesome! Seems to work as intended! 🌟

PR #1197, opened 2024-12-20
feat(supersearch): Keyboard handling (LWS-277) 

johanbissemattsson:
### Tickets involved
[LWS-277](https://kbse.atlassian.net/browse/LWS-277)

### Solves

Keyboards handling for the supersearch component.

The handling of the keyboard interaction tries to adhere to the [Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) in W3Cs ARIA Authoring Practice Guide (see: [Editable Combobox with Grid Popup](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/grid-combo/)) but further real-life testing will be needed to ensure the accessiblity is up to par.

### Summary of changes

- Support toggling expanded search using keyboard shortcut
- Add keyboard handling for rows and cells in expanded search
- Fire click events on focused grid cells when pressing enter
- Use [esm-env](https://www.npmjs.com/package/esm-env) instead of [SvelteKit-specific `$app/environment`](https://svelte.dev/docs/kit/packaging#Best-practices)
- Update readme
- Remove unused `submitFormOnEnterKey` util
- Fix `debouncedWait` prop
- Implement keyboard handling in `lxl-web`

Saved for later PRs:
- Add support for persistant result items shown before the rest of the search results (also when results are loading).
- Add `onfocuscell` event handlers (could be useful for preloading page data on focus).
- Loading snippet prop for custom loading indicators


packages/supersearch/src/lib/components/SuperSearch.svelte
```diff
@@ -153,53 +344,64 @@
...
+								rowIndex
+							)}
+						</div>
+					{/each}
+				</div>
+			{/if}
+			{#if search.isLoading}
+				Loading...
+			{:else if search.hasMorePaginatedData}
+				<li>
```
jesperengstrom, line 385: this `<li>` element seems to have no parent `<ul>` or `<ol>`?

johanbissemattsson: Oups my bad! Fixing it now!

jesperengstrom:
Really awesome work! 🙌 

Works great as far as I can tell, tried it in supersearch/lxlweb, also with screen reader.
Ran it through some a11y testers too (since it's such a focal point here) and they spotted some things. Wouldn't call them show stoppers, but we might want to fix it to keep our scores up:

* A `role="textbox"` element can't have `aria-expanded`
* A `role="combobox"` element can't have `aria-multiline`
* The expanded search `aria-activedecendant` is initially set (to `supersearch-result-item-0x0` for example) before the element exists in the DOM.

johanbissemattsson:
> Really awesome work! 🙌
> 
> Works great as far as I can tell, tried it in supersearch/lxlweb, also with screen reader. Ran it through some a11y testers too (since it's such a focal point here) and they spotted some things. Wouldn't call them show stoppers, but we might want to fix it to keep our scores up:
> 
>     * A `role="textbox"` element can't have `aria-expanded`
> 
>     * A `role="combobox"` element can't have `aria-multiline`
> 
>     * The expanded search `aria-activedecendant` is initially set (to `supersearch-result-item-0x0` for example) before the element exists in the DOM.

Thanks for doing such a thorough accessibility review! 🌟 

https://github.com/libris/lxlviewer/pull/1197/commits/81729df4db00c748af16a40cadf11d59a975d6d0 should hopefully resolve the issues listed above.

johanbissemattsson:
Doh, the last commit caused tests to fail so will have to update them (will do tomorrow).

jesperengstrom (approved):
🚢 !

PR #1199, opened 2025-01-14
feat(lxlquery): Improve grammar + lint

jesperengstrom:
### Tickets involved
[LWS-292](https://kbse.atlassian.net/browse/LWS-292)

### Solves

Upon discovering that the frontend language parsing was allowing stuff that the backend don't and vice versa, I tried to fill in the gaps in our grammar, but ended up rewriting the whole thing after inspecting the backend code.

see [source](https://github.com/libris/librisxl/blob/7990106ab2b5e5cfa49eee843c09847869e0d5b3/whelk-core/src/main/groovy/whelk/search2/parse/Parse.java#L7)

The root node now being `orcomb` (not exported), this exports pretty much the same nodes as before, but with more robust logic and less code. For example:

Used to be valid but is invalid:
`key>(sommar AND vinter)`

Used to be invalid but is valid:
`!(sommar AND vinter)`

The query language is clearly not settled (whitespace discussion), but with frontend/backend in sync, we can (must!) do parallel changes moving forward.

Instead of rewriting all of our tests, I replaced many of them with the backend tests most relevant for our use case.

#### Lint

To spot valid/invalid queries, I tried adding mark decorations to error nodes. This feels pretty useful for us and maybe for the end user too. So i ended up doing a linting extension that is now bundled with the language. This simply highlights syntax errors from the language + a couple of manual cases (next step can be to use Codemirror lint package for tooltips etc). Try some invalid queries:

`()`
`AND sommar`
`"Astrid Lindgren`
`NOT sommar NOT OR vinter`
`code:("something" and code:"whatever")`
`sommar!` (pretty strange that this is invalid, but as long as it is, the frontend should point it out)

#### NOT

The language now exports inequality term nodes, which can be used to highlight excluded terms. I added very basic styling, but this needs improvement if we want to handle double negatives (NOT NOT) for example, or style pills like we do today:

Skärmavbild 2025-01-14 kl  10 56 31

### Summary of changes

* Rewrite grammar
* Add lxlLinter extension for language
* Update (still rudimentary) styling
* Rewrite tests
* Fixed a bug in insertQuotes that inserted quotes for 'special' qualifiers without key/value such as `includeEplikt`

### DIFF
--- a/lxl-web/src/lib/styles/lxlquery.css
+++ b/lxl-web/src/lib/styles/lxlquery.css
@@ -34,9 +34,7 @@
 }
 
 .lxl-invalid,
-.lxl-invalid .lxl-qualifier,
-/* todo rename/remove */
-.invalid > .lxl-qualifier {
+.lxl-invalid .lxl-qualifier {
 	text-decoration: underline 1px solid;
 	text-decoration-color: rgba(255, 0, 0, 0.75);
 	text-decoration-style: wavy;
--- a/lxl-web/src/lib/styles/lxlquery.css
+++ b/lxl-web/src/lib/styles/lxlquery.css
@@ -1,5 +1,5 @@
 .lxl-qualifier {
-	display: inline-flex;
+	display: inline-block;
 	color: rgb(0, 128, 0);
 	background: rgba(14, 113, 128, 0.1);
 	padding-top: 3px;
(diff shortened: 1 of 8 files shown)


johanbissemattsson (approved):
LGTM! Awesome work! 🌟

PR #1201, opened 2025-01-16
feat(lxlweb): supersearch feature flag param (LWS-275)

jesperengstrom:
### Tickets involved
[LWS-275](https://kbse.atlassian.net/browse/LWS-275)

### Solves

Easiest possible solution to toggle supersearch using a param: we borrow the `_x` param and show it when set to `_x=supersearch`.

It's good 'cause `_x` is accepted and returned by the api, meaning the facet links etc will have it too.

Explanation for the outcommented code:
The api now accepts invalid property keys. Users of the 'old' search will get no feedback that this happened (only zero results). We manually thew an error for them ([here](https://github.com/libris/lxlviewer/pull/1195)). But it's no longer working when moving away from the `.env`. So instead of building another more complicated temp solution, I suggest users will have to live with this until supersearch is fully in place.

### Summary of changes

* change `useSuperSearch` variable
* small change to the search mapping 'edit' button
* comment out manual handling of _invalid type
* comment out test failing because of 👆

### DIFF
--- a/lxl-web/src/lib/components/Search.svelte
+++ b/lxl-web/src/lib/components/Search.svelte
@@ -19,7 +19,8 @@
 
 	let { placeholder, autofocus = false }: Props = $props();
 
-	const useSuperSearch = env?.PUBLIC_USE_SUPERSEARCH === 'true';
+	let useSuperSearch =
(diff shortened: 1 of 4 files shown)


johanbissemattsson (approved):
LGTM! 🙌 I'm fine with letting the beta users temporarily live without error messages for invalid property keys (it is a beta after all).

PR #1204, opened 2025-01-17
feat(lxl-web): Add debug mode for relevancy scoring

olovy:
Add debug mode for relevancy scoring.
* For each hit display total and per field score. And a percentage that relates the score to other hits _on the same page_.
    * (There is a backend bug that makes some field appear as e.g. "Synonym(hasTitle._str:hor hasTitle._str:")
* Enable (sticky) with query param `?_debug=esScore`, disable with `?_debug=false`
  * this is a bit messy and buggy...
* Click the score to show/hide detailed score calculation

bild

### Tickets involved
[LWS-296](https://kbse.atlassian.net/browse/LWS-296)

### DIFF
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
@@ -16,8 +16,10 @@ export const load = async ({ params, url, locals, fetch }) => {
 		redirect(303, `/`); // redirect to home page if no search params are given
 	}
 
+	const debug = locals.userSettings?.debug === true ? '&_debug=esScore' : '';
+
 	const searchParams = new URLSearchParams(url.searchParams.toString());
-	const recordsRes = await fetch(`${env.API_URL}/find.jsonld?${searchParams.toString()}`, {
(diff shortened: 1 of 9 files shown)


lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
```diff
@@ -16,8 +17,10 @@ export const load = async ({ params, url, locals, fetch }) => {
 		redirect(303, `/`); // redirect to home page if no search params are given
 	}
 
+	const debug = locals.userSettings?.debug.includes(DebugFlags.ES_SCORE) ? '&_debug=esScore' : '';
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts:20]: This is what's causing the test to fail.
Try changing it to
`const debug = locals.userSettings?.debug?.includes(DebugFlags.ES_SCORE) ? '&_debug=esScore' : '';`

Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts:20]: Because: the test in question is changing the facet sorting, and thereby creating a settings cookie in userSettings. So it has `locals.userSettings` but not `debug`.

Comment (olovy) [reply lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts]: Yes! That was it. Missed it when changing it from a boolean to an array.

PR #1205, opened 2025-01-17
feat(supersearch): Add persistent item support (LWS-289)

johanbissemattsson:
### Tickets involved
[LWS-289](https://kbse.atlassian.net/browse/LWS-289)

### Solves

Add persistent item support (shown before fetched result items).

### Summary of changes

- Add persistent item support
- Use default testid naming convention
- Rename id and class name of suggestion items
- Update readme


jesperengstrom (approved):
LGTM! 🌟

PR #1211, opened 2025-02-04
chore(idkbse): Remove unused fonts and related CSS

andersju:
We don't actually use Sentinel (a proprietary font) on id.kb.se so let's not keep it around.


olovy (approved):
:broom:

PR #1213, opened 2025-02-07
feat(supersearch): Improve keyboard navigation and refactor snippet props (LWS-297)

johanbissemattsson:
### Tickets involved
[LWS-297](https://kbse.atlassian.net/browse/LWS-297)

### Solves

Improves keyboard navigation (allowing navigation between items on the input row) and refactors snippet handling for better flexibility.

This one grew inadvertently (sorry for that!) – but will pave the way for [LWS-278](https://kbse.atlassian.net/browse/LWS-278).

### Summary of changes

- Add snippet props for `inputRow` and `resultItemRow`
- Use named snippet params instead of positional
- Add keyboard tabbing on input row
- Simplify tests (removing the need for two different SuperSearch instances)
- Only show persistent result items if query value has length
- Hide expanded search on navigation (in lxl-web)
- Add persistent row and update styling (in lxl-web)
- Update readme


jesperengstrom (approved):

johanbissemattsson:
> Wow, great work! ⭐ Two very small remarks
> 
>  * Search button has no visible focus style in dialog. If we just want to fix it for demo purposes, we could add:
> 
>  * Maybe we want to lose focus from the search input when navigating away, emulating a "hard load" of the page?

Added your change regarding focus styling on the search button in https://github.com/libris/lxlviewer/pull/1213/commits/421cb7780fe01f27360f25343bf9d0af2645c87e 😊

I'm not against removing the focus after navigations to resources but I think we should evaluate how it feels when adding qualifiers to a search query (my feeling is that we still want focus then?).

jesperengstrom:
> I'm not against removing the focus after navigations to resources but I think we should evaluate how it feels when adding qualifiers to a search query (my feeling is that we still want focus then?).

Yes, probably only when navigating away from /find

> Another thing that we should evaluate is if the search input should be cleared when navigating to resources, or if the entered query should stay put? (I'm currently leaning towards the latter).

Yeah, maybe. We should also take into consideration the role of supersearch in the context of "related stuff" on agent/concept pages. I've imagined today's two search inputs as one, with the "thing" pre-filled in.

PR #1216, opened 2025-02-11
Allow locking/unlocking of listed properties

lrosenstrom:
Make it possible to lock and unlock properties listed in settings for editing/removal.

Works on all levels in the JSON-tree.

If the user adds a protected property, it shows up in the form unlocked.  Currently, the only protected property is `usageAndAccessPolicy`. 

Locked:

Screenshot from 2025-02-11 14-14-12

Unlocked:

Screenshot from 2025-02-11 14-14-20

Modal:

Screenshot from 2025-02-11 14-14-31

### Tickets involved
[LXL-4178](https://kbse.atlassian.net/browse/LXL-4178)

### DIFF
--- a/cataloging/src/settings.js
+++ b/cataloging/src/settings.js
@@ -138,6 +138,9 @@ export default {
       },
     ],
   },
+  protectedProperties: [
+    'usageAndAccessPolicy',
+  ],
   warnOnSave: {
     'record.encodingLevel': ['marc:PrepublicationLevel', 'marc:PartialPreliminaryLevel'],
   },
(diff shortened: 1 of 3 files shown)


cataloging/src/settings.js
```diff
@@ -138,6 +138,9 @@ export default {
       },
     ],
   },
+  protectedProperties: [
```
niklasl, line 141: This could maybe also include chains `["publication", "date"]`, and (step 2) be a calculated list from category-tagged properties/chains in vocab?

lrosenstrom: I will look into the chain/property path part right away. Seems reasonable!

lrosenstrom:
It should now be possible to specify protected properties using vue-client-style paths such as:
`mainEntity.instanceOf.usageAndAccessPolicy`
`mainEntity.publication.date`
etc. 

Specification still happens in  `settings.js`, for now.

Just supplying the name of the property makes it protected throughout the form.

lrosenstrom:
> Nice! LGTM
> 
> Rework the `nextShelfControlNumber` lock to use this instead? Would need to make it possible to override the message key.

Yes! I'll give it a go.

olovy:
> > Nice! LGTM
> > Rework the `nextShelfControlNumber` lock to use this instead? Would need to make it possible to override the message key.
> 
> Yes! I'll give it a go.

We can merge this first.

PR #1217, opened 2025-02-12
fix(lxlweb): familyName givenName swap (LWS-270)

jesperengstrom:
### Tickets involved
[LWS-270](https://kbse.atlassian.net/browse/LWS-270)

### Solves

Should we do it??

### Summary of changes

* swap `familyName` `givenName` place, remove comma before properties, add 'no separator'.


olovy (approved):
`¯\_(ツ)_/¯`

If that is what has been decided. Trivial to revert.

PR #1218, opened 2025-02-12
fix(lxlweb): iterate parents in getEditedRanges

jesperengstrom:
### Solves

Discovered a flaw in `getEditedRanges` that only recognizes editing a qualifier when the `innerNode` is a direct child of `qualifierValue`. This is however not always the case. Consider for example:

`contributor:(astrid lindgren)` that produces the tree `QualifierValue { Group { String String } }`. We're editing the `String` and the parent is `Group`.

This means we fail to narrow down the query to `_q=(astrid+lindgren)+"rdf:type":Agent` in this case.

And since the query in theory could be `contributor:(((((Astrid)))))`, we need to iterate the tree to se if the `innerNode` is located _somewhere_ inside a `qualifierValue`.   

### Summary of changes

* Iterate parents in `getEditedRanges`
* Add test to cover this use case

### DIFF
--- a/lxl-web/src/routes/api/[[lang=lang]]/supersearch/getEditedRanges.test.ts
+++ b/lxl-web/src/routes/api/[[lang=lang]]/supersearch/getEditedRanges.test.ts
@@ -40,6 +40,31 @@ describe('getEditedRanges', () => {
 		);
 	});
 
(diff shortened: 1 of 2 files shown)


johanbissemattsson (approved):
Good find! LGTM! 🚢

PR #1219, opened 2025-02-12
feat(supersearch, lxl-web): Add start content inside supersearch (LWS-278)

johanbissemattsson:
### Tickets involved

### Solves

Adds start content in the supersearch component which is shown before the user has entered a value.

This is used in lxl-web to show examples of qualifier keys which can be used to build search queries. When clicking on one of these the qualifier key is added in the supersearch input (e.g. `contributor:""`). The cursor is also moved (to inside the added quotation marks) so it's easy to continue building the search query.

### Summary of changes

- Add `startContent` snippet
- Add support for adding qualifier keys by dispatching changes
- Implement `startContent` in lxl-web
- Update readme
- Add tests
- Rename `persistentItemRow `to `persistentResultItemRow` for better clarity


jesperengstrom (approved):
Great work, LGTM! ⭐ 
This in combination with https://github.com/libris/lxlviewer/pull/1220 will most certainly work great.

I would like an option to bring up the "start content" at other times (when the query contains stuff), but that can be saved for later.

johanbissemattsson:
> Great work, LGTM! ⭐ This in combination with #1220 will most certainly work great.
> 
> I would like an option to bring up the "start content" at other times (when the query contains stuff), but that can be saved for later.

Yes! I think that will be the natural next step – so that it is also visible when the user first expands the search _after_ adding a qualifier.

PR #1220, opened 2025-02-13
fix(lxlweb): implement mapping _key _value (LWS-327)

jesperengstrom:
### Tickets involved
[LWS-327](https://kbse.atlassian.net/browse/LWS-327)

### Solves

Frontend implementation of https://github.com/libris/librisxl/pull/1553 that lets us match a qualifier key/value with its corresponding mapping object, label & removelink.

Try for example...
To type any key that the backend understands, for example `ämne:`
To add multiple 'contributor' filters from the facets (no more mixed up labels)

### Summary of changes

* Implement search.mapping `_key` & `_value`
* Update test
* Fix bug in supersearch input field where many pills pushed away the search button 🙃

### DIFF
--- a/lxl-web/src/lib/components/SuperSearchWrapper.svelte
+++ b/lxl-web/src/lib/components/SuperSearchWrapper.svelte
@@ -123,7 +123,7 @@
 						<BiArrowLeft />
 					</button>
 				{/if}
-				<div class="flex-1">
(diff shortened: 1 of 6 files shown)


johanbissemattsson (approved):
Awesome! LGTM! 🙌

PR #1226, opened 2025-02-21
feat(lxl-web): Add option to sort (works) on number of holdings

olovy:
TODO
Doesn't take the type into account. So it is also displayed when searching for e.g. Person
But that is also the case for sort.publicationDesc/sort.publicationAsc

Depends on https://github.com/libris/librisxl/pull/1560

### Tickets involved
[LWS-169](https://kbse.atlassian.net/browse/LWS-169)


johanbissemattsson (approved):
Looks good code-wise (but it breaks for me due to the dependency on https://github.com/libris/librisxl/pull/1560)

PR #1228, opened 2025-02-24
Feature/lxl 4625 blank node id issue

lrosenstrom:
As described in [LXL-4625](https://kbse.atlassian.net/browse/LXL-4625) there are rare cases where the hidden ID's that are generated by the frontend differ in the `matchForm` and `targetForm` parts of the bulk change specification. This sometimes occured when saving a bulk change with the  `matchForm` part active. It was caused by unwanted sideeffects of calling `onInactiveMatchForm` (this method should only be called when actually switching between the forms).


olovy (approved):
:ship:

PR #1231, opened 2025-02-25
feat(supersearch, lxl-web): Better control over start content visibility (LWS-278)

johanbissemattsson:
### Tickets involved

### Solves

Adds better control over start content (introduced in https://github.com/libris/lxlviewer/pull/1219) visibility – as just checking for value length is not enough. The new `shouldShowStartContentFn` function prop can be used for more granular control.

In the case of `lxl-web` the start content is now shown when the cursor as long as the cursor isn't placed inside a qualifier or edited string part (so we can now show the start content while still having previously added qualifiers).

An example:
1. Add a qualifier (e.g. `contributor:"Astrid Lindgren"`)
2. The expanded search should now show the start content if the cursor is placed outside the previously added qualifier (this was previously not possible).

The lxl-web implementation relies on a parsed syntax tree (with the help of `codemirror-lang-lxlquery`). If performance becomes an issue we should investigate if we can reuse the tree created for the syntax highlighting, but I kept it separate for now for the sake of simplicity.

### Summary of changes

- Add function controlling if start content should be shown


olovy (approved):

jesperengstrom:
Should the `persistenItemRow` sometimes be shown in combination with the start content (i.e if input is not empty)?
consider:
Skärmavbild 2025-02-27 kl  11 28 23

Pressing enter now performs search...

Skärmavbild 2025-02-27 kl  11 29 30

Adding an space now completely changes the dialog and the focused row, pressing enter now adds a contributor. That took me by surprise, very much used to press enter -> search, unless I actively chose _another_ alternative in the drop-down...

PR #1234, opened 2025-02-27
feat(cataloging): Display backend validation errors

olovy:
Quick and dirty display of JSON-LD validation errors from backend. 
These might happen on old records that are broken or when creating a new record from a user uploaded file.

When we get a HTTP 400 because of invalid JSON-LD inside the record we already also get a JSON structure describing the individual errors. 

Example
```
{
  "message":"Invalid JSON-LD",
  "status_code":400,
  "status":"Bad Request",
  "errors":[
    {
      "type":"INVALID_IRI",
      "description":"Invalid IRI",
      "path":["@graph",0,"descriptionLanguage","@id"],
      "key":"@id",
      "value":"122434 https://id.kb.se/language/swe"
    },
    {
      "type":"OBJECT_TYPE_EXPECTED",
      "description":"Expected value type of key to be object",
      "path":["@graph",1,"closeMatch",0,"inScheme","language"],
      "key":"language",
      "value":"swe"
    }
  ]
}
```
Display these inside the form.

Also display a more specific error toast that the generic "Something went wrong".

Room for improvement
* For now, displays the error description as a tooltip.
* For now, it just expands the whole form instead of zooming in on the actual bad properties.
* Not sure if it properly display all kinds of errors

Added Swedish translations for error messages. They are somewhat reworded to hopefully be a bit more helpful for users. (We should probably use the error code/type (e.g. VOCAB_STRING_EXPECTED) that is also present in the error object as key for i18n but it doesn't fit with the way i18n.json is set up right now.)

Skärmbild från 2025-02-27 12-43-56
Skärmbild från 2025-02-27 12-47-31
Skärmbild från 2025-02-27 12-44-11

There is already some old code for displaying client side validation errors when in "felsökningsläge". 
Decided to do this completely separate from that,

### Solves
The user cannot save an existing record but doesn't know why.


lrosenstrom (approved):
This is very useful! :ship:

PR #1236, opened 2025-02-27
fix(supersearch): Prevent setting default row on start content (LWS-278)

johanbissemattsson:
### Tickets involved

### Solves

Prevent setting default row on start content

### Summary of changes

- Prevent setting default row on start content


jesperengstrom (approved):
Nice! 🚢

PR #1237, opened 2025-02-27
feat(supersearch): Add arrow key cursor handling (LWS-330)

johanbissemattsson:
### Tickets involved
[LWS-330](https://kbse.atlassian.net/browse/LWS-330)

### Solves

Adds arrow key cursor handling.

Arrow up and down to jump to the start and end of the string is prevented if the expanded search is open.
Arrow left and right cursor movement is prevented if there are more than one cell/column on the active row.

### Summary of changes

-  Add arrow key cursor handling


olovy (approved):
LGTM! :ship:

PR #1238, opened 2025-02-28
fix(lxlweb): set _stats to false

jesperengstrom:
### Solves

Frontend implementation of https://github.com/libris/librisxl/pull/1566

### Summary of changes

* Set `_stats=false` for api/supersearch requests
* omit `facetGroups` and `predicates` from response if stats is missing
* modify `SearchResult` type and component accordingly


johanbissemattsson (approved):
LGTM!

PR #1240, opened 2025-03-04
fix(lxlweb, supersearch): pill remove button instead of link (LWS-331)

jesperengstrom:
### Tickets involved
[LWS-331](https://kbse.atlassian.net/browse/LWS-331)

### Solves

Up until now we've used the api:s `@up` link to remove qualifier pills. It's good because we get them for free and get preloading of the result.

But it's really bad because links are only accurate after page load and are unaware anything inserted after that.
We could continue using links if we re-generate them on every input and (re-render the pill component) but that seems bad for performance.

Instead proposing to turn the link into a button + pass a function that generates a link, dispatches and navigates in the wrapper component.

### Summary of changes

* remove `removeLink` stuff
* add `removeQualifierFn` stuff


johanbissemattsson (approved):
LGTM! 🙌

PR #1243, opened 2025-03-05
fix(lxlweb): Hide lone wildcard from supersearch input

jesperengstrom:
## Description

Hide the fallback wildcard that appears when removing all filter pills.
It's especially annoying because of all the weird suggestions it brings up...

Skärmavbild 2025-03-05 kl  13 08 01

### Summary of changes

* Don't dispatch '*' when removing qualifier
* Don't set `q` to '*' when syncing after route change


johanbissemattsson (approved):
🙌 🚢

PR #1244, opened 2025-03-06
feat(lxl-web): Redesign and swap primary action on suggestions

johanbissemattsson:
### Tickets involved
[LWS-337](https://kbse.atlassian.net/browse/LWS-337)

### Solves

Redesign of the supersearch suggestions and swaps the primary action on qualifier suggestions. The primary action is now to add the qualifier to the search query instead of navigating to the qualifiers resource page.

The link to the resource page is instead placed behind a more button (represented by the icon with three dots)  together with a selector for adding different qualifier types (e.g. if the qualifier can be added as both a contributor or a subject).

### Summary of changes

- Redesign suggestions 
- Add drop down menu
- Move related components to supersearch components folder
- Remove peristent row
- Add nowrap to locale selector
- Add more utility classes


lxl-web/src/lib/components/supersearch/Suggestion.svelte
```diff
@@ -0,0 +1,296 @@
...
+	.action-type {
+		font-size: theme(fontSize.sm);
+		margin-left: auto;
+		padding-left: theme(padding.2);
+		color: theme(textColor.tertiary);
+		order: 1;
+		white-space: nowrap;
+	}
+
+	.badge {
```
jesperengstrom, line 209: If we're sticking to tailwind, I feel like this could be expressed as
`<span class="bg-positive rounded-full px-0.5 py-2 text-xs ml-1">`

To avoid creating the intermediary class? This applies to most css in this component...

johanbissemattsson: I'm ok with changing it to utility-only classes (even if I still often feel that I have to create intermediary classes while developing to help reason what the different elements are meant to do).

All design tokens in Tailwind 4.0 are exposed as native CSS variables so:
```
	.badge {
		background: theme(backgroundColor.positive);
		border-radius: theme(borderRadius.full);
		padding: theme(padding[0.5]) theme(padding.2);
		font-size: theme(fontSize.xs);
		margin-left: theme(margin.1);
	}
```
would become something similar to:
```
	.badge {
		background: var(--color-positive);
		border-radius: var(--radius-full);
		padding: var(--spacing-1) var(--spacing-2);
		font-size: var(--text-xs)
		margin-left: var(--spacing-1);
	}

```

I think my main gripe is the usage of `@apply` throughout the project (when it only becomes an added abstraction  – a different way to write CSS). 

There also seems to be some issues with `@apply` in [Tailwind 4.0](https://github.com/tailwindlabs/tailwindcss/discussions/16429).

jesperengstrom:
The new 'Go to' text cuts of contributors for long titles/small screens. Compare:
develop
Skärmavbild 2025-03-10 kl  11 52 43

And this branch
Skärmavbild 2025-03-10 kl  11 52 06

I'm not sure his text is really needed? 
* One would assume these are result suggestions that you could navigate to (do we have to point it out)?
* The placement suggest 'Go to' is some kind of separate action that you can choose?

johanbissemattsson:
> The new 'Go to' text cuts of contributors for long titles/small screens.
>
> I'm not sure his text is really needed?
> 
>     * One would assume these are result suggestions that you could navigate to (do we have to point it out)?
> 
>     * The placement suggest 'Go to' is some kind of separate action that you can choose?

I have removed it for now https://github.com/libris/lxlviewer/pull/1244/commits/261640336494d471c67377ccd3fe6ddc882256af.

There are some pros and cons with having the label. I think the main advantage of keeping it is with screen readers – as it tells what the action actually does (especially when it is in the form of a button instead of a link).
It's hierarchical position is placed before the resource title element (even if it is visually placed after) so it would be read as _Go to Den owäldiga granskaren..._ instead of _Den owäldiga granskaren... go to_

jesperengstrom:
> There are some pros and cons with having the label. I think the main advantage of keeping it is with screen readers – as it tells what the action actually does (especially when it is in the form of a button instead of a link). It's hierarchical position is placed before the resource title element (even if it is visually placed after) so it would be read as _"Go to Den owäldiga granskaren..."_ instead of _"Den owäldiga granskaren... go to"_

Can't the label then be `.sr-only`?

lxl-web/src/lib/components/supersearch/Suggestion.svelte
jesperengstrom: I'm all for moving away from `@apply`. Just that creating intermediary classes seems like a way to 'not have to use tailwind while still using it'. I mean:
* We might as well then replace the `theme` a css variables file
* We'll ship twice as much css if we use both the utility classes directly and a bunch of own classes that also uses them?

johanbissemattsson: > I'm all for moving away from `@apply`. Just that creating intermediary classes seems like a way to 'not have to use tailwind while still using it'. I mean:
> 
>     * We might as well then replace the `theme` a css variables file
> 
>     * We'll ship twice as much css if we use both the utility classes directly and a bunch of own classes that also uses them?

* The `tailwind.config.js` file is already replaced with a CSS file in v4.0.
* The utility classes are only added when they are used?

johanbissemattsson:
> > There are some pros and cons with having the label. I think the main advantage of keeping it is with screen readers – as it tells what the action actually does (especially when it is in the form of a button instead of a link). It's hierarchical position is placed before the resource title element (even if it is visually placed after) so it would be read as _"Go to Den owäldiga granskaren..."_ instead of _"Den owäldiga granskaren... go to"_
> 
> Can't the label then be `.sr-only`?

True! Fixed in [6062b6a](https://github.com/libris/lxlviewer/pull/1244/commits/6062b6a825c8e3079b339b5eaab7da5a064c43aa)

jesperengstrom (approved):
Nice work! 🌟 🚢

PR #1250, opened 2025-03-17
fix(lxl-web): Unquote qualifier values (LWS-346)

johanbissemattsson:
### Tickets involved
[LWS-346](https://kbse.atlassian.net/browse/LWS-346)

### Solves
Adds a work-around so results for e.g. `contributor:"Astrid Lindgren"` are shown.

This is done by unquoting the qualifier value so it is treated as _Astrid AND Lindgren_ instead of the exact phrase _"Astrid Lindgren"_ (the latter doesn't give any results as the search index uses _last name + first name_ instead of _first name + last name_).

This should only be seen as a temporary work-around as it's probably better to replace the automatically inserted quotes with (hidden?) parenthesis.

### Summary of changes

- Unquote qualifier values


kwahlin (approved):
LGTM

jesperengstrom (approved):
LGTM!

PR #1251, opened 2025-03-17
chore(lxlweb, supersearch, lxlquery): bump dependencies, fix warnings

jesperengstrom:
### Solves

Update almost all dependencies of lxlweb, supersearch and lxlquery to latest version. See diff for complete list.
`npm audit` gone from 11 vulnerabilities to 4 minor ones.

Major version bumps:
Vite 5 -> 6
Vitest 2 -> 3

Did _not_ upgrade:
`lxljs` dependencies 😱 
tailwind 3 -> 4 (saved for later)

`eslint-plugin-svelte` has some new rules that we'll bumb into immediately, like [each without key](https://sveltejs.github.io/eslint-plugin-svelte/rules/require-each-key/)

Tried to fix all vite warnings that happen on run build and dev (see below) - the only one that remains is 
```Module "crypto" has been externalized for browser compatibility...```
From the fact that we use a node-only package. But only server side...

### Summary of changes

* Update packages
* Add unused selector `.image` to an element (build warning)
* Refactor` i18n/index.ts` to avoid both static and dynamic import of translation files (build warning)
* Move cookie consent translations to own folder for same reason ☝️ (build warning)
* Move non-scope:able styles from cookie consent component (dev warning)
* Fix type imports for codemirror modules causing "unused" warnings (build warning)


johanbissemattsson (approved):
AWESOME! 🌟 🌟 🌟 

So nice not having to scroll through loads of warnings when running a simple `npm run dev` command. Thanks!

PR #1253, opened 2025-03-19
feat(supersearch): Add select event handler 

johanbissemattsson:
### Solves

Replaces bindable cursor prop with a selection event handler

### Summary of changes

- Add select event handler
- Use selection head instead of anchor for cursor
- Replace bindable cursor prop
- Use selection inside handleShouldShowStartContent
- Skip redundant selection sync
- Hide result items when text is selected
- Look both left and right when deciding if editing a part of query


jesperengstrom:
Seems to work like before (other PR). LGTM! 👍

jesperengstrom (approved):

PR #1254, opened 2025-03-20
Feature/lws 304 fav libraries

lrosenstrom:
Make it possible to search for libraries and add them to favorites under My pages. 

Currently, we store favorite libraries in cookies. Depending on how this works out when accessing the favorites/user settings we might want to change how to share this state, see e.g. https://svelte.dev/docs/svelte/stores for an introduction. 

The suggestion is to store libraries on the form (see [types/search.ts](https://github.com/libris/lxlviewer/compare/feature/lws-304-fav-libraries?expand=1#diff-b6a885e74312985608a7652df85f304f2ae3d37cdfcea72cfdda719e934c7243)):
 
```
export interface LibraryItem {
	[JsonLd.ID]: string;
	label: string;
	sigel: string;
}
```

Screenshot from 2025-03-20 15-23-30

**TODO:**
* Refactor `userSettings.ts` to be more general/flexible. Depends on solution for stores/sharing state.
* ~Translations~
* Temporarily remove link to My pages before the 'My Libraries' feature is complete, including usage throughout the app?


lxl-web/src/lib/components/my-pages/Libraries.svelte
```diff
@@ -0,0 +1,129 @@
+<script lang="ts">
+	import { page } from '$app/stores';
```
jesperengstrom, line 2: `$app/stores` is deprecated in the version of svelteKit we use now (I know it's all over the old components), so when creating new stuff we should use `$app/state` instead and then just change `$page` to `page`...
https://svelte.dev/docs/kit/migrating-to-sveltekit-2#SvelteKit-2.12:-$app-stores-deprecated

lxl-web/src/routes/(app)/[[lang=lang]]/HeaderMenu.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import { page } from '$app/stores';
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/HeaderMenu.svelte:2]: (see other comment about page/state)

lxl-web/src/lib/utils/search.ts
```diff
@@ -31,6 +32,22 @@ import { getUriSlug } from '$lib/utils/http';
 // import { error } from '@sveltejs/kit';
 // import { env } from '$env/dynamic/public';
 
+export async function asLibraryResult(
```
jesperengstrom, line 35: I kind of think `asLibraryResult` should be a local function of the api/my-pages/server file... I can't really see it being reused in some other case, so the export/import thing is kind of unnecessary?

jesperengstrom (approved):
Nice! 🚢

PR #1255, opened 2025-03-21
fix(idkbse): encode single quote char in requests to backend

andersju:
Terms with a `'` in their prefLabel 404, for example:

* https://id.kb.se/term/gmgpc/swe/%C3%89preuves%20d'artiste
* https://id.kb.se/term/saogf/Rock'n'roll

Looking with tcpflow, the idkbse app sends a request to the backend for `https://id.kb.se/term/saogf/Rock'n'roll` rather than (the correct) `https://id.kb.se/term/saogf/Rock%27n%27roll`. So, this should fix it.


olovy (approved):
:ship:

PR #1256, opened 2025-03-24
feat(lxlweb): Fav libraries filter (LWS-298) 

jesperengstrom:
### Tickets involved
[LWS-298](https://kbse.atlassian.net/browse/LWS-298)

### Solves

This adds:
1) A user settings management util that lets you do stuff like `userSettings.addLibrary(lib)` anywhere in the app.
2) A `MyLibrariesFilter` component for the facet panel.

Some notes:

1) We want the user settings to be a reactive state (to dynamically update the libraries list when adding etc), but absolutely avoid sharing that state on the server. But at the same time render the user settings during SSR (to avoid flashing etc). The best solution i've come up with is to turn cookie into page.data and use if `!browser`, if browser - set and use `$state`.

2) The filter creates a partial _q string from the fav libraries and creates a link appending/removing it from the query. It also looks for that exact string to decide if the filter is active. This is pretty dumb and we could put in additional work to make it smarter. But my guess is we'll end up falling short without proper language parsing. Consider for example:

`fav libraries: ['S', 'G']`

`_q=(itemHeldBy:"sigel:S" OR itemHeldBy:"sigel:G")` = match, "filter is active"
`_q=NOT (itemHeldBy:"sigel:S" OR itemHeldBy:"sigel:G")` = also match! (but has exact opposite meaning)

We will soon encounter this problem again when creating the 'My libraries' pill for supersearch. After discussion, the plan now is:
* To keep this working (but flawed) solution for now
* To create a `myLibraries` filter alias that filters the libraries under the hood in the backend, solving most of these issues. 

TODO: Remove mock code

### Summary of changes

* Create `layout.server.ts` file that adds user settings to `page.data` globally (no need to pass it in every load function)
* Create centralized settings state handling in `userSettings.svelte.ts`
* Init user settings client state in `layout.ts`
* Migrate `layout.svelte` to v5 sytax.
* Create `MyLibrariesFilter` component
* Implement `userSettings` state in `FacetGroup` component + refactor to v5 syntax


jesperengstrom:
Closing in favor of https://github.com/libris/lxlviewer/pull/1265

PR #1257, opened 2025-03-25
fix(lxlweb): catch parsing error, fix failing test

jesperengstrom:
### Solves

e2e-test failing now because we expect a 404-page at https://beta.libris-qa.kb.se/aaaaaaaaaaaaaaa, but for some reason getting a [502 html page](https://libris-qa.kb.se/aaaaaaaaaaaaaaa?framed=true) that we fail to parse as json, resulting in a 500 internal error page...

Changing the cursed /aaaaaaaaaaaaaaa route in the test, because any other non-existent fnurgel seems to result in 404 as expected.

### Summary of changes

* Change 404 route in test
* try/catch the error parsing to be able to display the actual error
* Replace deprecated app/stores `$page` in Error component

### DIFF
--- a/lxl-web/src/lib/components/Error.svelte
+++ b/lxl-web/src/lib/components/Error.svelte
@@ -1,14 +1,14 @@
 <script>
-	import { page } from '$app/stores';
+	import { page } from '$app/state';
 	import getPageTitle from '$lib/utils/getPageTitle';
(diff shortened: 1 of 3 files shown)


johanbissemattsson (approved):
LGTM!

PR #1258, opened 2025-03-25
feat(lxl-web): Add category property to display-web

johanbissemattsson:
### Solves

Adds category property to display-web.json

### Summary of changes

- Adds category property to display-web

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -24,7 +24,9 @@
 						"hasNotation",
 						"hasVariant",
 						"inCollection",
-						"genreForm",
+						{
+							"alternateProperties": ["category", "genreForm"]
+						},
 						"subject",
 						"classification",
 						"intendedAudience",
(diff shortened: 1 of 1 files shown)


niklasl (commented):
I don't think `alternateProperties` is necessary.

Note that `category` is experimental at this point. Even if it replaces `genreForm`, we want to see if something is off (and if it does replace it, `genreForm` will be gone; so there is probably no case where `alternateProperties` would be needed for these two?

johanbissemattsson:
> I don't think `alternateProperties` is necessary.
> 
> Note that `category` is experimental at this point. Even if it replaces `genreForm`, we want to see if something is off (and if it does replace it, `genreForm` will be gone; so there is probably no case where `alternateProperties` would be needed for these two?

👍  `alternateProperties` is now removed in https://github.com/libris/lxlviewer/pull/1258/commits/82d2b79653a8f0e0a7beebb2a0f62aeb915c1be5 .

My intention adding it was so `category` can easily be tested depending on the used enviroment (e.g. if category exists on the DEV but not QA) but if it still is very experimental then it's probably better to wait.

PR #1268, opened 2025-04-03
feat(lxlweb): MyLibraries holding indicator (LWS-300)

jesperengstrom:
### Tickets involved
[LWS-300](https://kbse.atlassian.net/browse/LWS-300)

### Solves

Displays an icon indicating that an item is held by one of user's favourite libraries. When hovered, a tooltip displays the library names. The indicator icon can appear in the search result card, on product pages instances list or, if a single instance, next to the libraries button.

All these places uses a shared util function, `getMyLibsFromHoldings`, that accepts one or many instances in the `holdingsByInstanceId` format and returns an intersected list of `LibraryItem`s.

### Summary of changes

* Create `MyLibsHoldingIndicator` component and add to `SearchCard.svelte`, `InstancesList.svelte`, fnurgel `page.svelte`.
* Create `getMyLibsFromHoldings` util function in holdings.ts
* Add `heldByMyLibraries` property to all `SearchResultItem`s if user has fav libraries.
* Convert `Filters.svelte` to v5 syntax and move `shouldShowMapping` function to Filters component (only used there).
* Convert `SiteHeader.svelte` to v5 syntax and close menu if navigating to another path client side.

### DIFF
--- a/lxl-web/src/lib/components/find/SearchCard.svelte
+++ b/lxl-web/src/lib/components/find/SearchCard.svelte
@@ -166,9 +166,9 @@
 				{/if}
 			{/key}
 		{/if}
-		{#if item.atMyLibraries?.length}
+		{#if item.heldByMyLibraries?.length}
 			<div class="card-libraries flex items-start">
-				<MyLibsHoldingIndicator libraries={item.atMyLibraries} />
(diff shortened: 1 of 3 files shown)


johanbissemattsson (approved):
LGTM! 🌟I like that it's so visually clear which items are held by a favourite library 👍

PR #1273, opened 2025-04-10
feat(lxlweb, supersearch, lxlquery): Handle filter alias in supersearch

jesperengstrom:
### Tickets involved
[LWS-347](https://kbse.atlassian.net/browse/LWS-347), [LWS-313](https://kbse.atlassian.net/browse/LWS-313)

### Solves

Handle filter aliases in supersearch like so:

Skärmavbild 2025-04-10 kl  09 51 07

When myLibs is available as a filter alias, we should be able to just add it to the grammar file.

### Summary of changes

* `Search.ts`: return `displayStr` and `_value` properties for `mapping.object`s
* `syntax.grammar`: define `filterAlias` as a type of `QualifierValue`. Update tests.
* `getLabelsFromMappings`:  can now return only a value label.
*  `lxlQualifierPlugin`: now creates an atomic decoration from just the value label.
* `QualifierComponent`: Enable rendering a value label only.

### DIFF
--- a/lxl-web/src/lib/styles/lxlquery.css
+++ b/lxl-web/src/lib/styles/lxlquery.css
@@ -11,6 +11,10 @@
 	border-bottom-left-radius: 5px;
 }
 
+.lxl-filter-alias {
+	padding-right: 5px;
+}
+
 .lxl-qualifier-remove,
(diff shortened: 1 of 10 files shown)


johanbissemattsson (approved):
LGTM 🙌

PR #1274, opened 2025-04-11
chore(lxlweb): upgrade Tailwind to v4

jesperengstrom:
### Tickets involved
[LWS-351](https://kbse.atlassian.net/browse/LWS-351)
### Solves

* Migrate Tailwind to v4 using the migration tool + manual fixes.
* Replace PostCSS plugin with new Vite plugin meaning we can remove it as a dependency.
* Migrate tailwind.config, colors.css to app.css

**Notes on the result of the migration:**
* The Tailwind 4 theme lives in `app.css`, but our current theme in `tailwind.config.js` could not be automatically migrated. ~~I didn't find it meaningful to do it manually (before a complete rework), so it's now just imported into the CSS as a temp solution.~~
* ~~Some styles (i.e `fontSize` that also included `line-height`) I did not manage to fully migrate. You can see that the vertical spacing is slightly off compared to before.~~
* Due to the nature of TW4 (css, not js) tailwind is not magically available in component `<style>` tags without manual `@reference` of the theme css. So in order to make everything not broken, adding `@reference` became the bulk of the work. [This should of course be undone again as soon as possible](https://tailwindcss.com/docs/compatibility#vue-svelte-and-astro).
* As a result of the above, it feels like the app took a **major** performance hit when developing. In TW4, every style tag that uses tw syntax is treated as a CSS module that tailwind needs to watch over. Now we have like 20+ of those, and it's not really sustainable.
* Did not get the vsCode extension to work autocompleting the inline tw classes after upgrade. I think we _really_ need this...
--
* In summary: seems like a bad idea to continue developing "as usual" after merging this, we should then really only focus on setting up the theme for the new design. Maybe even in this branch?
* Tailwind 4 seems great - if we can get it to run smoothly.

**Subsequent steps:**
* ~~We should scrap/rewrite everything in `tailwind.config.js` into `app.css`~~
* We should ban most uses of tailwind syntax in component `<style>` tags (a.k.a `@apply`) to see if this improves performance. I.e:
  - Use inline styling where possible (with new arbitrary selectors, that should really be _everywhere_ - [example](https://github.com/libris/lxlviewer/pull/1274/files#diff-0b6ae1b2308b779c1febe12aae9a7309141b1123d7b621a1a0b26352bca2c9e0R18))
  - Use style tags with plain CSS (can still use our css vars)
  - Create custom `@components` for complex and reusable classes, custom `@utility` for special needs.
* ~~We should (very soon) make sure we can get intellisense working again, maybe by bootstrapping a brand new svelte5 + tailwind4 project and examine its config.~~

**TODO:**
~~Get tailwind extension to work~~
~~Remove `theme()` function as it is deprecated in v4~~ (see [docs](https://tailwindcss.com/docs/functions-and-directives))


jesperengstrom:
I migrated (and removed) `tailwind.config` and `colors.css` into app.css, to see if that solved our problems. 
Somehow I think the performance bit is much better now.
And we also have a _proper_ tailwind 4 setup. Feels much better as a starting point...

jesperengstrom:
After this last commit and restarting VSCode, the extension started working perfectly 🥳 
I think it was the hackish migration (importing tailwind.config into app.css) that it couldn't handle...

jesperengstrom:
Removed uses of deprecated `theme()` in `<style>` tags. Since css variables cannot be used in media queries, I used the `@variant` directive as a way to reuse our breakpoints in custom css. Looks kinda clean and does not seem discouraged... [(discussion)](https://github.com/tailwindlabs/tailwindcss/discussions/15113)

jesperengstrom:
Quite happy with this now, think it can be merged :)

lxl-web/src/lib/actions/dropDownMenu/index.svelte.ts
```diff
@@ -31,7 +31,7 @@ export const dropdownMenu: Action<HTMLElement, Parameter> = (
 		node.addEventListener('mouseover', attachDropDownMenu);
 		node.addEventListener('mouseout', removeDropDownMenu);
 		node.addEventListener('focus', attachDropDownMenu);
-		node.addEventListener('blur', removeDropDownMenu);
+		node.addEventListener('blur-xs', removeDropDownMenu);
```
johanbissemattsson, line 34: The blur events in this file should be kept as-is (as far as I know there isn't a blur-xs event). 😊

I guess the suspected culprit is the Tailwind upgrade tool which probably was a little too eager when updating class names.

lxl-web/src/lib/actions/popover/index.ts
```diff
@@ -51,7 +51,7 @@ export const popover: Action<HTMLElement, Parameter> = (
 	node.addEventListener('mouseover', attachPopover);
 	node.addEventListener('mouseout', removePopover);
 	node.addEventListener('focus', attachPopover);
-	node.addEventListener('blur', removePopover);
+	node.addEventListener('blur-xs', removePopover);
```
johanbissemattsson, line 54: Same issue as in `lxl-web/src/lib/actions/dropDownMenu/index.svelte.ts` – `blur` should be kept as-is in this file.

johanbissemattsson (changes requested):
Awesome work – both with the thorough write-up and the execution. 🌟 

See the comments about the `blur` events, otherwise I also think it is ready to be merged!

lxl-web/src/lib/actions/dropDownMenu/index.svelte.ts
jesperengstrom: 🤦

PR #1283, opened 2025-04-24
feat(lxlweb, lxlquery): Implement alias-myLibraries in frontend (LWS-313)

jesperengstrom:
### Tickets involved
[LWS-313](https://kbse.atlassian.net/browse/LWS-313)

### Solves

Frontend implementation of filter alias API feature (https://github.com/libris/librisxl/pull/1581), also a continuation of https://github.com/libris/lxlviewer/pull/1273.

1. If `alias-myLibraries` is present in `_q`, append myLibraries cookie values to secret param before fetch to search api: `_alias-myLibraries=(itemHeldBy:"sigel:X" OR itemHeldBy:"sigel:Y")`.

2. The response then includes the alias in search mapping, but has no label since it is a custom filter that's not hardcoded into the backend. Allow `displayStr` to look for the value label in the frontend translation files.

3. Add `alias-myLibraries` as a reserved qualifier term in lxlquery grammar. It will then pick up the label from search mapping:
Skärmavbild 2025-04-24 kl  10 14 03

Skärmavbild 2025-04-24 kl  10 14 44

4. Modify `MyLibrariesFilter` to recieve the facets as prop, and use the filter alias's `checked` prop and `view['@id']` as remove link.

### NOTES
* the MyLibraries filter is now currently disabled for the fnurgel pages related search. Implementing it there will be more challenging - and since the page will be redesigned, this seemed unnecessary. Do you agree?

* In the case that `alias-myLibraries` is in `_q`, _but libraries cookie is missing or empty_ - we still append `_alias-myLibraries=""` or the filter alias won't "activate" and is treated as free text search, right @lrosenstrom ?

### DIFF
--- a/lxl-web/src/lib/components/find/Filters.svelte
+++ b/lxl-web/src/lib/components/find/Filters.svelte
@@ -46,7 +46,9 @@
 				type="search"
 			/>
 			<BiSearch class="text-icon absolute top-3 left-2.5 text-sm" />
-			<MyLibrariesFilter />
+			{#if page.url.pathname === '/find'}
+				<MyLibrariesFilter {facets} />
+			{/if}
 			<ol>
 				{#each facets as group (group.dimension)}
 					<FacetGroup {group} locale={page.data.locale} {searchPhrase} />
--- a/lxl-web/src/lib/components/find/Filters.svelte
+++ b/lxl-web/src/lib/components/find/Filters.svelte
@@ -46,7 +46,7 @@
 				type="search"
 			/>
 			<BiSearch class="text-icon absolute top-3 left-2.5 text-sm" />
-			{#if page.url.pathname === '/find'}
+			{#if page.route.id === '/(app)/[[lang=lang]]/find'}
 				<MyLibrariesFilter {facets} />
 			{/if}
 			<ol>
--- a/lxl-web/src/lib/components/find/MyLibrariesFilter.svelte
+++ b/lxl-web/src/lib/components/find/MyLibrariesFilter.svelte
@@ -1,40 +1,34 @@
 <script lang="ts">
 	import { page } from '$app/state';
+	import type { FacetGroup, MultiSelectFacet } from '$lib/types/search';
 	import popover from '$lib/actions/popover';
+	import { MY_LIBRARIES_FILTER_ALIAS } from '$lib/constants/facets';
 	import { getUserSettings } from '$lib/contexts/userSettings';
 	import BiCheckSquareFill from '~icons/bi/check-square-fill';
 	import BiSquare from '~icons/bi/square';
 
-	const libraryValues = $derived(Object.values(getUserSettings()?.myLibraries || {}));
+	type MyLibrariesFilterProps = { facets: FacetGroup[] };
+	const { facets }: MyLibrariesFilterProps = $props();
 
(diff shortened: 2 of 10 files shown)


olovy (approved):
LGTM!

> the MyLibraries filter is now currently disabled for the fnurgel pages related search. Implementing it there will be more challenging - and since the page will be redesigned, this seemed unnecessary. Do you agree?

Agree

olovy (changes requested):
The "My libraries" facet disappears when switching language to English

lxl-web/src/lib/components/find/Filters.svelte
```diff
@@ -46,7 +46,9 @@
 				type="search"
 			/>
 			<BiSearch class="text-icon absolute top-3 left-2.5 text-sm" />
-			<MyLibrariesFilter />
+			{#if page.url.pathname === '/find'}
```
olovy, line 49: does not work for `/en/find`

jesperengstrom: Whoops! will fix

olovy:
I think this bug should not block merging but should be fixed later:

* Select "My libraries" filter
* Edit the query an use "My Libraries" inside a more complex expression
* Select "My libraries" filter again
* Error!

Just hide the filter if "My Libraries" i used in another part of the query? (can use the parsed query in mappings for this?)
Or add it to the query in a more robust way.

Skärmbild från 2025-04-24 11-18-44
Skärmbild från 2025-04-24 11-18-58

lrosenstrom:
>     * In the case that `alias-myLibraries` is in `_q`, _but libraries cookie is missing or empty_ - we still append `_alias-myLibraries=""` or the filter alias won't "activate" and is treated as free text search, right @lrosenstrom ?

Sounds good!

lxl-web/src/lib/components/find/Filters.svelte
jesperengstrom: fixed in https://github.com/libris/lxlviewer/pull/1283/commits/8a70b35f972ab5899f1f7f771a9fca2835a51bfd

olovy (approved):

lrosenstrom (approved):
LGTM!

jesperengstrom:
> * Select "My libraries" filter
> * Edit the query and use "My Libraries" inside a more complex expression
> * Select "My libraries" filter again
> * Error!

@olovy I've fixed this perticular use case in https://github.com/libris/lxlviewer/pull/1283/commits/f8e95a6c5773ee8e2e582d84e65bdcfd93c2d1e0 by not removing occurances of `alias-myLibraries` before appending it to the end. This is probably better.

The more general problem is that the frontend has to contruct the 'apply filter' link without access to language parsing. For other bool-filters, we can always rely on `view['@id']`, i.e they have an 'unchecked' state. This filter only exists when it is active.

On a side note - considering the query `(myLibraries OR boll)` - seems like the filter does have effect here. Shouldn't it be checked?

olovy:
> On a side note - considering the query `(myLibraries OR boll)` - seems like the filter does have effect here? Shouldn't it be checked?

I think that could make sense even though it is not only limited to my libraries in this case.
If unchecking it means navigating to the up-link that would always work?

PR #1284, opened 2025-04-29
Support new filter for free digital material

lrosenstrom:
### Tickets involved
[LWS-242](https://kbse.atlassian.net/browse/LWS-242)

### Solves
Frontend support for free e-resource filter defined in https://github.com/libris/definitions/pull/525. Other suggestions for a descriptive name of the filter are welcome.  

### Summary of changes

* Add `freeDigital` filter to grammar 

Screenshot from 2025-04-29 13-53-25

Screenshot from 2025-04-29 13-53-07

### DIFF
--- a/lxl-web/src/lib/utils/search.ts
+++ b/lxl-web/src/lib/utils/search.ts
@@ -116,7 +116,7 @@ export function displayMappings(
 						? translate(`facet.${m.alias}`)
 						: capitalize(m.property?.labelByLang?.[locale] || m.property?.label) ||
 							m.property?.[JsonLd.ID] ||
-							'No label', // lensandformat?
+							m._key,
 					operator,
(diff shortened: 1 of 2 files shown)


lrosenstrom:
Suggested change -> display the entered key when no label is present:
Screenshot from 2025-04-29 15-00-58
Screenshot from 2025-04-29 15-00-33

kwahlin (approved):
Nice!

lrosenstrom:
>
> I think the most correct way would be to take the labels from `search.mapping.property.propertyChainAxiom` but this is definitely better than `No label`!

Yes we could do that and get pills with something like `tillhörande media / benämning / ...` as labels. This is not the only way to present this, I guess, so maybe we save this for later?

PR #1289, opened 2025-05-06
feat(lxlweb): Redesign theme - adjust breakpoints

johanbissemattsson:
Follow-up PR to https://github.com/libris/lxlviewer/pull/1285.

### Solves

Adjusts custom breakpoints so they are more inline with the Tailwind defaults.

### Summary of changes

- Custom `--breakpoint-md` has been removed in favour of default `lg` which had the same value (64rem / 1024px)
- Custom `--breakpoint-lg` has become `--breakpoint-2xl` so we can keep the default lg value.
- Custom `--breakpoint-xl` has become newly added `--breakpoint-3xl` so we can keep the default xl value.

So the available breakpoint prefixes are now:
 
| Breakpoint prefix | Minimum width |
| ------------- | ------------- |
| `sm`| `40 rem (640px)`|
| `md`| `48rem (768px)`|
| `lg`| `64rem (1024px)`|
| `xl`| `80rem (1280px)`|
| `2xl`| `90rem (1440px)` – overriding the default 1536px|
| `3xl`| `110rem (1760px)`|

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -48,7 +48,6 @@
 }
 
 @theme inline {
-	--breakpoint-2xl: 90rem; /* 1440px */
 	--breakpoint-3xl: 110rem; /* 1760px */
 
 	--font-sans: 'Inter', var(--font-fallback);
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -294,7 +294,7 @@
 	grid-template-columns: 70px minmax(0, 8fr) 1fr;
 	@apply grid gap-x-8 px-4 sm:px-6;
 
-	@variant md {
+	@variant lg {
 		grid-template-columns: minmax(240px, 1fr) minmax(0, 4fr) minmax(160px, 1fr);
 		padding-inline: calc(var(--spacing) * 2);
 	}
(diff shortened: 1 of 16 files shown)


johanbissemattsson:
Should we instead make the default 2xl value into 3xl and add an extra 4xl breakpoint?

The available values would then be:

| Breakpoint prefix | Minimum width |
| ------------- | ------------- |
| `sm`| `40 rem (640px)`|
| `md`| `48rem (768px)`|
| `lg`| `64rem (1024px)`|
| `xl`| `80rem (1280px)`|
| `2xl`| `90rem (1440px)`|
| `3xl`| `96rem (1536px)`|
| `4xl`| `110rem (1760px)`|

jesperengstrom:
Are we dependant on 2xl being 90 rem for any styling to look good? Otherwise just let it be 96 rem?

johanbissemattsson:
> Are we dependant on 2xl being 90 rem for any styling to look good? Otherwise just let it be 96 rem?

👍 I have now reverted it to the default value in https://github.com/libris/lxlviewer/pull/1289/commits/83f93c45bb0787c5a824d9c3e4385ad0e25383e1

PR #1291, opened 2025-05-09
Display links to things in local systems

lrosenstrom:
### Tickets involved
[LWS-249](https://kbse.atlassian.net/browse/LWS-249)

### Solves

The goal is to provide access to items in the local library catalog as well as additional library info such as address, opening hours etc. This is currently presented in the holding panel:

Screenshot from 2025-05-22 09-17-55

### Summary of changes

* Construct a mapping for library code (sigel) to the links/info on fnurgel page load 
* Tried to keep close to the Figma sketches for the frontend part
* Add a basic server side caching mechanism for full holder/library data. There is plenty of information about caching in the browser in a SvelteKit context but not so much about caching on the server. The approach used in this PR comes from https://www.loopwerk.io/articles/2025/svelte-5-stores/, which should be fine since the data we are caching is not tied to a specific user/session/browser. I think introducing an in-memory database such as Redis is overkill for this, but that's what most people seem to be suggesting in the forums.

### DIFF
--- a/lxl-web/src/lib/utils/holdings.ts
+++ b/lxl-web/src/lib/utils/holdings.ts
@@ -224,7 +224,8 @@ export async function getFullHolderData(allHolders: DecoratedHolder[]): Promise<
 export function getItemLinksByBibId(
 	holderBySigel: FullHolderBySigel,
 	bibIdsByInstanceId: Record<string, BibIdObj>,
-	locale: LocaleCode
+	locale: LocaleCode,
+	displayUtil: DisplayUtil
 ): ItemLinksByBibId {
 	const linksByInstanceId: ItemLinksByBibId = {};
 	for (const bibIdObj of Object.values(bibIdsByInstanceId)) {
@@ -278,8 +279,18 @@ export function getItemLinksByBibId(
 			const visitingAddress = address.find((a) => a[JsonLd.TYPE] === 'bibdb:VisitingAddress');
 
 			if (address && address.length !== 0) {
-				addresses.push(postalAddress);
-				addresses.push(visitingAddress);
+				console.log(
+					'lensed and formatted',
+					displayUtil.lensAndFormat(postalAddress, LensType.Card, locale)
+				);
+				console.log('postal address', postalAddress as FramedData);
+				addresses.push(
+					toString(displayUtil.lensAndFormat(postalAddress as FramedData, LensType.Card, locale)) ||
+						''
+				);
+				addresses.push(
+					toString(displayUtil.lensAndFormat(visitingAddress, LensType.Card, locale)) || ''
+				);
 				allLinks['address'] = addresses;
 			}
 
--- a/lxl-web/src/lib/utils/holdings.ts
+++ b/lxl-web/src/lib/utils/holdings.ts
@@ -246,9 +246,23 @@ export function getItemLinksByInstanceId(
 				linksToItem = [linkTemplateEod.replace(/%BIBID%/, bibIdObj.bibId), ...linksToItem];
 			}
 
+			const allLinks: { [linkType: string]: string[] } = {};
+
+			const linksToSite: string[] = [];
+			const linkToSite = getAtPath(fullHolderData, ['bibdb:ils', 'url'], undefined);
+			if (linkToSite && linkToSite.length !== 0) {
+				linksToSite.push(linkToSite);
+				//TODO: formalize keys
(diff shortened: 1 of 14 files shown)


johanbissemattsson (approved):
LGTM! 🙌  

We probably want some kind of support for refreshed cache data in the future (so updated libraries data will be reflected in the cache without having to restart the server) but this can maybe be saved for later? I would however want a comment or issue created regarding this if the question arises.

lxl-web/src/lib/utils/holdings.ts
```diff
@@ -185,3 +203,163 @@ export function getMyLibsFromHoldings(
...
+				const libraryMainEntity = resJson['mainEntity'] as FramedData;
+				if (libraryMainEntity) {
+					cachedHolders[h.sigel] = libraryMainEntity;
+				}
+			} else {
+				console.error(`Could not fetch holder data for ${id}`);
+			}
+		}
+	}
+}
```
olovy, line 246: could be the same fetch?
error handling missing in the first version.

lxl-web/src/lib/utils/holdings.ts
```diff
@@ -185,3 +203,163 @@ export function getMyLibsFromHoldings(
...
+				];
+
+				const lopacPaths = [[BibDb.lopac, BibDb.bibIdSearchUriByLang]];
+
+				let linksToItem = getLinksToItemFor(bibIdObj, fullHolderData, ilsPaths, locale);
+				const lopacLinksItem = getLinksToItemFor(bibIdObj, fullHolderData, lopacPaths, locale);
+
+				const linkTemplateEod = getAtPath(fullHolderData, [BibDb.eodUri], []);
+				if (linkTemplateEod && linkTemplateEod.length !== 0) {
+					linksToItem = [linkTemplateEod.replace(/%BIB_*ID%/, bibIdObj.bibId), ...linksToItem];
```
olovy, line 273: this will only replace the first occurence of %BIB_*ID%
same comment for all `replace`

olovy (approved):
LGTM!

See comments.

As discussed offline:
* The key for holders/libraries should be `@id` not sigel
* records should be referenced by `@id`, not controlNumber. controlNumber is a MARC identifier and can contain old numeric Libris IDs. If systemIds/fnurgels are preferred because they are shorter in the URL they should be constructed by removing the system base URI prefix.

This behavior hasn't changed in this PR and can be fixed later.

lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/HoldingInfo.svelte
```diff
@@ -0,0 +1,356 @@
...
+			{#if bibIds.at(0)}
+				{@const firstBibId = bibIds.at(0).bibId}
+				<ul class="my-2" style="white-space: pre-line">
+					{#if linksByBibId[firstBibId]?.[holder.sigel]?.[BibDb.OpeningHours]}
+						<li>
+							{linksByBibId[firstBibId][holder.sigel][BibDb.OpeningHours].at(0)}
+						</li>
+					{/if}
+					{#if linksByBibId[firstBibId]?.[holder.sigel]?.[BibDb.Address]}
+						{#each linksByBibId[firstBibId][holder.sigel][BibDb.Address] as address (address)}
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/HoldingInfo.svelte:290]: can't use adress as an index here because it's sometimes missing...

lxl-web/src/lib/utils/holdings.ts
lrosenstrom: The first function is unused and was probably used before adding the cache+error handling. Must have missed to delete it! Good catch

jesperengstrom:
Nice work!

Apart from [comment](https://github.com/libris/lxlviewer/pull/1291/files#pullrequestreview-2864182329) that breaks the app for me, I mainly have comments regarding the look-and-feel which can be iterated. For example...

Skärmavbild 2025-05-23 kl  14 12 32

* I think this blue button is too wide :)
* I think we need some visual hierarchy with so much information. For example: indenting the opening hours a bit so it's visually a child of the summary label.
* Shouldn't we check if there are opening hours earlier and skip this section if it's missing? (lots of empty 'opening hours')

lrosenstrom:
Put item availability together with the rest of the item/instance info and made some improvements to the layout

Screenshot from 2025-05-26 14-40-15

PR #1296, opened 2025-05-14
chore(lxl-web, supersearch): Remove redundant vite override

johanbissemattsson:
### Solves

Removes redundant vite override as it seems to work as intended without it?

The esbuild dependency in vite was bumped in february (see: https://github.com/vitejs/vite/pull/19389).

### Summary of changes

- Remove redundant vite override


jesperengstrom (approved):
👍

PR #1297, opened 2025-05-19
feat(lxl-web): Add beta banner (LWS-360)

johanbissemattsson:
### Tickets involved
[LWS-360](https://kbse.atlassian.net/browse/LWS-360)

### Solves

Adds a beta banner with a link for adding feedback

### Summary of changes

- Add beta banner with feedback link
- Fix positioning of search dialog when banner exists

### DIFF
--- a/lxl-web/src/lib/i18n/locales/en.js
+++ b/lxl-web/src/lib/i18n/locales/en.js
@@ -9,8 +9,10 @@ export default {
 		changeLang: 'På svenska',
 		searchPlaceholder: 'Search for title, author, library, subject...',
 		openMenu: 'Open main menu',
-		betaMessage: 'This is a test version, the catalogue may contain outdated information.',
-		betaLink: 'Share your feedback'
+		betaMessage:
+			'This is a test version of the new Libris website, the catalogue may contain outdated information.',
+		betaLink: 'Share your feedback',
+		betaLinkLong: ' '
 	},
 	myPages: {
 		myPages: 'My pages',
--- a/lxl-web/src/lib/i18n/locales/en.js
+++ b/lxl-web/src/lib/i18n/locales/en.js
@@ -11,7 +11,7 @@ export default {
 		openMenu: 'Open main menu',
 		betaMessage: 'This is a test version of the new Libris.',
 		betaDetails: 'Incomplete information and errors may occur.',
-		betaLink: 'Go to old Libris'
+		betaLink: 'Share your feedback'
 	},
 	myPages: {
 		myPages: 'My pages',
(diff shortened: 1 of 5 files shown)


lxl-web/src/lib/i18n/locales/en.js
```diff
@@ -8,7 +8,9 @@ export default {
 		myPages: 'My pages',
 		changeLang: 'På svenska',
 		searchPlaceholder: 'Search for title, author, library, subject...',
-		openMenu: 'Open main menu'
+		openMenu: 'Open main menu',
+		betaMessage: 'This is a test version, so the catalogue may contain outdated information.',
```
olovy, line 12:
```suggestion
		betaMessage: 'This is a test version, so the catalogue may contain outdated information.',
```
```suggestion
		betaMessage: 'This is a test version, the catalogue may contain outdated information.',
```

olovy (approved):
:ship:

jesperengstrom (approved):
LGTM!

(Though I think I'm rooting for a more washed-out shade of yellow as bg, maybe `bg-warning-200` or `300`. Users will see this on every page and can't remove it, so I think it can't be too "in your face" or it will eventually be an annoyance)

jesperengstrom:
Also, Supersearch dialog is no longer on top of regular input because of the banner. Maybe we need a CSS variable `like --beta-banner-height` to be able to position the dialog vertically...

johanbissemattsson:
> Also, Supersearch dialog is no longer on top of regular input because of the banner. Maybe we need a CSS variable like `--beta-banner-height` to be able to position the dialog vertically...

Fixed in https://github.com/libris/lxlviewer/pull/1297/commits/84b2c82c56ac1a9f0339071e98d7f1d58d89979a

johanbissemattsson:
> Also, Supersearch dialog is no longer on top of regular input because of the banner. Maybe we need a CSS variable like `--beta-banner-height` to be able to position the dialog vertically...

I think we have to check the height of the banner with Javascript (as it can grow depending on the content length and screen size) :/

johanbissemattsson:
> > Also, Supersearch dialog is no longer on top of regular input because of the banner. Maybe we need a CSS variable like `--beta-banner-height` to be able to position the dialog vertically...
> 
> I think we have to check the height of the banner with Javascript (as it can grow depending on the content length and screen size) :/

Fixed in https://github.com/libris/lxlviewer/pull/1297/commits/74849fb5889f6a1d7d536789168cf0bb419fac9a using `bind:offsetHeight` on the banner and a CSS variable passed to `SuperSearchWrapper.svelte`

PR #1300, opened 2025-05-21
fix(lxlweb): use progress bar with modals (LWS-369)

jesperengstrom:
### Tickets involved
[LWS-369](https://kbse.atlassian.net/browse/LWS-369)

### Solves

The original case for this ticket has since resolved, i.e after searching from the expanded supersearch it now closes, revealing the progress bar under it. 

But there's another case that doesn't give any loading feedback. Adding/removing filters from the filter modal gives no loading indication since the progress bar is hidden behind it.

And placing nprogress on top of it seems impossible since the dialog sits at the `#top-layer` that is not affected by z-index etc.

So if we want to achieve this, there's basically these workarounds:

1) Close the modal after adding/removing a filter. This PR does this. Seems reasonable?
2) Adding another nprogress inside full screen modals
3) Somehow rework the modal to add it to the document flow. For example by using `show()` instead of `showModal()` (which has downsides). Anyway, the fact that we can't stack anything on top of our modals it worth keeping in mind when we start working with the panels.

### Summary of changes

* Close filter modal beforeNagivate in search results
* Add keys to each blocks

### DIFF
--- a/lxl-web/src/lib/components/find/SearchResult.svelte
+++ b/lxl-web/src/lib/components/find/SearchResult.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { page } from '$app/stores';
-	import { goto } from '$app/navigation';
+	import { beforeNavigate, goto } from '$app/navigation';
 	import Modal from '$lib/components/Modal.svelte';
 	import SearchMapping from './SearchMapping.svelte';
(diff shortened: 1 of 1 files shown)


jesperengstrom:
Upon second thought, this behaviour (closing modal after adding/removing filter) is maybe not what we want. Closing this for now...

PR #1301, opened 2025-05-22
fix(supersearch): Get virtual search keyboard (LWS-370)

jesperengstrom:
### Tickets involved
[LWS-370](https://kbse.atlassian.net/browse/LWS-370)

### Solves

Swapping inputmode for [enterkeyhint](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/enterkeyhint) seems to be the only way to get a 'search' btn for a contenteditable div in IOS at least 🤷‍♂️


johanbissemattsson (approved):
Nice find! 🙌 LGTM!

PR #1303, opened 2025-05-22
fix(lxlweb): Hide link popovers for touch devices (LWS-376)

jesperengstrom:
### Tickets involved
[LWS-376](https://kbse.atlassian.net/browse/LWS-376)

### Solves

Hide link popovers for touch devices since they're of no use anyway.
(We also use popovers for buttons (focus-triggered), but they work on mobile since they don't trigger navigations)

### Summary of changes

Add a class if popover is attached to a link. Hide those popovers for devices where hover is unsupported.


johanbissemattsson (approved):
LGTM!

PR #1307, opened 2025-05-26
feat(lxlweb): better mobile pagination (LWS-361)

jesperengstrom:
### Tickets involved
[LWS-361](https://kbse.atlassian.net/browse/LWS-361)

### Solves

Pagination consists of `first`, `prev`, `next` & `last`, plus a sequence of x pages in the middle, including ellipsis if there's a gap to first/last. This sequence should be longer on bigger screens & shorter on smaller. 
Ideally we would just render as many numbers as we can fit and hide the rest with css. But that would make it impossible to detect a numerical gap to add the ellipsis, for example.

The idea here is instead to create two versions of the sequence, for small & medium screens. We don't want to use too much js to determine which one to show, so we basically render them both using a snippet, then use media queries to visually show the right one. 

For small screens, one could imagine there's room for more than 3 numbers in the sequence. But we need to take into account [4-digit page numbers](http://localhost:5173/find?_limit=20&_spell=true&_q=*&_offset=99880)... 

Maybe there's some smart way to make this even more flexible and relying even more on CSS. Suggestions are welcome...

### Summary of changes

* Refactor Pagination component
* Fix bug where Safari did not show the supersearch lint red squiggly line

### DIFF
--- a/lxl-web/src/lib/components/find/Pagination.svelte
+++ b/lxl-web/src/lib/components/find/Pagination.svelte
@@ -12,23 +12,13 @@
 	const { first, last, next, previous, totalItems, itemsPerPage, itemOffset, maxItems } =
 		$derived(data);
 
-	const numberOfPagesSm = 3;
-	const numberOfPagesMd = 7;
+	const MAX_PAGES_SM = 3;
+	const MAX_PAGES_MD = 7;
 
 	const showPagination = $derived(data.items.length > 0 && totalItems > itemsPerPage);
 	const currentPage = $derived(Math.floor(itemOffset / itemsPerPage) + 1);
 	const lastItem = $derived(totalItems > maxItems ? maxItems : totalItems);
 	const lastPage = $derived(Math.ceil(lastItem / itemsPerPage));
-
-	const pageSequenceSm = $derived.by(() => {
-		let pagesSm = numberOfPagesSm > lastPage ? lastPage : numberOfPagesSm;
-		return getPages(pagesSm);
-	});
-
-	const pageSequenceMd = $derived.by(() => {
-		let pagesMd = numberOfPagesMd > lastPage ? lastPage : numberOfPagesMd;
-		return getPages(pagesMd);
-	});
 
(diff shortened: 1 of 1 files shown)


lxl-web/src/lib/components/find/Pagination.svelte
```diff
@@ -9,18 +9,28 @@
...
 
 	const { data }: PaginationProps = $props();
-
-	const numberOfPages = 7;
-
 	const { first, last, next, previous, totalItems, itemsPerPage, itemOffset, maxItems } =
 		$derived(data);
+
+	const numberOfPagesSm = 3;
+	const numberOfPagesMd = 7;
```
johanbissemattsson, line 16: I would uppercased naming here for truly immutable constant values e.g. `NUMBER_OF_PAGES_SM` (or `MAX_PAGES_SM`)?

lxl-web/src/lib/components/find/Pagination.svelte
```diff
@@ -9,18 +9,28 @@
...
+	const pageSequenceSm = $derived.by(() => {
+		let pagesSm = numberOfPagesSm > lastPage ? lastPage : numberOfPagesSm;
+		return getPages(pagesSm);
+	});
+
+	const pageSequenceMd = $derived.by(() => {
+		let pagesMd = numberOfPagesMd > lastPage ? lastPage : numberOfPagesMd;
+		return getPages(pagesMd);
+	});
+
```
johanbissemattsson, line 32: I feel these derived statements are bit redundant when we instead could use snippet parameters directly in the templates.

e.g.
```
{#snippet sequence(size: number)}
	{@const pages = getPages(Math.min(size, lastPage))}
```
together with `{@render sequence(MAX_PAGES_SM)}`

jesperengstrom: Very nice! Added in https://github.com/libris/lxlviewer/pull/1307/commits/f31722cfe8ecc8a74a08a97eefae7380bf50e95f

PR #1309, opened 2025-05-27
fix(lxlweb): Adjust toolbar areas

jesperengstrom:
## Description

Adjusting the layout of the search result "toolbar" so that it looks a bit less randomly thrown on the page. 
(Until this gets a brand new look after redesign)

before 
Skärmavbild 2025-05-26 kl  16 35 52
after
Skärmavbild 2025-05-26 kl  16 35 32

### Summary of changes

* Layout adjustments to SearchResult.svelte
* Remove some `@apply`
* Add two `each` keys


jesperengstrom:
> LGTM! 🙌 The filter button and sort select could perhaps have equal width on small screen sizes?

Yep, the button should grow to take up the available space (to a certain point). The sort-select will we as wide as the longest option label, so it won't shrink.

PR #1315, opened 2025-06-02
Bugfix/holding modal etc

lrosenstrom:
## Description

Cleanup and minor fixes

### Summary of changes

* Allow holder label to take up two rows before ellipsing
* Don't show opening hours/address, if empty
* Remove unused CSS 
* Default to locale code 'sv' if missing 'en' in language-tagged link templates

### DIFF
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
@@ -52,7 +52,7 @@
 </script>
 
 <li class="border-neutral text-sm not-last:border-b">
-	<h2 class="holder-label mt-4 line-clamp-1">
+	<h2 class="mt-4 line-clamp-2 text-base font-medium">
 		<DecoratedData data={holder.obj} showLabels={ShowLabelsOptions['Never']} />
 	</h2>
 	<div class="">
@@ -101,10 +101,8 @@
 	</div>
 	{#if hasOpeningHoursEtc()}
 		<details>
-			<summary class="my-3 flex cursor-pointer items-baseline">
-				<span
-					class="text-3xs arrow text-subtle mr-0.5 h-3 origin-center rotate-0 transition-transform"
-				>
+			<summary class="my-3 flex cursor-pointer items-baseline gap-0.5">
+				<span class="text-3xs arrow text-subtle h-3 origin-center rotate-0 transition-transform">
 					<BiChevronRight />
 				</span>
 				{page.data.t('holdings.openingHoursEtc')}
@@ -162,10 +160,6 @@
 		& .arrow {
 			@apply rotate-90;
 		}
-	}
-
-	.holder-label {
-		@apply line-clamp-2 text-base font-medium;
 	}
 
 	.status-container {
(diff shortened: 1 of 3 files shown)


lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
```diff
@@ -40,10 +40,19 @@
...
+			id &&
+			(linksByBibId[id.bibId]?.[holder.sigel]?.[BibDb.OpeningHours] ||
+				linksByBibId[id.bibId]?.[holder.sigel]?.[BibDb.Address])
+		);
+	}
 </script>
 
 <li class="border-neutral text-sm not-last:border-b">
-	<h2 class="holder-label mt-4">
+	<h2 class="holder-label mt-4 line-clamp-1">
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte:55]: `line-clamp-1` added here and `line-clamp-2` included in `holder-label class`...

lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
```diff
@@ -40,10 +40,19 @@
 	function missingAtLeastOneLinkToItem() {
 		return bibIds.find((id) => !hasLinksToItem(id.bibId) && !hasLoanReserveLink(id.bibId));
 	}
+
+	function hasOpeningHoursEtc() {
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte:44]: I noticed this is not really working. The function seems to return true while the info is still missing. See for example "Bilblioteken i Lomma"
http://localhost:5173/cv4j8tk59bwd3p88?holdings=Print

This is because the opening hours are arrays that has multiple empty elements. You'll have to join it before checking, i.e

```
		if (id) {
			const openingHours = linksByBibId[id.bibId]?.[holder.sigel]?.[BibDb.OpeningHours];
			const adress = linksByBibId[id.bibId]?.[holder.sigel]?.[BibDb.Address];

			return ((openingHours || adress?.join('').trim()))
		} return false;
```

lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
```diff
@@ -151,15 +162,10 @@
...
-
-		& .holder-label {
-			white-space: normal;
-		}
 	}
 
 	.holder-label {
-		@apply flex-1 overflow-hidden text-ellipsis whitespace-nowrap;
-		@apply text-base font-medium;
+		@apply line-clamp-2 text-base font-medium;
```
Comment (jesperengstrom) [inline lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte:168]: Would prefer to remove `.holder-label` and move these styles inline (see above comment). We are trying to move away from using `@apply` (since it really duplicates styles) and it's not really needed here...

lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
Comment (lrosenstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte:55]: Woops!

lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte
Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/[fnurgel=fnurgel]/Holdings.svelte:44]: Sorry, it's adress that's often empty strings. The function would return true for all three of these.

Skärmavbild 2025-06-03 kl  14 49 28

jesperengstrom (approved):
LGTM!

PR #1316, opened 2025-06-02
fix(lxl-web): Show search start content less often (LWS-372)

johanbissemattsson:
### Tickets involved
[LWS-372](https://kbse.atlassian.net/browse/LWS-372)

### Solves

Shows the search start content a little less often as it could become intrusive when searching, e.g. when adding spaces after a string part.

The start content is now only shown if the value is empty or if the edited part is **NOT** part of a qualifier or a string  (including spaces).

TODO: fix start content visiblity when editing inside parentheses/groups (probably saved for a later PR).

### Summary of changes

- Show start content if not editing part of qualifier or string
- Ensure text cursor is visible before qualifier

### DIFF
--- a/lxl-web/src/lib/styles/lxlquery.css
+++ b/lxl-web/src/lib/styles/lxlquery.css
@@ -39,9 +39,6 @@
 	border-radius: 0;
 }
 
-.lxl-qualifier-key {
-	margin-left: 1px;
-}
 .atomic {
 	background-color: var(--color-accent-100);
 	user-select: none;
--- a/packages/supersearch/src/lib/extensions/lxlQualifierPlugin/index.ts
+++ b/packages/supersearch/src/lib/extensions/lxlQualifierPlugin/index.ts
@@ -70,7 +70,7 @@ class QualifierWidget extends WidgetType {
 	}
 	toDOM(): HTMLElement {
 		const container = document.createElement('span');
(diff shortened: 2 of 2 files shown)


lxl-web/src/lib/styles/lxlquery.css
```diff
@@ -39,6 +39,9 @@
 	border-radius: 0;
 }
 
+.lxl-qualifier-key {
```
jesperengstrom, line 42: Nice, but there's a similar problem on the other side of the pill + this does not cover 100% of the cases (filter aliases are atomic pills without qualifier key).

I propose adding `margin-left` + `margin-right` 1px to the widget wrapper element created by the plugin. I've been tampering with that to create a proper stylable wrapper element for all the pills. But this could be a start... Located at supersearch/extensions/lxlQualifierplugin/index.ts l. 71

jesperengstrom (approved):
See comment. Otherwise working nicely I think. A good starting point for investigating user preferences further.

lxl-web/src/lib/styles/lxlquery.css
johanbissemattsson: Good call, fixed in https://github.com/libris/lxlviewer/pull/1316/commits/8a8bd5aec405cfd57dabed304e39afcc00f2c3f8.

PR #1319, opened 2025-06-03
feat(lxlweb): Leading pane (LWS-385)

jesperengstrom:
[LWS-385](https://kbse.atlassian.net/browse/LWS-385)

## Description

Welcome to try out a rezisable left pane, including:
* A sticky `AppBar` header component on /find route (static elsewhere) (to replace `SiteHeader`)
* A sticky, collapsible, resizable `LeadingPane` component, saving the state & width in cookie. 
  * ...containing ye olde facets with some design tweaks 
  * Using a `Draggable` component + throttle util more or less copied from the prototype
* A sticky aside component that is positioned using a `content` container query
* The old massive `SearchResult` component temporarily replaced by `SearchResult-2`, upgraded to svelte 5 syntax. Some of its sections have been broken out to own components, including
  * `SearchResultToolbar` using the reusable new `Toolbar` component
  * `SearchResultInfo`
  * `SearchResultSort`
 
* For mobile: A sticky header with a non sticky search field + sticky toolbar

### Things to ponder

* Does the sticky approach feel ok? As opposed to having fixed sections of max 100vh that scrolls individually... The beta banner is sort of the problem here, as it would be always visible with a fixed layout 🤔 
* Sticky + body lock for opened expanded search does not work very well together. You lose the scrolling position...
* We now need _two_ SuperSearch components since (according to design) it should be placed inside the sticky `AppBar` on desktop, but outside of it on mobile. Not too happy about that...


olovy (approved):
Great work! LGTM!

PR #1320, opened 2025-06-04
feat(lxl-web): Add banner link to old libris

johanbissemattsson:
### Tickets involved
[LWS-360](https://kbse.atlassian.net/browse/LWS-360)

### Solves

Add banner link to old Libris

### Summary of changes
- Add banner link to old libris
- Adjust footer padding so it harmonizes with the header


johanbissemattsson:
> A bit of a shame that the banner went from two to three rows on screens < 380px. I kind of think we should remove "Tyck till" in those cases...

It was even worse in english so I did what you suggested in https://github.com/libris/lxlviewer/pull/1320/commits/5c14190bbf57e61c4fb9045cd8a40d14f84d1b4c. I prefer it as well...

PR #1321, opened 2025-06-04
fix(lxl-web): Wildcard flash when searching (LWS-379)

johanbissemattsson:
### Tickets involved
[LWS-379](https://kbse.atlassian.net/browse/LWS-379)

### Solves

Fixes flashes of wildcard results which could appear while searching (e.g. when the user types after having added a qualifier).

### Summary of changes

- Call `shouldShowStartContentFn` only once
- Fix flash of wildcard results when searching


johanbissemattsson:
I'm not entirely happy with the fix as it feels a tad bit hackish so I'm open for better solutions...

jesperengstrom (approved):
LGTM!

PR #1326, opened 2025-06-12
fix(lxlweb): fix range bug after navigating + blur input (LWS-381)

jesperengstrom:
### Tickets involved
[LWS-381](https://kbse.atlassian.net/browse/LWS-381)

### Solves

A bug (chrome only?) that causes a range error when navigating using a supersearch suggestion, to a resource page and continue to type in the focused input field.

Skärmavbild 2025-06-11 kl  13 41 03

Also, the collapsed input should probably _not_ be focused after navigation, to be consistent with other cases where it's not (performing a search, clicking a facet, clicking a result card) to convey a "finished" state after performing an action. 

The blur function should be separate however, since there are cases where we want to keep input focused after collapsing expanded search (click outside, press `esc`).

~~Todo: Cover this case in test, see https://github.com/libris/lxlviewer/pull/1325~~

### Summary of changes

* Sync the expanded search selection with the collapsed in `handleCloseExpandedSearch` (fixing the bug)
* export a `blur` function and call it after navigation in `superSearchWrapper` (removing the input focus)

### DIFF
--- a/lxl-web/tests/supersearch-lxlweb.spec.ts
+++ b/lxl-web/tests/supersearch-lxlweb.spec.ts
@@ -36,3 +36,11 @@ test('navigate to suggested resource using keyboard', async ({ page }) => {
 	await page.keyboard.press('Enter');
 	await expect(page.locator('.resource-page')).toBeVisible();
 });
+
+test('input loses focus when navigating', async ({ page }) => {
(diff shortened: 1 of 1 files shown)


johanbissemattsson (approved):
LGTM! 🙌

PR #1329, opened 2025-06-18
feat(lxlweb): implement _suggest (LWS-272)

kwahlin:
It's a mess but should at least give some guidance on how to use the new suggest functionality in the find API (https://github.com/libris/librisxl/pull/1602).

### Edit 26/6:
Made the following additions:
* Remove all unused code
* Added a new api/supersearch util function `insertWildcard` that sometimes modifies `_q` before passing on to `/find`,  inserting a wildcard inside an empty qualifierValue ("" or ()). This is needed if we want initial suggestions after adding a qualifier as it will not be handled by the backend.
* modified `DropdownMenu` so that is produces links instead of buttons, since the actions now come as links. It should maybe be able to produce both links and buttons, let's refactor it when the need arises.

**Tests failing:**
```
Locator: getByRole('dialog').getByLabel('Förslag').getByRole('link').filter({ hasText: 'Person' })
    Expected: 5
    Received: 0
```
It actually does its job here, as contributor qualifiers does not give Person suggestions. This is because some todo update of Definitions (@kwahlin ?) So basically - **we should not merge this until the tests passes**

**Possible TODO**
* We get the `cursor` back from the api response, but does nothing with it. The input now always loses focus on navigation. Maybe there is some case where we want to re-apply the cursor (after adding a qualifier?)

### DIFF
--- a/lxl-web/tests/supersearch-lxlweb.spec.ts
+++ b/lxl-web/tests/supersearch-lxlweb.spec.ts
@@ -143,7 +143,7 @@ test('qualifier keys can be added using the user interface', async ({ page }) =>
 		page.getByRole('combobox').locator('.lxl-qualifier-key').last(),
 		'pills for both contributor and language exists'
 	).toContainText('Språk');
-	await page.getByTestId('main-search').click({ position: { x: 10, y: 10 } }); // do not accidentally press a remove button on a puill
+	await page.getByTestId('main-search').click({ position: { x: 10, y: 10 } }); // Make sure to not accidentally click the remove button of a pill
 	await page.keyboard.press('Home'); // for PCs
 	await page.keyboard.press('Meta+ArrowLeft'); // for mac
 	await page


jesperengstrom:
Fixed tests in https://github.com/libris/lxlviewer/pull/1329/commits/ff25467c9627d4d894235a614e88ba4eb4e82929
Almost went crazy trying to understand why the dialog did not open on click, before I realized a pill x-button happened to be located at the place of the click  🙃 

Skärmavbild 2025-06-26 kl  21 48 32

jesperengstrom:
... But in the process I fixed a bug where the cursor sometimes jumps out of a qualifier "template". `addQualifierKey` now only inserts `key:`. `insertQuotes` and `addSpaceAroundQualifiers` takes care of the rest.

johanbissemattsson (approved):
Awesome work @kwahlin and @jesperengstrom! 🌟 

When testing it seems to work as intended (it also feels more performant due to links being preloaded on hover out-of-the-box). And the frontend code gets much simpler.

Again, great work!

PR #1330, opened 2025-06-19
feat(lxlweb/facets): Save facet expanded state. Move free online to top.

olovy:
- Save facet expanded state in userSettings. Default to 4 first open
- Fix bug where facets were not expanded during "find facet" filtering.
- Place 'free online' inside a new "Access" facet
    - Until we have a more detailed design
- Add new FacetValue component (still TODO to rename FacetGroup -> Facet)
    - this refactoring was because of the abandoned idea to place "free online" inside the myLibraries box. Could be reverted.

### DIFF
--- a/lxl-web/src/lib/components/find/FacetGroup.svelte
+++ b/lxl-web/src/lib/components/find/FacetGroup.svelte
@@ -23,10 +23,10 @@
 		group: FacetGroup;
 		locale: LocaleCode;
 		searchPhrase: string;
-		defaultExpanded: boolean;
+		isDefaultExpanded: boolean;
 	};
 
-	let { group, locale, searchPhrase, defaultExpanded }: FacetGroupProps = $props();
+	let { group, locale, searchPhrase, isDefaultExpanded }: FacetGroupProps = $props();
 
 	const matomoTracker = getMatomoTracker();
 	const userSettings = getUserSettings();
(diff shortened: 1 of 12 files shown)


jesperengstrom:
> * Saving expansion state & opening facets when searching are (unintentionally?) linked together - so that when you search 'a', most facets open and saving that as a preference, effectively overwriting some more deliberate setting. Expansion state should only be saved when actively opening/closing by click. Also, facets should probably go back to this saved/default state when the search input is cleared.

fixed in https://github.com/libris/lxlviewer/pull/1330/commits/09314e913d456fa980ce0686cd8b99d51d49158b
Added `onclick` handler to the `summary` element that saves the user setting. This hijacks the native toggle behaviour, but falls back to that with js turned off...

jesperengstrom:
> 'Limit to my libraries' could really fit inside the 'access' facet, to save some space and avoid having two separate one-purpose sections.

On second thought I will make a follow-up PR of this.

johanbissemattsson (approved):
LGTM!

I would prefer moving the logic controlling if the facet group should be expanded or not to the parent component `Filters.svelte`, so `FaceGroup.svelte` can act more like a _dumb_ component (unaware of the surroundings).  But it would require a major rewrite so I think this looks good for now!

PR #1334, opened 2025-06-24
feat(lxlweb): move myLibraries to access filters

jesperengstrom:
### Tickets involved
--

### Solves

Move the `myLibraries` filter from own component into the access facets (you can now always use it, but without having chosen a library, a message will display, which feels slightly more user friendly than before?):

Skärmavbild 2025-06-24 kl  14 21 47

Since this is a dynamic (non hard-coded) filter, it only exists in facet api responses if active - which means we need to hard code it ourselves. This hack was previously done in the frontend component, but is now done in a new search.ts function, meaning we can continue to just loop out the facets. This seems slightly less of a hack, but still, constructing links without language parsing is not a good idea. I think this filter should be hardcoded as well in the future.

### Summary of changes

* Add `alias-myLibraries` to access filter 'bucket' in facets
* `addMyLibrariesBoolFilter` function in search.ts to manually add it as a bool filter if missing
* Display an 'edit' button next to the filter in facets
* Display a message if filter is active but no libraries chosen
* Remove `MyLibrariesFilter` component
* Add tests


johanbissemattsson (approved):
LGTM! I think it's more user friendly now and I also like the space savings. 🙌 

I realized now that we should probably add the myLibraries filter (and available free online) to the persistent filter suggestions in the expanded supersearch dialog, but that is best saved for later.

PR #1336, opened 2025-06-26
Feature/link to holdings from hit list

lrosenstrom:
## Description

Allow access to books etc directly from the search results view.

Mobile:

Screenshot from 2025-06-26 11-38-19

Desktop:

Screenshot from 2025-06-26 11-37-09

### Tickets involved
[LWS-384](https://kbse.atlassian.net/browse/LWS-384)

### Summary of changes

* Extract relevant parts of product page to a new HoldingsModal component.
* Define a new api route for fetching everything that is needed for the holdings modal.
* Move data initialization taking place in the product page load function to the new api route, making it accessible from other parts of the app, such as the search results list. 
* Add button for accessing holding modal in search card.

It should be straight forward to make this align with the new panes design, but that can be done in a separate PR.

### DIFF
--- a/lxl-web/src/lib/components/find/SearchCard.svelte
+++ b/lxl-web/src/lib/components/find/SearchCard.svelte
@@ -8,11 +8,12 @@
 	import getInstanceData from '$lib/utils/getInstanceData';
 	import placeholder from '$lib/assets/img/placeholder.svg';
 	import DecoratedData from '$lib/components/DecoratedData.svelte';
-	import { page } from '$app/stores';
+	import { page } from '$app/state';
 	import SearchItemDebug from '$lib/components/find/SearchItemDebug.svelte';
 	import EsExplain from '$lib/components/find/EsExplain.svelte';
 	import SearchItemDebugHaystack from '$lib/components/find/SearchItemDebugHaystack.svelte';
 	import MyLibsHoldingIndicator from '$lib/components/MyLibsHoldingIndicator.svelte';
+	import { getHoldingsLink, handleClickHoldings } from '$lib/utils/holdings';
 
 	export let item: SearchResultItem;
 
@@ -43,7 +44,7 @@
 							src={item.image.url}
 							width={item.image.widthPx > 0 ? item.image.widthPx : undefined}
 							height={item.image.heightPx > 0 ? item.image.heightPx : undefined}
-							alt={$page.data.t('general.latestInstanceCover')}
+							alt={page.data.t('general.latestInstanceCover')}
 							class:rounded-full={item['@type'] === 'Person'}
 							class="object-contain object-top {item['@type'] !== 'Person'
 								? 'aspect-2/3'
@@ -132,7 +133,7 @@
 						{#if instances?.years}
 							{#if instances.count > 1}
 								{instances?.count}
-								{$page.data.t('search.editions')}
+								{page.data.t('search.editions')}
 								{`(${instances.years})`}
 							{:else}
 								{instances.years}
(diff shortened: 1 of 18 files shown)


lxl-web/src/lib/components/find/SearchCard.svelte
```diff
@@ -191,11 +221,9 @@
 				{/if}
 			{/key}
 		{/if}
-		{#if item.heldByMyLibraries?.length}
-			<div class="card-libraries flex items-start">
-				<MyLibsHoldingIndicator libraries={item.heldByMyLibraries} />
-			</div>
-		{/if}
+		<div class="hidden md:inline">
```
jesperengstrom, line 224: Also, I feel the layout could be achieved without rendering the button twice. 

Using 
```
grid-template-areas : 
'image content debug'
'. libraries libraries'
```
For smaller screens, for example. Also, it should use a container query so it wraps when resizing the panels, not only the window.

jesperengstrom: But I also wonder if the `libraries` section should maybe a column of the search card `footer` rather than the entire search card? The current layout takes up width of the entire card that could be used for long titles in narrow cards, for example.

Skärmavbild 2025-06-27 kl  13 41 18

lxl-web/src/lib/components/find/SearchCard.svelte
```diff
@@ -25,6 +27,30 @@
 	let showDebugHaystack = false;
 </script>
 
+{#snippet holdingsButton()}
+	<div class="card-libraries flex items-start pt-1">
+		{#if id}
+			<a
+				class="btn btn-primary h-7 rounded-full md:h-8"
+				href={getHoldingsLink(page.url, id)}
+				data-sveltekit-preload-data="false"
```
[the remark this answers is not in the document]
johanbissemattsson: > the presence of a `holdings` param causes a (lazyloaded) fetch in the search/load function, passing the data to the modal, rather than it being loaded `onMount`?

I think there are arguments for both using preloading and not preloading in this case:

- Preloading leads to better performance / instantaneously loaded modals.
- Clicking on a holdings button on a browser tab that has left opened for a while will show outdated data (it could be seen as crucial that the holdings data is up-to-date).

[Streaming](https://svelte.dev/docs/kit/load#Streaming-with-promises) the holdings data (using the holdings search params) would lead to some performance improvement (the data can start loading before the component has been mounted) while still showing up-to-date information.

lxl-web/src/lib/components/find/SearchCard.svelte
```diff
@@ -191,11 +221,9 @@
 				{/if}
 			{/key}
 		{/if}
-		{#if item.heldByMyLibraries?.length}
-			<div class="card-libraries flex items-start">
-				<MyLibsHoldingIndicator libraries={item.heldByMyLibraries} />
-			</div>
-		{/if}
+		<div class="hidden md:inline">
+			{@render holdingsButton()}
```
jesperengstrom, line 225: Also, I feel the layout could be achieved without rendering the button twice. 

Using 
```
grid-template-areas : 
'image content debug'
'. libraries libraries'
```
For smaller screens, for example. Also, it should use a container query so it wraps when resizing the panels, not only the window.

jesperengstrom: But I also wonder if the `libraries` section should maybe a column of the search card `footer` rather than the entire search card? The current layout takes up width of the entire card that could be used for long titles in narrow cards, for example.

Skärmavbild 2025-06-27 kl  13 41 18

johanbissemattsson (approved):
> Nice work! I made some comments - I'm ok with merging this now and continue working on it (since this is a much-requested feature), or polish it some more depending on @johanbissemattsson feedback.
> 
> The feature itself is working well I think.

I agree! We can absolutely merge this and polish it more later.

jesperengstrom:
> [Streaming](https://svelte.dev/docs/kit/load#Streaming-with-promises) the holdings data (using the holdings search params) would lead to some performance improvement (the data can start loading before the component has been mounted) while still showing up-to-date information.

Yes, streaming was the term I was looking for but confused with something else

lrosenstrom:
Great feedback. I'll look into the issues mentioned after merge.

PR #1337, opened 2025-06-26
bugfix(supersearch): Fix handleReset (LWS-392)

jesperengstrom:
### Tickets involved
[LWS-392](https://kbse.atlassian.net/browse/LWS-392)

### Solves

Bug in Chrome where you cannot type in input after clicking the clear button
Skärmavbild 2025-06-19 kl  13 14 49

### Summary of changes

* Dispatch a delete event in `handleReset` function, triggering a sync update between the instances in CodeMirror
* Add tests to cover this case


johanbissemattsson (approved):
LGTM! 🌟

PR #1339, opened 2025-07-09
feat(lxl-web): Simplified toolbar component

johanbissemattsson:
### Solves

Simplifies the toolbar component.

### Summary of changes

- Unnecessarily rendered empty list elements are removed (they were previously rendered if any of the snippet props were undefined).

- The list elements are reworked into normal div elements as it isn't certain that the contents of the toolbar makes sense in the context of a list. However, feel free to add lists _inside_ the leadingActions/trailingActions snippets e.g. lists of toolbar buttons).

- The `centerActions` snippet prop is removed in favour of normal children props.


johanbissemattsson:
The names of the snippets `leadingActions` and `trailingActions` implies lists so they maybe actually should be lists... 🤔 

We can decide this later.

PR #1340, opened 2025-07-11
feat(lxl-web): One-to-many resource pages (LWS-386)

johanbissemattsson:
### Tickets involved
[LWS-386](https://kbse.atlassian.net/browse/LWS-386)

### Solves

Redesigned resource pages with support for showing sections previewing different one-to-many relations.
These relations are currently displayed as scrollable lists (scroll buttons should be added in a subsequent PR!).

See screenshot:
Skärmavbild 2025-07-11 kl  14 53 33

### Summary of changes

- Add new reusable resource component ([Resource.svelte](https://github.com/libris/lxlviewer/blob/183807d12bfaff945054b840fea1bd41fcf78fce/lxl-web/src/lib/components/Resource.svelte)) – which will allow multiple panes showing different resource data at the same time.
- Add table of contents component ([TableOfContents.svelte](https://github.com/libris/lxlviewer/blob/183807d12bfaff945054b840fea1bd41fcf78fce/lxl-web/src/lib/components/TableOfContents.svelte)) which highlights the first visible section using [intersection observers](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).
- Add scrollable lists of relations to resource pages
- Add [SearchResultList.svelte](https://github.com/libris/lxlviewer/blob/183807d12bfaff945054b840fea1bd41fcf78fce/lxl-web/src/lib/components/SearchResultList.svelte) and [SearchResultItem.svelte](https://github.com/libris/lxlviewer/blob/183807d12bfaff945054b840fea1bd41fcf78fce/lxl-web/src/lib/components/SearchResultItem.svelte) components (there is currently some code repetition between these components and [SearchCard.svelte](https://github.com/libris/lxlviewer/blob/183807d12bfaff945054b840fea1bd41fcf78fce/lxl-web/src/lib/components/find/SearchCard.svelte) but the goal is let the new components be more flexible and reusable in multiple situations (so the page-specific SearchCard.svelte component could become obsolete).
- Add more container breakpoints
- Add wide layout variant (currently not configurable by the user but can be manually activated by adding the class `.prefers-wide-layout`).
- Add utility class for hidden scrollbars (`scrollbar-hidden`)
- Migrate `<ResourceImage />` component to Svelte 5 syntax
- Hide full resource image link on mobile
- Remove `<SearchResultOld />` component
- Replace old `getRelated/searchPromise` with new `getRelations` function together with preview fetches for each relation section/list.
- And numerous styling fixes...


olovy (approved):
LGTM!!

Seems to work fine. 
Except for relations that point to other things than works/instances, e.g. hasOccupation. 
Preview says Show all X results but list is empty. This is of course because of the hidden default "Type:Work" filter. We have to decide if the API should automatically "coerce" these queries containing a property with a domain with other types. Not a blocker. The client should anyway probably not rewrite the queries itself.

Did some small changes:
* Add isPartOf to relationsByBaseClass
* Move "related" section below instance list.
* Top heading "Inledning" -> "Översikt" as in the sketches
* Always highlight the clicked heading in TOC even if there is a visible section before it in the page. Before, the last heading/section was never highlighted which felt weird.

PR #1341, opened 2025-07-23
fix(cataloging): don't treat certain terms as fnurgels

andersju:
Some id.kb.se terms were displayed weirdly because they contained exactly 15 or 16 characters and were treated as fnurgels. Fix that + also decode URIs while we're at it (or?). 

https://kbse.atlassian.net/browse/LXL-3786


olovy (approved):
:ship:

PR #1344, opened 2025-08-12
chore(lxlweb, supersearch): Bump dependencies

johanbissemattsson:
### Solves

Bumps dependencies in `lxl-web` and `supersearch` (together with `codemirror-lang-lxlquery`)

### Summary of changes

- Bump dependencies to latest version with some exceptions... see below 👇 
- `sveltejs/kit` uses `2.27.1` instead of `2.28.0` due to a bug causing tests to fail
- `codemirror/view` is kept at `6.36.6` (instead of `6.38.1`) as `aria-placeholder` added in `6.36.7` doesn't play nicely with `role="combobox"`, causing AXE accessibility tests to fail. We should decide if we should supress these errors or work around it in some other way.
- Upgrade vite to version 7 (this will add Node ^22.12 as a requirement).
- Harmonize package.json in `packages/supersearch` with Sveltes library starter template. Note that `npm run prepack` or  `npm pack` should now be used instead of `npm run prepare`
- Add `lang="ts"` to `<script module>`
- Fix issue with multiple candidate TSConfigRootDirs


jesperengstrom (approved):
Nice job! ⭐ 
Everything seems to work just fine.

PR #1345, opened 2025-08-13
chore(lxlweb): Migrate to Svelte 5 syntax

jesperengstrom:
### Tickets involved
[LXL-NNNN](https://kbse.atlassian.net/browse/LXL-NNNN), [LXL-NNNN](https://kbse.atlassian.net/browse/LXL-NNNN)

### Solves

Description of what this solves

### Summary of changes

Summary of changes


lxl-web/src/lib/components/Modal.svelte
```diff
@@ -45,8 +59,8 @@
 <dialog
 	class="fixed top-0 left-0 flex h-screen max-h-full w-screen max-w-full"
 	tabindex="-1"
-	on:click|self={handleBackdropClick}
```
jesperengstrom, line 48: Seems like the `self` modifier isn't needed since `handleBackdropClick` checks if the click came from the element itself?

lxl-web/src/lib/components/Matomo.svelte
```diff
@@ -35,4 +40,4 @@
 		<script async defer src={`${URL}/matomo.js`} on:load={onMatomoScriptLoad}></script>
```
jesperengstrom, line 40: `on:load` here was the one thing I wasn't able to migrate without hassle. Native `onload` violates our CSP policy, blocking the script. 
`Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution.`

The old way still works, just leave it be?

lxl-web/src/lib/components/MarkdownContent.svelte
```diff
@@ -1,5 +1,13 @@
+<script lang="ts">
+	interface Props {
+		children?: import('svelte').Snippet;
```
johanbissemattsson, line 3: I haven't encounterd this way of defining snippet types before. I guess that there's nothing inherently wrong with it but we should maybe be consistent througout the codebase.

See https://svelte.dev/docs/svelte/snippet#Typing-snippets

lxl-web/src/lib/components/Matomo.svelte
```diff
@@ -2,6 +2,11 @@
 	import { env } from '$env/dynamic/public';
 	import { afterNavigate } from '$app/navigation';
 	import { setMatomoTracker, getMatomoTracker, setMatomoContext } from '$lib/contexts/matomo';
+	interface Props {
+		children?: import('svelte').Snippet;
```
johanbissemattsson, line 6: Same as previous comment regarding importing snippet types.

lxl-web/src/lib/components/Matomo.svelte
johanbissemattsson: I cannot seem to reproduce this issue (when testing with both `onload` and `on:load` in Firefox and Chrome – both when running `npm run dev` and `npm run preview`.

lxl-web/src/lib/components/find/LeadingPane.svelte
```diff
@@ -13,7 +13,11 @@
 	import BiArrowBarLeft from '~icons/bi/arrow-bar-left';
 	import IconSliders from '~icons/bi/sliders';
 
-	const { children } = $props();
+	interface Props {
+		children?: import('svelte').Snippet;
```
johanbissemattsson, line 17: Same as previous comment regarding importing snippet types.

johanbissemattsson (commented):
LGTM!

I cannot however reproduce the the error with `onload` violating our CSP policy. Both `onload` and `on:load` seems to work fine for me. 🤷

PR #1347, opened 2025-08-13
chore(lxlweb): revert rollup upgrade

jesperengstrom:
### Solves

Dependency upgrades in https://github.com/libris/lxlviewer/pull/1344 broke some of the supersearch functionality. It seems that the Rollup upgrade in `codemirror-lang-lxlquery` caused this and I have no idea why. 

Since it's only responsible for bundling the language, there's no urgent need to upgrade as long as it does its job...?

### Summary of changes

* Revert rollup upgrade

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,286 +63,6 @@
         "vitest": "^3.2.4"
       }
     },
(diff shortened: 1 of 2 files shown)


jesperengstrom:
Nooo... These test errors again. Time to go home 😄

PR #1350, opened 2025-08-15
feat(lxl-web): Scroll buttons on horizontal lists (LWS-402)

johanbissemattsson:
### Tickets involved
[LWS-402](https://kbse.atlassian.net/browse/LWS-402)

### Solves

Adds scroll buttons on horizontal lists indicating that the lists are scrollable.

Skärmavbild 2025-08-15 kl  14 42 40

The scroll buttons are only shown when the user has at least one accurate input pointing device (such as a mouse) and hovers over the lists.

The scroll buttons are therefor hidden on mobile devices where users are more accustomed to horizontally scrollable sections.

The new [::scroll-button](https://developer.mozilla.org/en-US/docs/Web/CSS/::scroll-button) CSS psuedo element could have been used to remove the need for JavaScript but sadly the support is still quite lacking ([64.34% at the time of writing according to caniuse.com](https://caniuse.com/mdn-css_selectors_scroll-button)).

For now, the scroll-buttons are instead hidden if Javascript is disabled.

### Summary of changes
- Add icon button utility class
- Add scroll buttons
- Use localized number formatting

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -285,6 +285,25 @@
 		}
 	}
 
+	.btn-icon {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		background-color: var(--color-page);
+		border-radius: calc(infinity * 1px);
+		border: 1px solid var(--color-neutral);
+		color: var(--color-body);
+		width: calc(var(--spacing) * 11);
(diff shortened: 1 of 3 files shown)


jesperengstrom (approved):
LGTM!

PR #1354, opened 2025-08-19
fix(lxl-web): Add editions to table of contents

johanbissemattsson:
### Solves

Add editions to table of contents on resource pages and fixes issue where each work instance got an observer (causing console warnings on e.g. https://beta.libris-qa.kb.se/9s2g6qpv7j59f46p).

### Summary of changes

- Add editions to table of contents
- Ensure only toc sections get intersection observers


jesperengstrom (approved):
LGTM!

Anchor links as generated in the toc seems to lead to some weird routing behavior (like i showed you), but that is a separate issue. Ship it!

PR #1361, opened 2025-08-29
feat(lxl-web): Navigation between results (LWS-221)

johanbissemattsson:
### Tickets involved
[LWS-221](https://kbse.atlassian.net/browse/LWS-221)

### Solves

Adds navigation between search result items on the resource page.

It works by passing along the search results data (with only the data which is needed) when clicking on a search result link on the `/find` page. This data is then used to calculate the previous/next result fnurgel and index of total results.

Subsequent searches are done if the user navigates to results which aren't part of the original search results. SvelteKit's newly added [remote functions](https://svelte.dev/docs/kit/remote-functions) are used here for the sake of simplicity; the feature is however still experimental but hopefully won't change too much in upcoming SvelteKit versions. 🤞 

### Summary of changes

- Pass along search result when navigating to resource page (using `page.state`)
- Add navigation between results on resource page
- Add `id` attributes to search result `<article>` elements. These ids will be used for anchor links to the specific resources when pressing _"Show in search results"_ (we first have to fix the issue where the [`<base>`-element](https://github.com/libris/lxlviewer/blob/develop/lxl-web/src/routes/(app)/%5B%5Blang%3Dlang%5D%5D/%2Blayout.svelte#L13) causes hashed urls to redirect to the start page).
- Add tests
- Add missing `@id` on search result data, and missing `previous` property on `PartialCollectionView` type.

### DIFF
--- a/lxl-web/package.json
+++ b/lxl-web/package.json
@@ -54,6 +54,7 @@
 		"typescript": "^5.9.2",
 		"typescript-eslint": "^8.39.1",
 		"unplugin-icons": "^22.2.0",
+		"valibot": "^1.1.0",
 		"vanilla-cookieconsent": "^3.1.0",
 		"vite": "^7.1.2",
 		"vitest": "^3.2.4"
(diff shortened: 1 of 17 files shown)


jesperengstrom (approved):
LGTM! Working really well as far as I can tell 🌟 

I wouldn't mind shrinking this just a tiny bit in height (`min-h-10` instead of `12`?) to allow more of the resource to fit above the fold, on mobile especially.

PR #1362, opened 2025-09-04
feat(lxl-web): Localized links using util (LWS-412)

johanbissemattsson:
### Tickets involved
[LWS-412](https://kbse.atlassian.net/browse/LWS-412)

### Solves

Localizes links using `localizeHref` util passed along with `page.data`. The util function is initialized with the current locale of the page but this can of course be overwritten using the `options.locale` param.

This also allows us to more easily allow locale switching without needing to do a full page reload (which caused `page.state` to be reset).

The old method of localizing links using a `<base>` element was unfortunately brittle and regressions in later versions of SvelteKit [caused issues where anchor links redirected to the start page](https://kbse.atlassian.net/browse/LWS-412). It was also easy to forget to remove the leading slash required for the relative links to work.

We now instead have to remember to add `localizeHref` to links but it should be more straightforward/explicit in it's functionality (and therefor easier for new developers to grasp). It also mimics the method we use for localizing strings today.

A newly added test should help with catching missing localized links. The routes which this test checks should preferably be generated programmatically but a  [constant](https://github.com/libris/lxlviewer/blob/633c9b65fcb7e6b93da5f2a45939b8fd4d546dc2/lxl-web/tests/locales.spec.ts#L4-L11)  will have to do for now...

### Summary of changes

- Add initialized `localizedHref` util to `page.data`
- Remove base element
- Skip removing opening slash from `relativizeUrl` (as base href element isn't used anymore)
- Add `trimSlashes` util for cases where needed (we cannot rely on relativizeUrl to strip slashes as leading slashes are now needed for relative links)
- Fix localization of SuperSearch component without having to do a full reload
- Remove trailing slash from `base` in `page.data`
- Add tests
- **Bonus:** Highlight resource after clicking _"Show in search results_" on resource page (see: #1361)

### DIFF
--- a/lxl-web/tests/find.spec.ts
+++ b/lxl-web/tests/find.spec.ts
@@ -57,11 +57,11 @@ test('expanded filters have no detectable a11y issues', async ({ page }) => {
 
 test('sorting the facet sets a cookie', async ({ page, context }) => {
 	const beforeCookies = await context.cookies();
-	expect(beforeCookies).toEqual([]);
+	await expect(beforeCookies).toEqual([]);
 	await page.getByTestId('facet-sort').nth(1).getByRole('combobox').selectOption('alpha.asc');
 	const afterCookies = await context.cookies();
-	expect(afterCookies[0].name).toEqual('userSettings');
-	expect(afterCookies[0].value).toEqual(
+	await expect(afterCookies[0].name).toEqual('userSettings');
+	await expect(afterCookies[0].value).toEqual(
 		'{%22leadingPane%22:{%22open%22:true}%2C%22facetSort%22:{%22rdf:type%22:%22alpha.asc%22}}'
 	);
 });


johanbissemattsson:
> Nice work! 🌟 Instant language change working nicely everywhere i've looked, _except_ `myLibraries` pill when it's not in the url, but passed along from a result list to another page. Switching language then results in:
> Skärmavbild 2025-09-15 kl 15 24 40
> 
> We also have to make sure that the `load` function responsible for creating cites runs again when switching language, since it fetches language-specific labels from the xl api (i.e `computedLabel=sv`) But I think we're saved by the fact that you can't switch language while the modal is open 😅

https://github.com/libris/lxlviewer/pull/1362/commits/790cf5589fcae3248d1e77f3b0d01ea738cd4ef3 should fix the issue (but is absolutely not an elegant solution...)

jesperengstrom:
Yep, this doesn't change the pill language in the case above, but the pill still remains. Which is fine by me!

PR #1365, opened 2025-09-09
Migrate to ESLint 9, (.eslintrc.js -> eslint.config.js)

lrosenstrom:
### Tickets involved

[LXL-4566](https://kbse.atlassian.net/browse/LXL-4566)

### Solves

ESLint has not been working in the libris cataloging project for  > a year.

### Summary of changes

* Migrate to the new "flat" ESLint config which is required since ESLint 9.xx. The official migration guide: https://eslint.org/docs/latest/use/migrate-to-9.0.0#flat-config.  

* Removed some of the old dependencies such as eslint-config-airbnb which is not yet supported in ESLint 9, and as I understand it, was more of "thing" back in 2015.

* It is no longer possible to exclude files/paths using `.gitignore` which was previously supported by using the `--ignore-path` flag, so we might manually want to add  some ignored paths in the new config file instead.

If we feel something is missing we can add it as we go along.


johanbissemattsson (approved):
LGTM! Great work! 🌟 

I would also add `eslint-plugin-cypress` to mitigate all cypress-related issues (e.g. `'expect' is not defined` and `'test' is not defined`).

It seems to be enough to add `import pluginCypress from 'eslint-plugin-cypress'` to eslint.config.js and then adding `pluginCypress.configs.recommended` to the extends array.

PR #1368, opened 2025-09-11
feat(lxl-web): Holdings pane incl improvements

jesperengstrom:
**Edit:**
To benefit from the changes below without having to go all the way implementing the holdings _panel_, I've added a feature flag called `USE_HOLDINGS_PANEL`. Set to `false`, everything should work like today, but a little faster.

---

## Description

Trying out displaying holdings in /find route in the resizable `TrailingPane`.
Also load the necessary data in a `load` function & stream it into the component.

**Features:**

* Quicker UI due to preloading + streaming the data = panel opens instantaneous.
* Getting rid of all `$effects` in the old `HoldingsModal`.
* `HoldingsModal` and `CitationsModal` components are removed in favour of a context-agnostic approach, i.e
```
<modal>
  <HoldingsContent />
</modal>
```
* No unnecessary fetching of data. Current implementation always calls the `/api/holding` endpoint TWICE on resource pages, even though most of the data is already available trough `page.data` (see screenshot). Everything needed is now fetched server-side and passed as props.
Skärmavbild 2025-09-10 kl  09 55 53

* `SearchCard` uses container queries and grid-template-areas (to look good independent of panel widths)

### DIFF
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
@@ -14,11 +14,10 @@ export const load = async ({ params, url, locals, fetch }) => {
 	const locale = getSupportedLocale(params?.lang);
 
 	const debug = locals.userSettings?.debug?.includes(DebugFlags.ES_SCORE) ? '&_debug=esScore' : '';
-	// const searchParams = new URLSearchParams(url.searchParams.toString());
 
 	const searchParams = new URLSearchParams();
 
-	// find page load function reloads on change in these params
+	// find page load function reloads on change in these params:
 	const reactiveParams = ['_q', '_limit', '_offset', '_sort', '_spell'];
 	reactiveParams.forEach((p) => {
 		searchParams.set(p, url.searchParams.get(p) || '');
--- a/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
+++ b/lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { page } from '$app/state';
 	import { afterNavigate, goto } from '$app/navigation';
-	import { SvelteURLSearchParams } from 'svelte/reactivity';
+	import { MediaQuery, SvelteURLSearchParams } from 'svelte/reactivity';
 	import type { SearchResult } from '$lib/types/search';
 	import type { HoldingsData } from '$lib/types/holdings';
 	import getPageTitle from '$lib/utils/getPageTitle';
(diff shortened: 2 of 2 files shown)


lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
```diff
@@ -13,12 +13,16 @@ export const load = async ({ params, url, locals, fetch }) => {
...
-	const searchParams = new URLSearchParams(url.searchParams.toString());
+	// const searchParams = new URLSearchParams(url.searchParams.toString());
+
+	const searchParams = new URLSearchParams();
+
+	// find page load function reloads on change in these params
+	const reactiveParams = ['_q', '_limit', '_offset', '_sort', '_spell'];
+	reactiveParams.forEach((p) => {
+		searchParams.set(p, url.searchParams.get(p) || '');
+	});
```
Comment (johanbissemattsson) [inline lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts:25]: I find this code a little strange, is it really needed?

Relying searchParams either from `new URLSearchParams(url.searchParams.toString());` or `const searchParams = new URLSearchParams([...Array.from(url.searchParams.entries())]);` usually is enough.

If the reasoning is to filter the search params sent to the api url, then I think it should be placed directly above the related fetch.

johanbissemattsson (commented):
Awesome work! 🌟

I almost lean towards saying –  _Yes let's go all the way and enable the panes directly!_ –  but I think we should ensure the history navigation works as intended first (preferably with saved scroll positions between navigations).

I however concur that it’s appropriate to merge this PR as-is as it also includes other worthy improvements/fixes that we could have use of directly (as the fix where duplicate unnecessary data fetches are removed).

For future implementations of the holdings content: we don’t necessarily need all _the_ holders data up-front, instead we could wait until the user expands the holders details element and _then_ fetch the specific data for that holder. Another solution could be using intersection observers to only fetch the holders data visible in the viewport. This should also fix the rate-limiting issues when running CI tests (side-stepped in https://github.com/libris/lxlviewer/pull/1374).

lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte
```diff
@@ -1,32 +1,71 @@
...
+	import { USE_HOLDING_PANE } from '$lib/constants/panels';
 
 	const searchResult: SearchResult = $derived(page.data.searchResult);
+	const holdings: Promise<HoldingsData> | undefined = $derived(page.data?.holdings);
+
+	let innerWidth = $state(null);
+	let isSmallScreen = $derived(innerWidth && innerWidth <= 640);
+	const HoldingsComponent = $derived(
+		isSmallScreen ? Modal : USE_HOLDING_PANE ? TrailingPane : Modal
+	);
```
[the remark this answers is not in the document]
Comment (jesperengstrom) [reply lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.svelte:29]: Thanks, will try it!

lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts
Comment (johanbissemattsson) [reply lxl-web/src/routes/(app)/[[lang=lang]]/find/+page.server.ts:25]: Oh then it totally makes sense! 👍 I learned something new now, thanks for that!

jesperengstrom:
> We could wait until the user expands the holders details element and _then_ fetch the specific data for that holder. 

I think that's hard to pull off with the current design since some of the data being fetched (blue links below) is shown up-front.

Skärmavbild 2025-09-18 kl  17 19 52. 


> Another solution could be using intersection observers to only fetch the holders data visible in the viewport. This should also fix the rate-limiting issues when running CI tests (side-stepped in #1374).

Yes, sounds like a good future improvement.

johanbissemattsson (approved):

PR #1369, opened 2025-09-15
Fix broken generation of label tail after • changed to ·

lrosenstrom:
### Tickets involved
[LG-59](https://kbse.atlassian.net/browse/LG-59)

### Solves

Sigel not visible in facets for libraries with long names/labels.

### Summary of changes

Match on · instead of • 

Screenshot from 2025-09-15 10-37-07


olovy (approved):
:ship:

PR #1374, opened 2025-09-17
chore(lxlweb): CI test fail debug

jesperengstrom:
Test if any of the following is causing ci test fails:
`fetchHoldersIfAbsent` - fetching holder data in a loop


johanbissemattsson (approved):
Good find! 🚢

PR #1376, opened 2025-09-24
feat(lxl-web): Improve search and menu styling

johanbissemattsson:
### Tickets involved

[LWS-375](https://kbse.atlassian.net/browse/LWS-375)

### Solves

Improves the styling and functionality of the header and its search input and menu, with focus on better responsiveness (fallbacks for when javascript is disabled are also added).

Tests for new functionality is unfortunately still missing but will added in a follow-up PR!

### Summary of changes

- Improve search
    - Add support for multiple rows when search is expanded
    - Add lazy loaded search component
    - Add fallback search input (shown when javascript is disabled and when lazy loaded search component is loading)
- Add new menu
    - Add `/browse` route (used instead of new menu when javascript is disabled)
    - Show expanded menu if hash is `#menu`
- Other
    - Add skip to main content functionality
    - Move beta banner to separate component (hiding banner on scroll is temporarily removed as it was difficult to achieve while also ensuring correct placement of dialog – a dismissable banner can however be enabled using a prop)
    - Add `aria-atomic` to search results info


olovy (commented):
Nice!

Haven't looked at the code.

My notes from a quick test:

Search component taking up the whole window on "desktop half screen" is a bit disorienting.
(since no "search button" like on mobile etc)
Compare with current version:
Skärmbild från 2025-09-24 16-22-53

Third row of search box is cut in half.
Skärmbild från 2025-09-24 16-26-19

Clicking the search icon on resource page breaks the interface completely because of uncaught exception:
Skärmbild från 2025-09-24 16-28-44

jesperengstrom (approved):
Great work! ⭐ 

LGTM, agree with the feedback and that it can be done in follow-up PR:s now that the critical stuff is fixed.

PR #1379, opened 2025-09-26
feat(lxl-web): Increase SuperSearch debounce; don't preload facet values

andersju:
Preloading facet values, i.e., triggering search on hover, seems wasteful and potentially harmful given the finite search server resources.

Also increased SuperSearch debounce time to prevent more unnecessary requests; 400 ms seems like a common, "lagom" number.


olovy (approved):
:ship:

olovy (commented):
(failing test)

PR #1381, opened 2025-09-30
Add copyright and phonogram symbol to templates and snippets

lrosenstrom:
## Description

For the copyright property: include pre-filled copyright and phonogram symbols together with date. Adding these symbols is enforced by current cataloging practice but see [LXL-4160](https://kbse.atlassian.net/browse/LXL-4160) for a discussion about this.

Question: 

- There's snippet for CopyrightNumber. How to handle that?
- I added the phonogram symbol to the following templates: `music-cd`, `audiobookStreaming` and `audiobookDisc`. Is this correct? @klngwll

### DIFF
--- a/cataloging/src/utils/data.js
+++ b/cataloging/src/utils/data.js
@@ -86,7 +86,9 @@ export function removeNullValues(inputObj) {
 }
 
 export function normalizeBeforeSave(inputObj) {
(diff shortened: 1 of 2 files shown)


olovy (approved):
🚢

PR #1383, opened 2025-10-03
feat(lxlweb): Subsets (LWS-414)

jesperengstrom:
### Tickets involved
[LWS-414](https://kbse.atlassian.net/browse/LWS-414)

### Solves

Basic support for viewing subsets in Libris.

1. "Entry point"

The following `.env` can now be added to for example https://arkdes.libris-dev.kb.se/
```
PUBLIC_SERVICE_NAME=ArkDes
APP_THEME=accent
```

This swaps out the header & page title and overwrites the `--brand-primary` colors as defined in themes.css
A very basic first step that can be fine-tuned and customized in later iterations...

**Backend TODO**: actual filtering in subdomain app settings...

2. "Subset filter"

Filter pills corresponding to the `_r` param are placed in the header (far from perfect design-wise, but let's iterate 🙃 ).
The `_r` param follows the user around everywhere on the app and affects searches ('normal' and related). For example, [this would be the Arkdes "start page"](http://localhost:5173/?_r=itemHeldBy%3A%22sigel%3AArkm%22)

A new instance filtering search (`_p` + `_o`) is performed on resource pages, showing instances matching the search ([example](http://localhost:5173/l3crh1zzjs8kwv8l?_r=itemHeldBy%3A%22sigel%3AArkm%22)). Multiple instances are now shown as  _instance cards_, meaning we now also have instance pages! 

In addition, `_q` is also passed on from search results to the resource page, and treated the same way. This makes for some cool/powerful searches:

[Searching by ISBN](http://localhost:5173/h08ndxppf96q96ff?_q=isbn%3A9129657490)
[Searching for a book published by...](http://localhost:5173/vcm1r997sfwf3bjr?_q=%22publication.agent%22%3A%22Norstedts%22)
[Looking for new editions](http://localhost:5173/9s2g6qpv7j59f46p?_q=yearPublished%3E%3D2010+yearPublished%3C%3D2025)

### TODOS:
* `_r` (and `_q`?) should affect holdings modal by promoting filtered libraries (big one, own ticket?)
* Instance pages (one of many) should have breadcrumb-like link back to work?

**Backend TODOS:**
* ~~`up` links for `_r` (will get x-button when this happens)~~
* `_r=alias-myLibraries` + `myLibraries=sigel,sigel` does not work
* ~~instanceType filtering does not work ([example](http://localhost:5173/h08ndxppf96q96ff?_q=hasInstanceType%3ADigitalResource))~~
* Create filter aliases for 'named' subsets (Suecia? DST?)

### Summary of changes

* Create `themes.css` and allow it to ovewrite tw theme variables depending on html `data-theme`
* Enable fetching search mappings in app `layout.server.ts` (to get `_r` pill labels on any page)
* Display subset in header and instance list using refactored `SearchMapping` component
* Show subset as search placeholder using new `displayMappingToString` function
* Append `_r` to all links by hijacking the `localizeHref` function
* On resource page: If multiple instances and url query, perform a filter search
* Refactor `TabList` component (WAI-ARIA-compliant keyboard nav etc) and use here
* Adjust `instance-web-cards`, modify and display `SearchCard` for instances
* Modify `HoldingsResourceCard` to accept card-formatted data

### DIFF
--- a/lxl-web/tests/subsets.spec.ts
+++ b/lxl-web/tests/subsets.spec.ts
@@ -1,0 +1,23 @@
+import { expect, test } from '@playwright/test';
+
+test('A subset filter can display on any page', async ({ page }) => {
+	await page.goto('/help?_r=itemHeldBy%3A"sigel%3AArkm"');
+	expect(page.locator('.subset-container').getByText('ArkDes')).toBeVisible();
+});
+
+test('The search placeholder is replaced when using a subset', async ({ page }) => {
+	await page.goto('my-pages?_r=itemHeldBy%3A"sigel%3AArkm"');
+	expect(page.getByRole('combobox', { name: 'Sök' }).getByText('ArkDes')).toBeVisible();
+});
+
+test('_r param is preserved when navigating around the app', async ({ page }) => {
+	await page.goto('/find?_q=f&_offset=0&_limit=20&_r=itemHeldBy%3A"sigel%3AArkm"');
+	await page.getByRole('main').getByRole('article').getByRole('link').first().click();
+	await expect(page).toHaveURL(/_r=itemHeldBy%3A%22sigel%3AArkm%22/);
+	await page.getByRole('main').getByRole('link').getByText('Nästa').click();
+	await expect(page).toHaveURL(/_r=itemHeldBy%3A%22sigel%3AArkm%22/);
+	await page.getByRole('main').getByRole('link').getByText('Visa i träfflista').click();
+	await expect(page).toHaveURL(/_r=itemHeldBy%3A%22sigel%3AArkm%22/);
+	await page.getByRole('link', { name: 'Sparat' }).click();
+	await expect(page).toHaveURL(/my-pages\?_r=itemHeldBy%3A%22sigel%3AArkm%22/);
+});


lxl-web/src/app.css
```diff
@@ -89,17 +103,17 @@
...
+	--color-primary-100: var(--theme-primary-100);
+	--color-primary-200: var(--theme-primary-200);
+	--color-primary-300: var(--theme-primary-300);
+	--color-primary-400: var(--theme-primary-400);
+	--color-primary-500: var(--theme-primary-500);
+	--color-primary-600: var(--theme-primary-600);
+	--color-primary-700: var(--theme-primary-700);
+	--color-primary-800: var(--theme-primary-800);
+	--color-primary-900: var(--theme-primary-900);
+	--color-primary-950: var(--theme-primary-950);
```
johanbissemattsson: Note to our future selves:

The implementation here works for now but it's a little unfortunate that we have to introduce new CSS variables instead of just changing the preexisting `--color-primary-*` variables depending on the theme, and that we have to pass along all theme values to all clients irrespective of the current entry point.

Dynamically imported CSS-files for a specific entry would be preferable (they can in turn import something similar to  `base.css` for base styling) as it would allow more advanced customizations. I did some quick experimentations trying to solve but it doesn't seem to be entirely trivial as Tailwind needs to do some precompiling.

lxl-web/src/lib/components/TabList.svelte
```diff
@@ -1,40 +1,118 @@
...
+				newIndex = 0;
+				break;
+			case 'End':
+				newIndex = tabButtons.length - 1;
+				break;
+			default:
+				return;
+		}
+
+		e.preventDefault();
```
johanbissemattsson: I would prefer having `e.preventDefault()` at the the top of `handleKeyDown` so it's directly apparent the default event actions should be skipped.

johanbissemattsson (approved):
Great work with this! 🌟 I think this will be a very good starting point which we can build upon when we have defined which specific cases for subsets we should start with.  

My main concern/reservation is that I personally feel that the `_r` param is a bit to "sticky"/omnipresent when _not_ part of a custom entry point (e.g. different domain/sub-domain).

I think it would be easier to grasp if the search filters would be reset when navigating away from `find`or `resource` pages (e.g. when clicking on the Libris logo). They can always navigate back if they feel that they accidentally removed the subset filter.

"Invisible" r-params which are applied from custom entry points  (e.g. different domain/sub-domain) can be treated differently as the users are not in the Libris context.

lxl-web/src/routes/+layout.server.ts
```diff
@@ -1,7 +1,36 @@
-export async function load({ locals }) {
+import { env } from '$env/dynamic/private';
+import { getTranslator } from '$lib/i18n/index.js';
+import { getSupportedLocale } from '$lib/i18n/locales.js';
+import { displayMappings } from '$lib/utils/search';
+
+export async function load({ locals, url, params }) {
```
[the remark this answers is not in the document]
jesperengstrom: Fixed in https://github.com/libris/lxlviewer/pull/1383/commits/a1beffc818daac9d8f634110effd0f3a71ceab95

jesperengstrom:
> We can keep it as-is for now but I'm inclined to prefer letting the logo/homepage link act as a full-blown reset for search queries (similar to how it's done on GitHub with repositaries). I'm not sure if the x-button is enough – best way to find out is with user-testing

Yep. I agree that the layout (specifically the / slash) may remind users of breadcrumbs, and that they expect them to behave as such. I'm ok with resetting from start page too, it's an easy fix anyhow.

(I didn't realize that GitHub uses the slash too, I must have mimicked it subconsciously)

lxl-web/src/lib/components/TabList.svelte
jesperengstrom: Sorry, had to revert this in https://github.com/libris/lxlviewer/pull/1383/commits/3ba826ca7ef4079e5f309f7a07aa2e617d3c3f39 Since the early preventDefault is blocking the onclick action...

PR #1385, opened 2025-10-06
Wait 1 second before loading search hitlist after removing a record

lrosenstrom:
## Checklist:

### Tickets involved
[LXL-4689](https://kbse.atlassian.net/browse/LXL-4689)

## Solves

When navigating to a holding directly from the search hitlist, then removing it, the user is navigated back to the hitlist. In the search card of the record for which the holding was `itemOf`, the button for creating a new holding was not available (instead this button stated, incorrectly, that the user still had a holding for this record).

### Summary of changes

Add 1000ms timeout.

### DIFF
--- a/cataloging/src/views/Inspector.vue
+++ b/cataloging/src/views/Inspector.vue
@@ -396,8 +396,12 @@ export default {
           type: 'success',
           message: `${labelByLang(this.recordType)} ${StringUtil.getUiPhraseByLang('was deleted', this.user.settings.language, this.resources.i18n)}!`,
         });
-        // Force reload
-        this.$router.go(-1);
+
+        // Wait so that the change is available in the index when the search hitlist is loaded
+        setTimeout(() => {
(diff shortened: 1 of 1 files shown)


cataloging/src/views/Inspector.vue
```diff
@@ -396,8 +396,12 @@ export default {
           type: 'success',
           message: `${labelByLang(this.recordType)} ${StringUtil.getUiPhraseByLang('was deleted', this.user.settings.language, this.resources.i18n)}!`,
         });
-        // Force reload
-        this.$router.go(-1);
+
+        // Wait so that the change is available in the index when the search hitlist is loaded
+        setTimeout(() => {
+          this.$router.go(-1);
+        }, 1000);
```
olovy, line 403: ES does index refresh every 1000 ms. Let's do 1100 ms.

olovy (approved):
🚢

PR #1514, opened 2026-03-04
feat(lxlweb): Free online in search result card

jesperengstrom:
### Solves

If the item is available online, display a card action that is the first link, also display a popover with all the available links
Skärmavbild 2026-03-04 kl  15 11 41

Also fixes the bug with duplicate properties on some resource pages by not `copyMediaLinksToWork` unless work.
Example [before](https://beta.libris-qa.kb.se/sdzfjsz5q7vd72gt) [after](http://localhost:5173/sdzfjsz5q7vd72gt)
### Summary of changes
* Add `mediaLinks` prop to `SearchResultItem`
* Extract `copyMediaLinksToWork` util to own file, then use it to get the links and format them.
* Enable the `popover` component to also accept a `snippet` to render as content. Render the links using `DecoratedData`.
* Add prop to optionally turn of `focus` and `blur` listeners for the popover (default is on). This is because this popover is useless with keyboard navigation unless you can tab to the links. We could make it accessible, but it would then mean you need to tab past all the links when navigating in the result list. This would probably come across as annoying. The links can still be accessed through the resource page.
* Fix duplicate display of `associatedMedia` prop (+ more duplicates) by only using `copyMediaLinksToWork` for a _work_ resource.

### DIFF
--- a/lxl-web/src/lib/actions/popover/Popover.svelte
+++ b/lxl-web/src/lib/actions/popover/Popover.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { onMount } from 'svelte';
+	import { onMount, type Snippet } from 'svelte';
 	import { computePosition, offset, shift, inline, flip, arrow } from '@floating-ui/dom';
 	import SearchCard from '$lib/components/find/SearchCard.svelte';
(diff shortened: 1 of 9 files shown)


olovy (approved):
Very nice! 👌 

LGTM!

PR #1515, opened 2026-03-05
fix(lxlweb): Use web-token for search mapping

jesperengstrom:
### Solves

How about we use our new `web-token` to get rid of person lifespan in pills?
Ever since we hid the qualifier key, birthdate etc seems (even more) superfluous.
Also created a new library token omitting `sigel`.

Skärmavbild 2026-03-05 kl  09 55 33

### Summary of changes
* Format `display` and `displayStr` in search mapping using `web-tokens`
* Add `Library-web-tokens`


johanbissemattsson (approved):
**Y E S!** Good idea!

olovy (approved):

PR #1516, opened 2026-03-05
feat(lxl-web): Add qualifier keys from filters help page

johanbissemattsson:
### Solves

Adds support for adding qualifier keys from the `/help/filters` page.

Currently there is no smartness regarding inserted whitespace; it inserts an extra whitespace before the qualifier key as a safety precaution (but could also lead to repeated whitespaces if there already is one present).

### Summary of changes

- Expose function to change supersearch query using context
- Use changeQuery function to add qualifier keys
- Add test


olovy (approved):
LGTM! 🚢

PR #1518, opened 2026-03-06
feat(lxl-web): Use different aria-labels on expanded supersearch

johanbissemattsson:
### Solves

Use different aria-labels on expanded supersearch as _"Search all Swedish libraries"_ was very repetitive when using screen readers (_"Search"_ is used instead).

`npm run prebuild-local-packages` is required.

### Summary of changes
- Use different aria-labels on expanded search
- Set aria-label props from `AppSearch` component instead of `AppBar`


jesperengstrom (approved):
LGTM!

PR #1519, opened 2026-03-06
fix(supersearch): set estimated height on pill widget

jesperengstrom:
### Solves

Would like some help to see if this fixes the Firefox "cursor jumping down" bug in supersearch pills. 
I've rarely seen it myself, and not at all on this branch...

`npm run prebuild-local-packages`  

### Summary of changes

* Set widget `estimatedHeight`


johanbissemattsson (commented):
Sadly doesn't seem to help here :/

Image

johanbissemattsson (approved):
Lets try this together with set height on `lxl-ghost-group`!

`estimatedHelp` should still be helpful. 🚢

PR #1521, opened 2026-03-09
feat(lxlweb): icons/thumbs in pills

jesperengstrom:
### Solves

Just trying out display of type icon / thumb images (for Persons) in pills
Skärmavbild 2026-03-10 kl  16 35 13

### Summary of changes

* `type` and `id` are appended to props passed from mapping -> validator function -> `QualifierPill`
* Use type with `TypeIcon` component
* If `id` && `Person`, get and display the thumb using a `getPersonImage` util:
   * `getPersonImage` fetches a resource once, then caches the image url, returning it for subsequent calls. This is because QualifierPill re-mounts all the time (and it has to, because the removeLink needs to recalculate as you type) and we can't fetch the resource again on every keystroke.

### DIFF
--- a/lxl-web/src/lib/components/supersearch/QualifierPill.svelte
+++ b/lxl-web/src/lib/components/supersearch/QualifierPill.svelte
@@ -2,12 +2,13 @@
 	import { page } from '$app/state';
 	import type { QualifierRendererProps } from 'supersearch';
 	import IconClose from '~icons/bi/x-lg';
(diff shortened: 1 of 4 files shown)


olovy:
Yes!! Beautiful :nail_care: 

This will maybe make us want to hide labels for specifically subject with type Subject

olovy (approved):

PR #1522, opened 2026-03-09
fix(lxl-web, supersearch): Fix supersearch cursor issue

johanbissemattsson:
### Solves

Fixes issue with text cursor not being able to escape qualifier values on Firefox, while using the latest versions of Codemirror (https://github.com/libris/lxlviewer/pull/1523 was a temporary fix which required a downgraded version of `@codemirror/view`). 

This is now done by replacing ghostgroup parens with a inline whitespace (instead of an `inline-block` element).

### Summary of changes

- Replace ghostgroup parens with whitespaces and use inline elements instead of inline-blocks (which [doesn't seem to be supported by the library](https://github.com/codemirror/dev/issues/1594#issuecomment-3095456994))
- Proper highlighting inside qualifiers
- Prevent unnecessary contentattributes effect triggers
- Bump codemirror packages and use exact version as of now
- `repairGhostGroup` seems to be redundant with the additions made in `handleChangesInGhostGroup`so it is removed.

### DIFF
--- a/lxl-web/src/lib/components/find/SearchMapping.svelte
+++ b/lxl-web/src/lib/components/find/SearchMapping.svelte
@@ -42,8 +42,8 @@
 								type={m.display?.[JsonLd.TYPE]}
 								id={m.display[JsonLd.ID]}
 								{isRedundantKeyLabel}
-							/>
-						</span>
+							/>{' '}</span
+						>
 					{:else}
 						<!-- free text search -->
 						<span class="flex h-full items-center gap-1 pl-1.5">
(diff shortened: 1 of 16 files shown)


jesperengstrom:
Got some errors when trying to build this...

Needs env
`PUBLIC_DEBUG_GHOST_GROUP `
in supersearch root? (devops addition needed?)

Then:
`"handleInputBeforeGroup" is not exported by "src/lib/extensions/lxlQualifierPlugin/ghostGroup.ts", imported by "src/lib/extensions/lxlQualifierPlugin/index.ts"`

jesperengstrom:
And of this fixes the codemirror/view breaking change, we should maybe bump it all the way to 6.40.0?

johanbissemattsson:
Please try again @jesperengstrom , the latest commits should remove the need for `PUBLIC_DEBUG_GHOST_GROUP`.
The codemirror packages are now also bumped.

I have to fix the merge conflicts and update the tests but it should otherwise be ready..

jesperengstrom:
Nice work! Seems to work pretty much like before + nice improvement with the visible selection. 
I detected a regression where a lone paren is visible if you select and delete the start of a group. That case was covered previously.

 ghost

johanbissemattsson:
> I detected a regression where a lone paren is visible if you select and delete the start of a group. That case was covered previously.

Good find! This should now be fixed in https://github.com/libris/lxlviewer/pull/1522/commits/e6646bfb363ec61d93dfee73ae021bae05c4cf2d

PR #1525, opened 2026-03-12
fix(lxl-web/resource): Don't display empty details section

olovy:
### Solves
On resource page: Remove Details section when there are no details to display.

Example:
https://beta.libris.kb.se/g0bsnzc9d3kb2d25


jesperengstrom:
LGTM. My only remark is that the permalinks etc look even more displaced now, especially on small screens. But that that can be fixed later.

Skärmavbild 2026-03-12 kl  14 37 36

jesperengstrom (approved):

PR #1526, opened 2026-03-12
fix(lxlweb): fix mylibraries indicator for instances / too many btns on narrow card

jesperengstrom:
### Solves

* Make `getHoldingsByType` handle more diverse data to fix the issue where myLibraries indicator not working correctly for instance cards.

Skärmavbild 2026-03-12 kl  14 27 46
VS
Skärmavbild 2026-03-12 kl  14 27 54

* Add min-width to library button to achieve button alignment in search result
Skärmavbild 2026-03-12 kl  14 31 00

* Re-organize `SearchCard` grid to allow card-actions (buttons) take up a row below the image on small screens. Prevents the case of three buttons looking horrible on small screens
Skärmavbild 2026-03-12 kl  12 45 29
Skärmavbild 2026-03-12 kl  12 45 23

* Use container queries for pagination component to prevent page overflow when pane is resized

### DIFF
--- a/lxl-web/src/lib/components/find/Pagination.svelte
+++ b/lxl-web/src/lib/components/find/Pagination.svelte
@@ -82,7 +82,7 @@
 	<nav
 		aria-label={page.data.t('search.pagination')}
 		data-testid="pagination"
-		class="pagination mt-4 py-4"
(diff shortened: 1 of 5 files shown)


olovy (approved):
LGTM! 🚢

PR #1527, opened 2026-03-12
feat(lxl-web/filters): Display/hide filter label based on vocab

olovy:
Depends on https://github.com/libris/definitions/pull/566

## Description
Hide labels for filters only where the filter is obvious from the value.
For now this is just a simple `:category` on the filter definition. (naming TBD!)
That is, the label for specific property/filter is always visible or not when it has a linked value.
Never show "language", always show "orginal language"

I started sketching a solution where this was dependent on the type of the value. 
e.g something like `:subject :impliedByValueOfType :Topic`
But I'm not sure that is necessary. I think `subject` is the one filter that is on the fence here.

Example:
Skärmbild från 2026-03-12 13-22-34

Also here:
bild
Skärmbild från 2026-03-12 16-52-00
Skärmbild från 2026-03-12 16-52-41

### DIFF
--- a/lxl-web/src/lib/components/find/SearchMapping.svelte
+++ b/lxl-web/src/lib/components/find/SearchMapping.svelte
@@ -24,7 +24,17 @@
 >
 	{#each mapping as m, i (`outer-${i}-${depth}`)}
 		{#if !MAPPING_IGNORE_VARIABLE.some((v) => v === m.variable)}
-			{@const { children, operator, up, variable, displayStr, label, display, invalid } = m}
+			{@const {
+				children,
+				operator,
+				up,
+				variable,
+				displayStr,
+				label,
+				display,
+				isRedundantKeyLabel,
+				invalid
+			} = m}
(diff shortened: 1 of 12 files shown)


jesperengstrom:
Great! such a nice feature ⭐
Moving towards perfect pills now 💊 . Would really like to see this in combination with the current pr:s, ie [tokens](https://github.com/libris/lxlviewer/pull/1515), [hide colons](https://github.com/libris/lxlviewer/pull/1517), [icons](https://github.com/libris/lxlviewer/pull/1521) (does it for instance make sense to display the label _and_ the icon?)

A bit unsure about the change to skip labels for the string version. Without the visual help, filters and free text could be impossible to distinguish. Don't know if that's a problem in practice, so let's try it out.

Skärmavbild 2026-03-12 kl  17 07 24
Skärmavbild 2026-03-12 kl  17 07 33

jesperengstrom (approved):

PR #1528, opened 2026-03-13
fix(lxlweb): fix IOS Safari input zoom

jesperengstrom:
### Solves

IOS Safari zooms in when putting the cursor in an input field in a very disturbing way. Apparently, this is an "accessibility feature" - when Safari thinks the font size in an input is too small, it zooms it in for better visibility. "Too small" being below 16px.

I think the correct fix here is not to hack around it but to actually increase the font-size, it looks better and seems standard.

adding `text-base sm:text-sm` for all our inputs. Tested it on real mobile using dev:host and it works.

Also some other minor fixes, mostly increased sizes that looks too small in comparison with the new input size.

### Summary of changes

* input `text-sm` -> `text-base`

### DIFF
--- a/lxl-web/src/lib/components/Holder.svelte
+++ b/lxl-web/src/lib/components/Holder.svelte
@@ -92,7 +92,7 @@
 				</a>
 			{:else if numInstances > 1 && hasSomeItemLink}
 				<button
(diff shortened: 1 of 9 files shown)


olovy (approved):
LGTM!

PR #1529, opened 2026-03-13
fix(lxlweb): Use QualifierPill for SearchMapping

jesperengstrom:
### Solves

* I was about to fix so that the `SearchMapping` component renders icons etc just like the supersearch pills. Then realized we should just render the `QualifierPill` component directly. Less code - one source of truth.
Skärmavbild 2026-03-13 kl  16 11 17

* Also tweaked the pills a bit (re-inserted the divider line for example)

* Filter out `_r` filters (finally) from the modal filter display

* Plus a fix for the automatically inserted wildcard, so that we don't insert it after operator. 

### Summary of changes

* Use `QualifierPill` in `SearchMapping`
* add `NO_WILDCARD_AFTER_WORD`

### DIFF
--- a/lxl-web/src/lib/components/find/SearchMapping.svelte
+++ b/lxl-web/src/lib/components/find/SearchMapping.svelte
@@ -30,10 +30,10 @@
 			{@const { children, operator, up, variable, displayStr, label, isRedundantKeyLabel } = m}
 			{#if displayStr || label}
 				<li class={['lxl-qualifier inline-block', variable && `variable-${variable}`]}>
-					{#if m._key}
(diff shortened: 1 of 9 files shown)


jesperengstrom:
The tests luckily alerted that some `_r` pill functionality broke. Will fix it

olovy:
Looks great!

Two things I found:

Trashcan if more than one `_r`
Skärmbild från 2026-03-13 17-45-28

Can't remove the last `_r` with the X button
Skärmbild från 2026-03-13 17-43-45

olovy (approved):
LGTM!

PR #1531, opened 2026-03-17
feat(lxl-web): Add search tips

lrosenstrom:
Add a search tips section. 

I think it's good to keep this kind of info separate as it might be considered (at least semi) advanced. Plus, we can add more search tips & tricks!

~(i) -> ⓘ, 💡???~


olovy (approved):
🚢

PR #1532, opened 2026-03-19
feat(lxlweb): Library subset tweaks

jesperengstrom:
### Solves

A few improvements when using a library as a subset filter, in these examples [ArkDes](http://localhost:5173/?_r=itemHeldByOrg%3A%22sigel%3Aorg%2FARKM%22).

The biggest change under the hood is the way to fetch subset mappings. Earlier, they were fetched in the root layout file and passed as page data to the client. Now we need access to the subset mappings in other server `load` functions (and we don't want to rely on `await parent()`). The new solution is to fetch the mappings in the `hooks.server` file and add it to `event.locals`. The hooks file is delicate and needs to be performant, so i've implemented a util that caches `_r` with their mappings to avoid any unnecessary fetches. This data should be safe to share between users on the server, but feel free to express any concerns with this approach.

Visual changes:

1) Front page now takes `_r` into account when fetching previews. Preview sections without any items are now hidden.
Skärmavbild 2026-03-19 kl  12 47 46

2) When using a library as a subset filter, the text of the holdings button changes.
Skärmavbild 2026-03-19 kl  12 50 52

Same thing with instances on resource pages (instances not held by subset display the same as before)
Skärmavbild 2026-03-19 kl  12 53 33

3) "Holdings by type" buttons keep their CTA look only if subset has holdings, others get a more neutral look (we could also hide them completely, but that would mean you can't access the modal if somehow entering a resource without "the right" holding).
Skärmavbild 2026-03-19 kl  12 56 22

4) Fixed regression in navbar + subset pill. Now fade + sidescroll if pill reaches its maximal width.
Skärmavbild 2026-03-19 kl  13 09 09

### Summary of changes
* Add `subsetCache.server` + update root `+layout.server`, `hooks.server.ts`
* Add `heldBySubset` to `SearchResultItem`
* rename `getMyLibsFromHoldings` -> `getLibsFromHoldings`, `getHeldByMyLibraries` -> `getHeldByLibraries`
* modified `SearchCard` and `ResourceHoldings` with `conditional` text
* Account for `_r` in `searchResult.remote` (start page fetch)
* Update `AppBar` CSS
* Add a few tests to cover cases in this PR

### DIFF
--- a/lxl-web/src/hooks.server.ts
+++ b/lxl-web/src/hooks.server.ts
@@ -8,6 +8,7 @@ import { DebugFlags, type MyLibrariesType, type UserSettings } from '$lib/types/
 import displayWeb from '$lib/assets/json/display-web.json';
 import { DisplayUtil, VocabUtil } from '$lib/utils/xl';
 import { startRefreshLibraries } from '$lib/utils/getLibraries.server';
+import { getSubsetMapping } from '$lib/utils/subsetCache.server';
 
 type Util = [VocabUtil, DisplayUtil];
 let utilCache: Util | undefined;
@@ -86,6 +87,11 @@ export const handle = async ({ event, resolve }) => {
 
 	// set data-theme defined in themes.css
 	const dataTheme = site?.themeName || 'libris';
+
+	// get subset mapping
+	const _r = event.url.searchParams.get('_r');
+	const subsetMapping = await getSubsetMapping(_r, event.locals, lang);
+	event.locals.subsetMapping = subsetMapping;
 
 	return resolve(event, {
 		transformPageChunk: ({ html }) => html.replace('%lang%', lang).replace('%theme%', dataTheme)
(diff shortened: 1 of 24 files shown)


olovy (commented):
Yes!

Longer subset labels break the layout on small screens.
Other than that, LGTM

Image

Image

Image

Image

Image

jesperengstrom:
Whoops, hopefully fixed in https://github.com/libris/lxlviewer/pull/1532/commits/1340e66b2c55eeb5f7171071739ac2aeb3f7ab1c

Impossible to make long pills look good on the front page with the current design, so I hid them for now. We should probably display some other text than "Sök på alla svenska bibliotek" instead?

frontpage_pill

johanbissemattsson (approved):
LGTM! 🙌 🚢 

I think it would be good to truncate the label in the subset pill so the remove button always is visible, but that can come later.

PR #1536, opened 2026-03-24
fix(lxl-web, supersearch): Fix issue with qualifier pills not appearing due to nbsp chars

johanbissemattsson:
### Solves

A (not-so-elegant) fix to the issue with qualifier pills not appearing because of strange `nbsp` chars.

I'm not sure where the non-breaking spaces come from; I have tried searching through the entire front-end codebase for nbsp characters without any luck.

### Summary of changes

- Add extenstion to replace all nbsp chars with spaces


johanbissemattsson:
This should not be merged as I'm working on a better solution.

PR #1537, opened 2026-03-24
fix(lxlweb): HTML validation warnings/errors

jesperengstrom:
### Solves

Ran some pages through the [W3C markup validation](https://validator.w3.org/) and tried to fix the stuff it reported. I think this needs to be done in iterations...

### Summary of changes
* Remove self-closing of void elements, like `<link>`, `<meta>`, `<input>`, `` (not required, but recommended)
* Fix duplicate id `”search”`
* Remove manual `selected` for `<select>` when it is also bound (resulting in duplicate `selected=""`)
* Remove invalid empty form `action`
* The `form` attribute of the sort-select did not refer to a form element. Add `search-form` to both.
* `<h2>` is not allowed as a child of `<label>`. Added sr-only `<h2>` outside of the toc-mobile toggle to preserve page semantics.
* Remove `aria-expanded` for input element:  "Attribute aria-expanded not allowed on element [input] at this point"
* Prevent `<span>` from getting `target=”_blank”` in `decoratedData`

### DIFF
--- a/lxl-web/maintenance.html
+++ b/lxl-web/maintenance.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<meta charset="utf-8" />
+<meta charset="utf-8">
 <title>Underhållsarbete</title>
 <style>
 	body {
--- a/lxl-web/src/app.html
+++ b/lxl-web/src/app.html
@@ -1,11 +1,11 @@
 <!doctype html>
 <html lang="%lang%" data-theme="%theme%">
 	<head>
-		<meta charset="utf-8" />
-		<link rel="icon" href="%sveltekit.assets%/favicon.svg" type="image/svg+xml" />
-		<link rel="icon" href="%sveltekit.assets%/favicon.png" type="image/png" />
-		<link rel="apple-touch-icon" sizes="180x180" href="%sveltekit.assets%/apple-touch-icon.png" />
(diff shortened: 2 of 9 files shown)


johanbissemattsson (changes requested):
LGTM! (except for below)

> Remove aria-expanded for input element:  "Attribute aria-expanded not allowed on element [input] at this point"

The WAI-ARIA combobox has [only one required attribute ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role#associated_wai-aria_roles_states_and_properties)which happens to be `aria-expanded` – and I think WAI-ARIA currently outweighs the W3C markup spec due to our focus on accessibility.

jesperengstrom:
> The WAI-ARIA combobox has [only one required attribute ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role#associated_wai-aria_roles_states_and_properties)which happens to be aria-expanded – and I think WAI-ARIA currently outweighs the W3C markup spec due to our focus on accessibility.

Alright! Reverted in https://github.com/libris/lxlviewer/pull/1537/commits/e49feaf28befe24e9504ab7cbe472ce227f64155

johanbissemattsson (approved):

PR #1539, opened 2026-03-27
fix(lxl-web): resource summary and toc width

olovy:
Summary and table of contents on resource page 
* max width. `60ch` seems to give ~70 characters 
* left justify

TODO
Split unstructured MARC style tocs with `--` to rows


johanbissemattsson (approved):
MUCH better (more readable)! 🚢 ✨

PR #1540, opened 2026-03-30
feat(lxweb): stream stats

jesperengstrom:
### Solves
With `_stats=falseThisRequest` we can omit stats from the response without any trace in the output links. That means we can fetch the **search results** and the **facets** separately and let the facets stream in when done.

I chose to put the stats fetch in the `+layout.server` file because we want to run it as seldom as possible. I.e when paginating or sorting, the stats are the same (and by not having the layout load function dependent on those, we can reuse the data instead).

Some additional features:
* We cache the old facet data in the component to display (if available) while waiting. That means the "disabled" look lasts from navigation start until we have new data. If no cached data, we display a "skeleton" instead. 
* To ensure users with javascript disabled can still access the filters, we await the data when the request is not coming from server side navigation (`isDataRequest`).

All and all, the site will hopefully feel a lot faster. Compare it to the beta :)

### Summary of changes

Summary of changes
* add `getFacets` to /find `+layout.server`
* Refactor `Filters` component
* Add '_stats', 'falseThisRequest' to some other fetches where stats are not needed.

### DIFF
--- a/lxl-web/src/app.css
+++ b/lxl-web/src/app.css
@@ -468,7 +468,7 @@
 }
 
 @utility skeleton {
-	@apply rounded-lg bg-neutral-100;
+	@apply animate-pulse rounded-lg bg-neutral-100;
 }
 
 /* OLD THEME */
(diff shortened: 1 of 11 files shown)


olovy (approved):
YES! This is it!

🌬️ 🍂

johanbissemattsson (approved):
LGTM! 🙌 Awesome work!

PR #1542, opened 2026-03-31
fix(lxlweb): fix hydration mismatch

jesperengstrom:
### Solves

We get a hydration mismatch warning on every page
Skärmavbild 2026-03-31 kl  11 49 46

The culprit seems to be `--page-y-offset={pageYOffset}`
Removing it fixes the problem. 
I can't really detect the value it's supposed to set. Removing it makes no visible difference to me.
Maybe it has some function that i missed?

### Summary of changes

* remove `--page-y-offset={pageYOffset}`


johanbissemattsson (approved):
Good find!

`--page-y-offset` is used to correctly position the expanded search dialog on the start page when scrolling down – but the `AppBar` component shouldn't need to be aware of it. I guess it's a rest product of earlier implementations.

Removing it from `SuperSearchWrapper` would however break the positioning. Supersearchwrapper is lazy loaded so there should be no need to worry about hydration mismatches there.

PR #1544, opened 2026-04-09
fix(lxlweb): fix zotero tag

jesperengstrom:
### Solves

Page data from resource pages must have changed along the way with `resourceId` replacing `uri` (that now includes `#it`) silently breaking the zotero api call.

However, we should probably use `instance.@id`s instead, allowing all the instances to be listed for works (this was the case in the past, can't remember how that worked with the old code)

Anyway, also wrote a couple of tests so this can never happen again 😉 

### Summary of changes

* Fix `<abbr>` tag
* Write tests


olovy (approved):
LGTM!

PR #1545, opened 2026-04-09
feat(lxlweb): more item details pt 1 

jesperengstrom:
### Solves

Add more data under three labels in the holdings panel; _placering_, _bestånd_, _relaterad länk_ (in line with old Libris)

Old Libris uses `fresnel:mergeProperties` to squash all relevant properties under a new namespace. We don't support that (yet), so the hack around that is to recreate these as `Item` lenses for our custom lens groups. i.e:

`Item-web-overview` -> _relaterad länk_
`Item-web-overview2` -> _placering_
`web-overview-footer` -> _bestånd_

example _placering_ & _bestånd_ ([link](http://localhost:5173/l3wgvwhx1q80cv2?holdings=PhysicalResource)):
Before:
Skärmavbild 2026-04-09 kl  16 55 04
Now:
Skärmavbild 2026-04-09 kl  16 55 54
Old:
Skärmavbild 2026-04-09 kl  16 56 45

----
example _relaterad länk_ (`associatedMedia` & `isPrimaryTopicOf`)

Now ([link](http://localhost:5173/m5z9j7xz1d01cbv?holdings=DigitalResource)):
Skärmavbild 2026-04-13 kl  15 26 31

Old:
Skärmavbild 2026-04-09 kl  17 00 24

Now ([link](http://localhost:5173/cvn9gpgp123cj65?holdings=PhysicalResource)):
Skärmavbild 2026-04-13 kl  15 30 17

Old:
Skärmavbild 2026-04-13 kl  15 29 49

----

Of course the old site thought of everything, and I found a case that I can't solve (there are maybe more).
The property `copyNumber` gets some kind of special treatment, but I can't even find a mention in the display file.

([link](http://localhost:5173/l3wl4jtx4xg6p25?holdings=PhysicalResource))

Skärmavbild 2026-04-13 kl  15 36 11

Skärmavbild 2026-04-09 kl  17 08 17

----

**TODO**: part 2 which is to list all item `subject`, `classification` etc among resource details.

### Summary of changes

Summary of changes

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -338,11 +338,23 @@
 						}
 					]
 				},
+				"Item": {
+					"@id": "Item-chips",
+					"@type": "fresnel:Lens",
+					"classLensDomain": "Item",
+					"showProperties": ["heldBy", "itemOf"]
+				},
 				"Library": {
 					"@id": "Library-chips",
 					"@type": "fresnel:Lens",
 					"showProperties": ["name", "qualifier", "sigel"],
 					"classLensDomain": "Library"
+				},
+				"ItemAvailability": {
+					"@id": "ItemAvailability-chips",
+					"@type": "fresnel:Lens",
+					"showProperties": ["label"],
+					"classLensDomain": "ItemAvailability"
 				},
 				"Bibliography": {
 					"@id": "Bibliography-chips",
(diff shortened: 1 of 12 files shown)


olovy (approved):
Nice, LGTM! 📑  📚 

Got an empty Placering in some places. But I think this is good to merge.

http://localhost:5173/n12l1fw4lrhbqgh4
https://libris-qa.kb.se/katalogisering/8mp8lr5s62g3shcx
Image

jesperengstrom:
> Got an empty Placering in some places

Yeah, it's because we display `hasComponent` but it contains only an `Item` that we don't display. 
In combination with the dumb condition for the label: `shelfData._display.length`

In this case it's more nested until finally resolving into nothing (and this could go on for any number of levels)

```
shelfData: {
  _display: [{
    hasComponent: {
      _display: []
    }
  }]
}
```

I'm not sure about a quick fix, but I would **reaallly** like to fix this at the root, i.e when the displayDecorated baking machine hits a dead end, it goes back and erases the track. This would mean less data sent and also maybe getting rid of stray separators `, , , ` that we print because the node _might_ contain data deep down.

olovy:
> I'm not sure about a quick fix, but I would **reaallly** like to fix this at the root, i.e when the displayDecorated baking machine hits a dead end, it goes back and erases the track.

Agreed! That is the correct approach.

PR #1546, opened 2026-04-13
fix(supersearch, lxl-web): Cancel debounced fetching timeouts

johanbissemattsson:
### Solves

The debounced fetching in the supersearch component had a tendency to show the results of unfinished fetch requests started _before_ resetting data – this fix ensures that the timeouts are cancelled/cleared when calling `resetData`.

(The problem is more apparent when the expanded search open is kept open (coming in a upcoming PR)).

### Summary of changes

- Clear debounced timeouts when resetting data
- Skip intentional `AbortError` messages


olovy (approved):
LGTM

PR #1547, opened 2026-04-13
feat(lxl-web): Download MARC-XML

olovy:
`/_compilemarc` can now take a `format` that overrides the format in the export profile. Use it.


jesperengstrom (approved):
LGTM!

PR #1548, opened 2026-04-13
perf(lxl-web): Compute qualiferSuggestions once

olovy:
Compute qualiferSuggestions once instead of on every request.
They are static in relation to vocab.

took ~14% cpu when profiling locally with stress_search2_api_simple.py

- Move `getQualifierSuggestions()` to separate file
- Call once in hooks.server.ts `loadUtilCached()`

Before:
Skärmbild från 2026-04-13 16-41-46

### DIFF
--- a/lxl-web/src/lib/utils/getQualifierSuggestions.ts
+++ b/lxl-web/src/lib/utils/getQualifierSuggestions.ts
@@ -34,9 +34,9 @@ export function getQualifierSuggestions(
 
 function mapSearchFilterDefinition(
 	def: FramedData,
-	locale,
+	locale: LocaleCode,
 	display: DisplayUtil,
-	compare
+	compare: (x: string, y: string) => number
 ): QualifierSuggestion2 | null {
 	try {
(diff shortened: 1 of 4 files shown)


lxl-web/src/lib/utils/getQualifierSuggestions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// TODO where do we specify this?
+const CURATED_QUALIFIERS = [
+	'contributor',
+	'title',
+	'language',
+	'yearPublished',
+	'subject',
+	'originalLanguage'
+];
```
johanbissemattsson, line 71: I think this works for now! (or that we are more consistent in placing constants in `src/lib/constants`, which I could do more often)

johanbissemattsson (approved):
Looks good to me! 🙌 I also like that getQualifierSuggestions was moved to a separate file. 👍

jesperengstrom (approved):
Nice! ⭐

PR #1549, opened 2026-04-13
perf(lxl-web/xl.ts): Memoize getVocabLabel()

olovy:
It is static in the context of vocab.

search results/facets before, bottom-up
Skärmbild från 2026-04-13 17-05-41

### DIFF
--- a/lxl-web/src/lib/utils/xl.ts
+++ b/lxl-web/src/lib/utils/xl.ts
@@ -39,10 +39,20 @@ export class VocabUtil {
 	//vocabId: string
 	vocabIndex: Map;
 	context;
+	labelCache: Record<LangCode, Map>;
 
 	constructor(vocab: VocabData, context: ContextData) {
 		this.context = lxljsVocab.preprocessContext(context)[JsonLd.CONTEXT];
 		this.vocabIndex = lxljsVocab.preprocessVocab(vocab);
+		this.labelCache = {};
+	}
+
+	getLabelCache(locale: LangCode) {
+		if (this.labelCache[locale] == undefined) {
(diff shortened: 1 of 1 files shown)


lxl-web/src/lib/utils/xl.ts
```diff
@@ -730,12 +740,19 @@ class Formatter {
...
+		}
+
 		try {
-			return toLabel(
+			const label = toLabel(
 				this.displayDecorate(
 					this.displayUtil.applyLensOrdered(this.vocabUtil.getDefinition(vocabName), LensType.None)
 				)
 			);
+			cache.set(vocabName, label);
```
johanbissemattsson: Should we save it to the cache even if the label doesn't exist? Quite alot of repeated `[] => undefined` are currently present in the cache. Or is it worthwhile for the performance improvements?

Image

johanbissemattsson (commented):
🌟 Looks good but see the comment about saving undefined labels.

The label cache could also be be useful for the client to achieve instant pills when adding qualifiers (but that can be the scope of another PR).

lxl-web/src/lib/utils/xl.ts
olovy: Good catch!
Fixed in [1412bfe](https://github.com/libris/lxlviewer/pull/1549/commits/1412bfec8893af2adbc8ca85f6699e4d7abc436f)

olovy: > Should we save it to the cache even if the label doesn't exist?

Yes we don't want to recompute it. The number of different values for the `vocabName` argument is finite and quite small.

johanbissemattsson (approved):

PR #1550, opened 2026-04-13
perf(lxl-web/xl.ts): Memoize find...Format()

olovy:
`getFacets` + `asResult`: ~45% -> ~37% CPU


johanbissemattsson (approved):
🚢 
I was a little confused with the differences between `formatIndex` and `formatCache` – the latter memoises (never know how to spell it) the function parameters (which is however implied by the PR title...).

Maybe `formatCache` -> `formatMemo` ?

PR #1551, opened 2026-04-14
perf(lxl-web/xl.ts): Build base lenses array once

olovy:
Used in a hot path DEFAULT_SUBLENS_SELECTOR() -> isDerivedLens() 

Shaves ~10% CPU


johanbissemattsson (approved):
🚀 LGTM!

PR #1552, opened 2026-04-14
fix(lxlweb): avoid squeezed pills

jesperengstrom:
### Solves

The only left-right padding of an unlinked pill value is now the (hidden) parenthesis. But sometimes the generated queries don't include them, resulting in 
Skärmavbild 2026-04-14 kl  12 02 56

How about we add padding when the value does not contain `lxl-ghost-paren`?

Skärmavbild 2026-04-14 kl  12 02 47

### DIFF
--- a/lxl-web/src/lib/styles/lxlquery.css
+++ b/lxl-web/src/lib/styles/lxlquery.css
@@ -51,6 +51,10 @@
 	background-color: var(--color-accent-50);
 }
 
+.lxl-qualifier-value:not(:has(.lxl-ghost-paren)) {
+	padding-inline: calc(var(--spacing) * 1.5);
+}
+
 .atomic .lxl-qualifier-value {
 	border-radius: 0;
 }


johanbissemattsson (approved):
LGTM! I'm thinking we could also use:

```
	&::before {
		content: ' ';
	}
	&::after {
		content: ' ';
	}
```
...to ensure that the spacing is exactly the same as the whitespaces replacing parentheses.

But neither implementation would show highlighting when selected. 🤔 

Image

jesperengstrom:
> ...to ensure that the spacing is exactly the same as the whitespaces replacing parentheses.

But those chars don't take up the same width. Resulting in a jumping effect when you type & the parens are added.

I guess the perfect thing would be

```
	&::before {
		content: '(';
		color: transparent;
	}
	&::after {
		content: ')';
		color: transparent;
	}
```

But that has a nasty side effect, you now can't position the cursor at immediately before the value, that place is "taken" by the `::before` element somehow...

johanbissemattsson:
Ah thats true, probably best to merge it as-is then! 🚢 🌟

PR #1556, opened 2026-04-15
fix(lxlweb): resolve() causing broken links

jesperengstrom:
### Solves

Both these:

Skärmavbild 2026-04-15 kl  09 40 50
Skärmavbild 2026-04-15 kl  09 40 15

Is because `resolve()` cant handle links with uri:s in their params. I've encountered it before...

### Summary of changes

* Remove `resolve()`


johanbissemattsson (approved):
LGTM!

PR #1558, opened 2026-04-15
feat(supersearch): Add wrapping left/right navigation

johanbissemattsson:
### Solves

Adds horizontal wrapping when the supersearch component using left/right arrow keys (if `wrappingArrowKeyNavigation` is enabled).

This is useful for further improvements of the expanded search in `lxl-web` (running `npm run prebuild-local-dependencies` is then required).

### Summary of changes

- Add wrapping left/right arrow key navigation
- Fix wrapping down arrow key navigation on last row
- Add test for left/right wrapping arrow key navigation


jesperengstrom (approved):
Nice!

PR #1562, opened 2026-04-16
fix(cataloging,idkbse,lxl-web): add Anubis cookies to cookie consent info + fix language bug

andersju:
Adds info about Anubis cookies to the cookie consent modals.

Also fixes the bug where if you used the language switcher on Libris sök, the cookie consent modal language wouldn't change.


johanbissemattsson (approved):
LGTM! 🙌 Good find with the language bug!

PR #1563, opened 2026-04-16
fix(lxlweb): avoid pointless truncation

jesperengstrom:
### Solves

The automatic truncation is great when it's helping:

Skärmavbild 2026-04-16 kl  16 34 22

But confusing when it's not helping:

Skärmavbild 2026-04-16 kl  16 35 55

The user might think "maybe i got zero hits because of that weird char I didn't even type" and try removing it and it can't be done 🙈 

This _redirects back_ to the original query when the truncation was pointless. I thought the double redirect would take too much time but turns out it doesn't. This feels like much better UX.

Skärmavbild 2026-04-16 kl  16 39 07

### Summary of changes

* Redirect back when `_relaxed` and still 0 hits


johanbissemattsson (approved):
Nice! 🌟 LGTM!

PR #1564, opened 2026-04-20
feat(lxl-web): Prevent lost value and selection when loading supersearch

johanbissemattsson:
### Solves

When loading the site on slow connections (e.g. on trains) users could start editing the fallback input before the lazy-loaded supersearch component has finished loading. The contents added (as well as the selection) would then be reset when mounting the lazy-loaded component which can be frustrating. This PR solves this by passing along the current value and selection.

Further improvements can later be made where so the selection is placed directly after pills when selecting qualifiers before mount (as this requires parsing the syntax grammer).

The PR also further implements context usage for controlling the component from outside instead of using component binding.

### Summary of changes

- Use searchContext to control SuperSearch from outside
- Move fallback search input to separate component
- Pass along current value and selection when mounting SuperSearchWrapper
- Use same focus styling on collapsed and expanded state

### DIFF
--- a/lxl-web/src/lib/components/supersearch/SuperSearchFallback.svelte
+++ b/lxl-web/src/lib/components/supersearch/SuperSearchFallback.svelte
@@ -1,0 +1,96 @@
+<script lang="ts">
+	import { page } from '$app/state';
+	import IconSearch from '~icons/bi/search';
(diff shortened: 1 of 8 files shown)


jesperengstrom (approved):
Nice! ⭐

PR #1565, opened 2026-04-20
feat(lxl-web): Add keyboard shortcut help for expanding search

johanbissemattsson:
### Solves

Add keyboard shortcut help for expanding search.

Skärmavbild 2026-04-20 kl  11 23 43

### Summary of changes

- Add keyboard shortcut help for expanding search

### DIFF
--- a/lxl-web/src/lib/components/supersearch/SuperSearchWrapper.svelte
+++ b/lxl-web/src/lib/components/supersearch/SuperSearchWrapper.svelte
@@ -524,6 +524,19 @@
 					>
 						<IconClear aria-hidden="true" class="size-4.5 sm:size-4" />
 					</svelte:element>
+				{:else if !expanded}
+					<button
+						type="button"
+						onclick={() => showExpandedSearch()}
+						tabindex={-1}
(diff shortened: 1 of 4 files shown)


jesperengstrom (approved):
Working as intended.
But I think we should maybe ask users about this too - to what extent it's helpful or comprehensible. Must be careful not to overload the search area.
* Novice users would need to hover this to understand what it's doing, somewhat contradictory if you want to use keyboard only?
* Pro users would just start typing the name of a filter they like to add, no need to simply expand the search?

But i'm totally fine with trying this out.

olovy (commented):
LGTM

I think the text should be e.g. "Gå till sökruta" or similar.
The point is selecting the search field without clicking, nor that it is expanded?

johanbissemattsson:
> LGTM
> 
> I think the text should be e.g. "Gå till sökruta" or similar. The point is selecting the search field without clicking, nor that it is expanded?

Fixed in https://github.com/libris/lxlviewer/pull/1565/commits/8b0ecd239640677647136c07a88ec01206f429f6

PR #1566, opened 2026-04-20
feat(lxl-web): Holdings button in main column

olovy:
- Move holdings button in main content area on resource page
- Make category regular links instead of pills. So that they clash less with all buttons visually
    - Maybe the difference between subject and category is more pedagogical now that they're side by side and labelled.

Skärmbild från 2026-04-20 13-16-06

Different button widths on mobile does not look great but ok for now?
bild


johanbissemattsson (approved):
I like! LGTM!

PR #1567, opened 2026-04-20
feat(lxlweb): add item information

jesperengstrom:
### Solves

Add an `Item` lens for `web-details` to display as "biblioteksspecifika detaljer" on instance pages:

Skärmavbild 2026-04-20 kl  15 37 12. 

* Displayed per holder instead of per prop as previously. Ok?
* Mapped as `itemInformation` to align with filter "Bestånd", even though "Bestånd" _also_ includes `itemShelf` properties _not_ displayed here. Too confusing?

I also went out on a limb and vibe coded a function to cleanup the decorated data. The aim was to get rid of empty nodes mainly caused by `hasComponent` (that might have stuff we want to display but usually not):

Skärmavbild 2026-04-20 kl  16 12 28

Basically we have no way of knowing if a branch in `displayDecorated` ends up as a dead end:

```
{
  "@id": "https://libris-qa.kb.se/xx",
  "@type": "Item",
  "_display": [
    {
      "hasComponent": {
        "@type": "Item",
        "_display": [
          {
            "something": {
              "_display": [
                {
                  "couldGoOnForever": {
                    "_display": [] // nothing here
                  }
                }
              ]
            }
          }
        ],
        "_style": [
          "ext-link"
        ],
        "_label": "Bestånd"
      },
      "_label": "har komponent"
    }
  ],
  "_style": [
    "ext-link"
  ],
  "_label": "Bestånd"
}
```

It we squash it like 👇 we can decisively skip the entire section

```
{
  "@id": "https://libris-qa.kb.se/xx",
  "@type": "Item",
  "_display": [],
  "_style": [
    "ext-link"
  ],
  "_label": "Bestånd"
}
```

Cleaning is now done as the last step of `lensAndFormat` and seems to be working fine.
However:

* Hope it does not have a negative performance impact
* Best option would be to not even construct these nodes in the first place

### Summary of changes

* Add `item-web-details` lens, format and pass as resource `itemInformation`
* Add `<details>` section in `Resource` component
* Create and use `/cleanupDecorated.ts`
* Rename `shelfData` -> `itemShelf` (= Placering)
* Rename `itemNoteData` -> `itemNote` (= Beståndsuppgift)
* Move `immediateAcquisition` from shelf to information.


olovy:
Ok!

Some initial testing:

immediateAcquisition is not working? 
http://localhost:5173/wf7hxsw705qtktz
https://libris-qa.kb.se/katalogisering/8dmfj7hf2nvl39t
```
:immediateAcquisition [ a :ImmediateAcquisition ;
      marc:methodOfAcquisition "Gåva;" ;
      marc:sourceOfAcquisition "Måltidsakademiens biblioteksstiftelse" ] ;
```
bild

-----------------------

hasComponent should be handled in some way. Render it the same way as editions in the holdings panel? i.e. grouped with a vertical bar. flatten when 1 component?

should show the props inside hasComponent, not hasComponent:
bild

bild

-----------------------

I think `marc:hasTextualHoldingsBasicBibliographicUnit` (Oformaterad beståndsuppgift - Huvudpublikation) should only be shown in the holding panel? Not in details.

jesperengstrom:
> I think marc:hasTextualHoldingsBasicBibliographicUnit (Oformaterad beståndsuppgift - Huvudpublikation) should only be shown in the holding panel? Not in details.

ok, fixed in https://github.com/libris/lxlviewer/pull/1567/commits/32aeb6935611ee9f3fef58b8f257b0caae5e72d3
`hasNote` is in both shelf (panel) and in details. Maybe same thing there, only in panel?

jesperengstrom:
removing `hasNote` would also fix the `hasComponent` issue in your example. Don't know if there are others...

kwahlin:
Does it make more sense to move `hasNote` to Beståndsuppgift? I guess the note can sometimes be about the location (Placering) but not always and if we move it to Beståndsuppgift it will be displayed alongside Placering anyway.

I also don't think `cataloguersNote` should be displayed (we don't show it in webbsök?) since its for internal stuff such as the cataloguers signature.

jesperengstrom:
> Does it make more sense to move hasNote to Beståndsuppgift? I guess the note can sometimes be about the location (Placering) but not always and if we move it to Beståndsuppgift it will be displayed alongside Placering anyway.

> I also don't think cataloguersNote should be displayed (we don't show it in webbsök?) since its for internal stuff such as the cataloguers signature.

Fixed in https://github.com/libris/lxlviewer/pull/1567/commits/90f257974282c3b92c484b6ac98dad9539d6ef8f

jesperengstrom:
Now flattening the items nested in `hasComponent` and looping them for each label. + Added `copyNumber`. Result:

Skärmavbild 2026-04-27 kl  17 44 05

for http://localhost:5173/vc510p46000mmsz

olovy:
Yes! :ship: 

The loop should be changed so that the info for each component is kept together. 
But let's do that in a follow up PR.

See for example:
http://localhost:5173/k3ww38kw00mbmps?_q=itemInformation%3A%28exemplar%C2%A0a%29
Skärmbild från 2026-04-27 17-51-17

jesperengstrom:
> The loop should be changed so that the info for each component is kept together.
But let's do that in a follow up PR.

Doh... Will do it now. I never want to revisit this again.

jesperengstrom:
> The loop should be changed so that the info for each component is kept together.

fixed in https://github.com/libris/lxlviewer/pull/1567/commits/cccfd1e60bde655eab78795fd47eaaf0783e61d2

lxl-web/src/lib/utils/holdings.server.ts
```diff
@@ -79,11 +80,19 @@ export function getHoldingsByInstanceId(
 
 		const items = instance[JsonLd.REVERSE]?.itemOf ?? [];
 		result[id] = items.map((item) => {
+			const _item = { ...item };
+			const component = _item?.hasComponent;
+			delete _item.hasComponent;
+			const allItems = component?.length ? [_item, ...component] : [_item];
```
olovy:
```suggestion
			const allItems = component?.length ? [...component, _item] : [_item];
```

I suggest switching the order of item and components (like it was before)

It is quite common for placering and beståndsuppgift to be split over the main item and a single component.
Switching them makes it more consistent IMO.

Without this change:
1:
Image
2:
Image

With this change:
1:
Image
2:
Image

olovy (approved):
Great! LGTM!

I have a single suggestion. See comment.

lxl-web/src/lib/utils/holdings.server.ts
jesperengstrom: fixed in https://github.com/libris/lxlviewer/pull/1567/commits/8145ff6a4230dd64d227f1892966afed84c88d16

PR #1568, opened 2026-04-21
fix(lxlweb): restore subset look

jesperengstrom:
### Solves

Skärmavbild 2026-04-21 kl  15 14 23

Does not look great at some scroll positions, but hard to fix due fixed positions etc on start page.

### Summary of changes

* Re-add subset placeholder text
* Unhide subset pills on start page


olovy (approved):
LGTM!

PR #1571, opened 2026-04-24
fix(lxlweb): hopefully fix 'return key label is context-aware' test

jesperengstrom:
### Solves

Let's see if this fixes the infamous test

### Summary of changes

* Move `await` in 'return key label is context-aware' test


johanbissemattsson (approved):
🙏 🙏 🙏  Thanks!!! Looks good to me!

PR #1573, opened 2026-04-27
feat(lxl-web/search): Split filters in three levels

olovy:
Three levels of filters:
- default suggested
- in "show more"
- only in reference

"Reference" instead of "List"


johanbissemattsson (approved):
LGTM! 🙌 

I personally think that the pipe separator is redundant and would prefer having the pills on one row (with horizontal scroll if needed).

Reference could also maybe be moved to the footer _or_ the same row as the label "Lägg till filter"?

👆 These things can be addressed later, I think it should be shipped as is! 🚢

PR #1575, opened 2026-04-28
feat(lxl-web/help): Some more "search in" in filter reference

olovy:
Handle propertyChainAxiom with restriction on range.
Code is a bit tedious. We might want to do all formatting with `displayUtil` in the future.

restriction on range with type
bild

restriction on range with anonymous subclass (`owl:Restriction`)
`<translator>` for now since we don't have that resource available without fetching it (not part of base vocab)
bild

### TODO
Other types of filters
* composite
* merged property
* ...


jesperengstrom (approved):
LGTM!

PR #1576, opened 2026-04-28
Use preformatted _q string from API response for relation search

kwahlin:
Fixes bad query formatting in relation search.

Before:
bild

After:
bild

Depends on https://github.com/libris/librisxl/pull/1756

The code can likely be be simplified even more now that backend provides all the required data. Feel free to step in :slightly_smiling_face:


olovy (approved):
LGTM

olovy (commented):
Check tests

PR #1577, opened 2026-04-28
feat(lxlweb): support markdown for bibliography description

jesperengstrom:
### Solves

* Added `"fresnel:valueStyle": ["markdown()"]` for `Bibliography.description`
* `markdown()` is a styler that takes a md string transforms it to html using packages `marked` and `sanitize-html` (do `npm install`)
* It outputs it in a new `_html` prop. I.e: `'* some text' => _html: '<ul><li>some text</li></ul>'`
* When `DecoratedData` sees _html, it renders it as `{@html data}`

[example](http://localhost:5174/0000n8zk12kmp71s)

---

The bulk of the diff here is renamed/split files.

`sanatize-html` is a node-only module, and it makes sense to make the transformation serverside. However, it did not work to make the util `.server` only, because `xl.ts` runs on the client too. Why does it run on the client? 🤔 There's really no need.

Apart from some simple utils, it only handles serverside data. 
So i split it into `xl.server.ts` and `xl.ts`. `search.ts` now only use `xl.server` stuff, so it can also be made `search.server` etc. By applying the same thing to `auxd.ts`, we get rid of the 'crypto' warning. 

Maybe it's a bit bold to make these changes just before the release. No rush to merge this.

### Summary of changes

* `markdown()` styler
* Add marked and sanatize packages and format
* split util files


olovy:
Nice! :raised_hands: 

> because xl.ts runs on the client too. Why does it run on the client? 🤔 There's really no need.

It looks like `toString()` is only used 
- to build a key for the `{#each}` [here in FacetGroup.svelte](https://github.com/libris/lxlviewer/blob/b05b6f8d5797f4a7dd99d65559a83f0c8e715c0f/lxl-web/src/lib/components/find/FacetGroup.svelte#L157) ? 
- and `getMetaDescription.ts` ?

Utils like `asArray` / `first` / `isObject` can be moved out if they are useful. There is a point of keeping `xl.ts` somewhat self contained though. If we want to move it to / replace `lxljs` at some point.

----------------------------------

Would be nice if links inside markdown were  "external" (_blank and arrow icon)

jesperengstrom:
> Would be nice if links inside markdown were "external" (_blank and arrow icon)

fixed in https://github.com/libris/lxlviewer/pull/1577/commits/50c8eb9301acb508b65035aaed565400b7d269c9

johanbissemattsson (approved):
LGTM! 🌟

olovy (changes requested):
I think we should remove client side use of xl.ts instead of splitting it

jesperengstrom:
> I think we should remove client side use of xl.ts instead of splitting it

The remaining (`toString`, `asArray`, `isObject`) now renamed as `misc.ts` 🤷‍♂️

olovy (approved):
LGTM!

PR #1579, opened 2026-04-29
feat(lxlweb): Add new banner

jesperengstrom:
### Solves

Adds a new banner informing about the new site + link to old
Skärmavbild 2026-04-29 kl  16 03 38

* `dismissedNewBanner` is now a UserSetting entry, storing dismissal in a cookie
* if dismissed, a `.dismissed-banner` class is set on `.app-bar`, setting the `--banner-height` css variable to 0. This is needed because of fixed positioning, the app bar position is calculated based on banner height etc.
* Added a menu entry for the old site, in case someone needs to find the link after dismissal.

Styling can be discussed of course :)

### Summary of changes

* Add `AppBanner` component
* remove `BetaBanner` component
* add `dismissedNewBanner` user setting 
* Update menu


olovy (approved):
LGTM!

Maybe have the link to the old site in the footer as well?

Would be nice to have the link by env e.g. gamla.libris-qa.kb.se
but not worth it

olovy (approved):

PR #1581, opened 2026-04-30
feat(lxlweb): featured categories component

jesperengstrom:
### Solves
I've extracted the stuff we said we want to remove from the homepage search context, and put it in a component of its own.
This gives us the possibility to still show it on the front page on a place of our choosing.

Skärmavbild 2026-04-30 kl  10 04 10.

I think:
* the text is very clarifying and should be kept
* The categories add 'information scent'
* They work well together
* No need to throw this away

### Summary of changes

* create `FeaturedCategories.svelte`

### DIFF
--- a/lxl-web/src/lib/i18n/locales/en.js
+++ b/lxl-web/src/lib/i18n/locales/en.js
@@ -141,6 +141,7 @@ export default {
 		search: 'Search',
 		searchResults: 'Search results',
 		searchFor: 'Search for',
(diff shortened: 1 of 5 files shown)


olovy (approved):
Agree with everything!

On mobile atm but screenshot and code looks good.

jesperengstrom:
👍 Don't merge this until the duplicate content is actually gone

johanbissemattsson (approved):
LGTM!

PR #1583, opened 2026-05-04
feat(lxl-web): Handle legacy setorg parameter

olovy:
For example https://libris.kb.se?setorg=ARKM

Uses `itemHeldByOrg` instead of `library`. When using `library` sigel has higer priority that org code. Which would lead to "Sök inom ArkDes Bibliotek" instead of "Sök inom ArkDes". Maybe that is what we want? Let's sort that out and adjust accordingly later.


jesperengstrom (approved):
LGTM!

PR #1584, opened 2026-05-04
fix(lxl-web): Hide explore link for empty collections

olovy:
Hide "Visa samlingens innehåll" link for empty collections.
Define "empty" as ten or less. For example OSGO contains a single record

This is mostly a workaround for search API not including `@reverse.isPartOf` for bibliographies.


jesperengstrom (approved):
Nice!

PR #1594, opened 2026-05-05
feat(lxl-web): Move dismissed banner to separate session cookie

johanbissemattsson:
### Solves
Moves the dismissed banner flag to a [separate session](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#removal_defining_the_lifetime_of_a_cookie) cookie (which is deleted when the "current session" ends).

I think this can be preferred initially but can later easily be changed back to a custom max-age, separate from the max-age of userSettings.

### Summary of changes

- Move dismissed banner flag to separate session cookie
- Add description to cookie consent

### DIFF
--- a/lxl-web/src/app.d.ts
+++ b/lxl-web/src/app.d.ts
@@ -17,6 +17,7 @@ declare global {
 			vocab: VocabUtil;
 			display: DisplayUtil;
 			userSettings: UserSettings;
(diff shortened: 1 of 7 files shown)


jesperengstrom (approved):
👍 working as intended

PR #1600, opened 2026-05-07
feat(lxlweb): instant key labels

jesperengstrom:
### Solves

When adding a filter (from suggestions) or typing a qualifier key, we don't need to wait for suggestion mappings, we have all the querycodes available. Just find & show the label.

### Summary of changes

Summary of changes
* pass `qualifierSuggestions` to `getLabelsFromMapping` 
* Search it if only needing a key label


olovy (approved):
Awesome!

(simple fix in failing test needed)

johanbissemattsson (approved):
I love it! 🙌  LGTM!

PR #1602, opened 2026-05-08
feat(lxl-web): Update footer

olovy:
- New footer section "Shortcuts"
- Add link to Libris Fjärrlån
- Update link to Libris kundservice
- Try to make app menu a little bit more consistent with footer
    - Other stuff in submenu under "About Libris" is still a bit weird?

### DIFF
--- a/lxl-web/src/lib/components/AppMenuContent.svelte
+++ b/lxl-web/src/lib/components/AppMenuContent.svelte
@@ -78,16 +78,16 @@
 			currentPage: page.route.id === '/(app)/[[lang=lang]]/about',
 			children: [
 				{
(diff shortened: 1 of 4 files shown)


andersju (approved):
LGTM but agree with "Other stuff in submenu under "About Libris" is still a bit weird?" -- I'd just place them as regular menu items on the same level as everything else. It's not overly much and the important things (Home, Search, My pages, Special collections) stand out anyway since they have icons.

olovy:
> LGTM but agree with "Other stuff in submenu under "About Libris" is still a bit weird?" -- I'd just place them as regular menu items on the same level as everything else. It's not overly much and the important things (Home, Search, My pages, Special collections) stand out anyway since they have icons.

done.
Also added "help" to footer

PR #1603, opened 2026-05-08
feat(lxlweb): handle deleted records (410) properly

andersju:
Deleted records e.g. https://libris.kb.se/k41knrq6hmptfrpf were not handled properly. The server returned a HTTP 200 while showing a generic "Something went wrong" page with the text
```
410
Something went wrong
Gone
```
So the 410 was received from backend but not propagated.

With this fix we return the actual status code and show a custom 410 message.
```
 curl -I "http://localhost:5173/k41knrq6hmptfrpf"
HTTP/1.1 410 Gone
```
and
```
410
Record has been deleted

This record no longer exists in Libris.

[Back to the home page](http://localhost:5173/en)
```

### DIFF
--- a/lxl-web/src/lib/components/Error.svelte
+++ b/lxl-web/src/lib/components/Error.svelte
@@ -8,6 +8,9 @@
 		if (page.status === 404) {
 			return getPageTitle(page.data.t('errors.notFound'), page.data.siteName);
 		}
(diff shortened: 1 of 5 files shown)


olovy (approved):
📑 🗄️ LGTM!

PR #1605, opened 2026-05-08
fix(lxlweb): square brackets in query causing page to break

jesperengstrom:
### Solves

a query like `https://libris.kb.se/find?_q=[hej]` causes the page to break
Mostly a theoretical issue, same query inputted from the search field gets encoded.

Problem is `resolve()` uses square brackets as template for slugs, see https://svelte.dev/docs/kit/$app-paths#resolve
so we need to re-encode the search params before using it.

### Summary of changes

* Calculate more safe `otherLangLink` 
* Add test


johanbissemattsson (approved):
LGTM! 🙌

PR #1609, opened 2026-05-12
fix(lxlweb): select type string first in suggestions

jesperengstrom:
### Solves

Skärmavbild 2026-05-12 kl  14 57 25

### Summary of changes

* move element


johanbissemattsson (approved):
I think this is a big improvement when searching! 👍

PR #1613, opened 2026-05-19
fix(lxlweb): fallback labels for facets

jesperengstrom:
### Solves

Adds a fallback for facet labels. When they don't  have an english label, `toLite` returns []. If we want to keep it that way we must perform a length check and then try the default locale.

But the real bug here is `dimension`. It used `str` to create a 'path' for the user setting. But `str` is also the label; when it's missing it created an unmatchable user setting, resulting in the facet not being able to expand. It also created two different settings for swedish and english for the same thing:

`"librissearch:findCategory/Moving image/librissearch:identifyCategory":"OPEN",
"librissearch:findCategory/Rörlig bild/librissearch:identifyCategory":"OPEN"}}`

### Summary of changes

* update `mapSlices`

### DIFF
--- a/lxl-web/src/lib/utils/search.server.ts
+++ b/lxl-web/src/lib/utils/search.server.ts
@@ -489,7 +489,7 @@ function mapSlices(
 							locale,
 							translate,
 							undefined,
-							dimension + '/' + str
+							dimension + '/' + o.object[JsonLd.ID]
 						)
 					}),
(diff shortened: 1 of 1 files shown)


jesperengstrom:
Replaced by https://github.com/libris/lxlviewer/pull/1619

Still think we should add the `dimension` change:

`dimension + '/' + o.object[JsonLd.ID]`

PR #1614, opened 2026-05-20
fix(lxlweb): try/catch on json parse libris session cookie

jesperengstrom:
### Solves

Add try/catch when parsing the `LIBRIS_SESSION` (fjärrlån) cookie. To prevent app from breaking when tampering with the cookie, since it is httpOnly.

### Summary of changes

* update `hooks.server.ts`


johanbissemattsson (approved):
LGTM! 🌟

PR #1615, opened 2026-05-20
feat(lxlweb): show error ID if available

andersju:
With https://github.com/libris/librisxl/pull/1770 we now get `error_id` from the backend, so show it if available.

Before (try it with `/find?_q=doi%3Afoo`):
```
500
Something went wrong

Invalid property chain axiom for shorthand property 'doi'.
```
(internal, should not be seen by users, annoying to correlate with logs)

After:
```
500
Something went wrong

Please try reloading the page. If the problem persists, report the error ID below.

Error ID: 4ac902be-ad56-43da-9df8-343600d6d977
```
(easily searchable in logs)

### DIFF
--- a/lxl-web/src/app.d.ts
+++ b/lxl-web/src/app.d.ts
@@ -12,6 +12,7 @@ declare global {
 	namespace App {
 		interface Error {
 			status?: string;
+			errorId?: string;
 		}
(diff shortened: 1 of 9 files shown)


olovy (approved):
🚢

PR #1616, opened 2026-05-20
fix(lxlweb): remove unused files

jesperengstrom:
### Solves

Remove some unused files:
* `/api/resources/server.ts` - this was an alternate(?) way of loading app resources that was never put in place.
* `i18n.json` - an alternate file of translations, only used by ☝️  
*  `displayGroups.json` - only used by ☝️ 
* `relationsByBaseClass.json` - was used, no longer imported anywhere
* `prefixesByNameSpace.json` - same

### Summary of changes

* Remove files


johanbissemattsson (approved):
Sorry for the late review, but it looks good to me! 🚢

PR #1620, opened 2026-05-26
feat(lxl-web): set favourite libraries via link

olovy:
Make it possible to set favorite libraries via query parameter.
This is an alternative / addition to setting library/org with `_r `. Can search the whole catalog but highlight some libraries.
same as `e/getPermaLink.jsp` in webbsök.

example
http://localhost:5173/?favouriteLibraries=org/BIN,Skfb
### Summary of changes

Summary of changes
* handle settings params in hooks.server.ts.
* remove obsolete label from `userSettings.myLibraries`. It was never read.
* Add settings links section to "my pages"
* fix: searching for favourite libraries in my-pages cleared _r parameter

### DIFF
--- a/lxl-web/src/hooks.server.ts
+++ b/lxl-web/src/hooks.server.ts
@@ -59,12 +59,6 @@ export const handle = async ({ event, resolve }) => {
 	let cookieEdit = false;
 	let redirectHome = false;
 
-	if (event.url.searchParams.has(SettingsParams.wipeSettings)) {
-		userSettings = {};
-		cookieEdit = true;
-		redirectHome = true;
-	}
-
 	if (event.url.searchParams.has('_debug')) {
 		let flags = event.url.searchParams
 			.getAll('_debug')
--- a/lxl-web/src/hooks.server.ts
+++ b/lxl-web/src/hooks.server.ts
@@ -88,7 +88,7 @@ export const handle = async ({ event, resolve }) => {
 			.map((s) => s.trim())
 			.map((s) => LIBRARY_URI_PREFIX + s)
 			.filter((id) => getLibrary(id) || getOrgMembers(id).length > 0)
-			.forEach((l) => (myLibraries[l] = '')); // label is obsolete
+			.forEach((l) => (myLibraries[l] = ''));
 		userSettings.myLibraries = myLibraries;
 
 		cookieEdit = true;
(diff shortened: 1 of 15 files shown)


jesperengstrom:
Nice! Working as intended. I think the tricky thing here is copy. We need to make sure the purpose is clear and that this is "special need" only.

Maybe it the link should not be visible when no fav libraries?

Skärmavbild 2026-05-28 kl  10 01 10


**Permanent länk för inställningarna**
-> Permanent länk för favoritbibliotek? Not all settings are copied.

**Klistra in länken i en webbläsare så kommer du till Libris med dina inställningar sparade.**
Effectively saying the same thing as the link text? Either remove it or make the link raw?

**Kopiera den här länken för att komma till Libris startsida med dina valda favoritbibliotek**
A bit worried that some read this as "click this link to apply your settings". I.e that you _need_ to use it. I think old Libris solved this by making it a an url text string. We could have a copy button beside it 🤔   

Also, maybe make the heading same size as "Hitta och lägg till favoritbibliotek" to not emphasize it too much?

olovy:
> Nice! Working as intended. I think the tricky thing here is copy. We need to make sure the purpose is clear and that this is "special need" only.
> 
> Maybe it the link should not be visible when no fav libraries?
> Skärmavbild 2026-05-28 kl 10 01 10

:+1: 

> 
> **Permanent länk för inställningarna** -> Permanent länk för favoritbibliotek? Not all settings are copied.
> 
> **Klistra in länken i en webbläsare så kommer du till Libris med dina inställningar sparade.** Effectively saying the same thing as the link text? Either remove it or make the link raw?
> 
> **Kopiera den här länken för att komma till Libris startsida med dina valda favoritbibliotek** A bit worried that some read this as "click this link to apply your settings". I.e that you _need_ to use it. I think old Libris solved this by making it a an url text string. We could have a copy button beside it 🤔

:+1: 
Yes, I had more or less the exact same thoughts.
The link text needs to be able to stand on its own but it also needs some kind of introduction.
I think making it a text string is a good solution for making it clear that you don't need to use it.

> 
> Also, maybe make the heading same size as "Hitta och lägg till favoritbibliotek" to not emphasize it too much?

 :+1:

jesperengstrom (approved):
LGTM!

olovy:
https://github.com/libris/lxlviewer/pull/1620/commits/a5c6bd854d748277146a933f1130c8f9679f5326 parse URLSearchParams in same place as building them

PR #1621, opened 2026-05-28
fix(lxlweb): more specific loan status

andersju:
We've had complaints about loan status always being shown as Unavailable/Otillgänglig even when when the actual status is On loan/Utlånad. Webbsök shows the correct status.

This is a regression, so let's be clear when we can, because "Unavailable" and "On loan" typically have two distinct meanings, where the former is used to mean that the item is missing/broken/etc.


olovy (approved):
LGTM!

PR #1626, opened 2026-06-02
fix(lxl-web/decorated-data): never render nested link elements

olovy:
Never render nested link elements in DecoratedData.

In the best of worlds data+display should not generate these links. But it has happened on multiple occasions. I think it is OK to block them here since we never want them.

Example: "är del av" http://localhost:5173/h0shkqjt07wpbzf


johanbissemattsson (approved):
LGTM! 🙌

Context _could_ be used here (e.g. `hasContext('insideLink')`) to prevent prop-drilling but I prefer your solution as all other props are currently passed along.

PR #1627, opened 2026-06-02
feat(lxlweb): Display instance category in holding btns

jesperengstrom:
### Solves

Trying out instance category instead of type to group holdings into buttons and tabs in holdings panel.

Sometimes neat:
Skärmavbild 2026-06-02 kl  15 15 10

Sometimes a bit messy:
Skärmavbild 2026-06-02 kl  15 16 09

the key used for the types is a 'select slug' from the category `@id`, this can be done now because book got the fake id `_book` in another pr. This is also used for the url so that is is not language dependent.

Let's discuss on design meeting.


olovy:
Very promising!
Let's keep exploring this.

On the work level I think we should be able to group more categories under Book and E-book. Same logic that is used in the hit list.
bild

jesperengstrom:
Pushed some improvements. Now all tests are failing for some reason.

I did not manage to group more categories however. What the select type str does in the cards is a bit of a mystery to me. 

For the panel, we also need to place every instance in a "bucket" individually. That poses a problem since an OnlineResource is only an Ebook if there are other Ebooks, right?

PR #1631, opened 2026-06-05
fix(lxlweb): show fjarrlan btn for work single instance

jesperengstrom:
### Solves

The condition for displaying fjärrlån btn did not cover works with a single linked instance:
https://libris.kb.se/scwhhrl2q74wfdmd

The instance page looks much better but is unreachable in the ui:
https://libris.kb.se/s9sjqtf6qfwzg918

* Work is missing bibliography
* Work can't get the citation right

Maybe some other maneuver fixes all this...


olovy (approved):
LGTM

PR #1633, opened 2026-06-11
fix(lxlweb): fix missing org names

jesperengstrom:
### Solves

Removing labels from myLibraries cookie resulted in missing labels in panel and on fav libraries icon on hover.
This became increasingly complicated as we don't store the names for bibdb:Orgs anywhere else.

The solution is to fetch and cache all the orgs together with the libs. The scheduled fetch now does both
`fetchLibraries()` and `fetchLibOrgs()`.

You can then get it by `getOrg(id)`
`RefinedOrgs` does this, so we get a subset of the orgs based on current selection, with needed data.

Upside: we now display (almost) fresh data.

Skärmavbild 2026-06-11 kl  13 52 56
 
Skärmavbild 2026-06-11 kl  13 54 05.

Wrote tests for the panel so this won't happen again, using a test post where (hopefully) the holdings are stable.

### Summary of changes

* fetch and cache bibdb:org in `fetchLibOrgs()`, add `getOrg` function
* Add `LibOrg` type
* Refactor `refinedOrgs`
* Needed subsequent UI changes
* rename `getLibsFromHoldings` -> `getHeldBy`
* write tests

### DIFF
--- a/lxl-web/src/lib/components/HoldingsPanel.svelte
+++ b/lxl-web/src/lib/components/HoldingsPanel.svelte
@@ -1,9 +1,14 @@
 <script lang="ts">
 	import { page } from '$app/state';
 	import { getUserSettings } from '$lib/contexts/userSettings';
-	import type { LibraryWithLinksAndInstances, OrgId, UnknownLibrary } from '$lib/types/holdings';
+	import type {
+		LibOrg,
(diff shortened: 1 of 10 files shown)


olovy (approved):
Yes! LGTM!

🏘️ 🏠

PR #1634, opened 2026-06-11
feat(lxl-web, supersearch): Shared supersearch state and history

johanbissemattsson:
### Solves

Shareable state shared across SuperSearch instances for a consistent experience.

This is required for the redesigned/more accessible start page and app bar (which renders separate supersearch instances on different media breakpoints).

Proper history is also added (and is also sharable between instances) so users can undo and redo changes.

### Summary of changes

- Add support for sharable state and history across SuperSearch instances using context.
- Add a CodeMirror history extension and shared history so undo/redo persists across instances and navigation.
- Reimplemented shallow routing in `SuperSearchWrapper.svelte` (so `packages/supersearch` has no sveltekit-specfic imports)
- Simplification of the CodeMirror component and search context handling (the context now only includes properties for accessing the "active" supersearch component and the last updated editor).


jesperengstrom:
Nice, undo state working nicely as far as I can tell. Liking the code simplification also.

Search query disappearing when entering a resource page is ok I guess. But when you reload the page, it appears again. Bug or feature?

Skärmavbild 2026-06-15 kl  11 45 07

And then when you go back in the browser, the pill breaks :(

Skärmavbild 2026-06-15 kl  11 45 57

johanbissemattsson:
> Nice, undo state working nicely as far as I can tell. Liking the code simplification also.
> 
> Search query disappearing when entering a resource page is ok I guess. But when you reload the page, it appears again. Bug or feature?
> And then when you go back in the browser, the pill breaks :(


I think this should be resolved in https://github.com/libris/lxlviewer/pull/1634/commits/48aaa49090627a30cdcbc17c6ee10f9a9d511d96

jesperengstrom:
Yes, seems to be fixed 👍 
More people might want to weigh in on hiding the query on resource pages, as it was implemented as a feature. But i'm ok with merging this to move forward.

jesperengstrom (approved):

PR #1636, opened 2026-06-12
fix(lxlweb): Show instance card for a single linked instance

jesperengstrom:
### Solves

Works with a single linked instance provides no way of navigating to that instance, while the information on work itself is not always sufficient. [Example](https://libris.kb.se/7s1nh5xm50c5w929)

In the long term, information should maybe be combined in the manner of local work + instance.
But for now tweaking the condition for when displaying the instance list.

Also make sure cite button does not appear for Work.

Skärmavbild 2026-06-12 kl  13 30 55

[Example](http://localhost:5173/7s1nh5xm50c5w929?_q=seriesMembership%3A%28Doktorns+d%C3%B6ttrar%29)
[More examples in series](http://localhost:5173/find?_q=seriesMembership:%28Doktorns+d%C3%B6ttrar%29)

### Summary of changes

* Tweak instances list condition
* Tweak cite button condition

### DIFF
--- a/lxl-web/src/lib/components/Resource.svelte
+++ b/lxl-web/src/lib/components/Resource.svelte
@@ -311,7 +311,7 @@
 								{page.data.t('resource.moreDetails')}
 							</a>
 						{/if}
-						{#if instances?.length === 1 && fnurgel}
+						{#if !isWork && instances?.length === 1 && fnurgel}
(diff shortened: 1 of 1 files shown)


olovy (approved):
LGTM!

PR #1635, opened 2026-06-12
feat(lxl-web): Redesigned start page with hero image

johanbissemattsson:
### Solves

An much-needed visual refresh to the start page – adding hero images .

### Summary of changes

- Redesigned the start page with a hero image
- Reworked app bar and app search styling
    - Better tab-ordering
    - Larger contents on larger breakpoints
    - Visible labels on smaller breakpoints
    - Slightly better support for subsets
- Added `@sveltejs/enhanced-img` for optimized hero images
- Subpath imports for `app.css`
- Adjusted margins on sub pages

PR #1637, opened 2026-06-12
feat(lxlweb): handle all uris

jesperengstrom:
### Solves

Handle all uri:s as a resource, for example http://localhost:5173/library/LOB

`[fnurgel=fnurgel]` and `params.fnurgel` is no more

instead `[...resource]` works as a fallback for everything that is not a defined route.

Meaning we will try to fetch `/000n6qjj10vt6ss5`,  `/library/LOB` and `/some-route` alike, and go to 404 when recieved.

Vaguely related to https://github.com/libris/lxlviewer/pull/1638

Description of what this solves
* Replace `fnurgel` route with `...resource`
* replace `page.params.fnurgel` with a sliced `data.recordId`


olovy (approved):
Nice! LGTM.

We can decide later if test environments should rewrite links they display.

Or maybe we should display _both_ record-id and thing-id when 
a) thing URI does not start with BASE_URI 
b) only when record has VirtualRecord type

See e.g. http://libris-qa.kb.se/k02w4jdhh0bvtpk8

jesperengstrom:
If we want to merge https://github.com/libris/lxlviewer/pull/1635 soon, we should probably merge this after to avoid conflict caused by renaming folder...

olovy:
> We can decide later if test environments should rewrite links they display.
> 
> Or maybe we should display _both_ record-id and thing-id when a) thing URI does not start with BASE_URI b) only when record has VirtualRecord type
> 
> See e.g. http://libris-qa.kb.se/k02w4jdhh0bvtpk8

Basic solution: https://github.com/libris/lxlviewer/pull/1642

PR #1638, opened 2026-06-12
fix(lxlweb): Fix adjacent search results using non-fnurgels

jesperengstrom:
### Tickets involved
[LWS-427](https://kbse.atlassian.net/browse/LWS-427)

### Solves

Adjacent search results not working when fetching next/prev results for Bilbiographies, subjects etc. Non-fnurgel links resulting in 404 page.

This is because when `adjacentSearchResults` fetches new results, they are not formatted. 
Unformatted results have record id:s in `meta.@id`
`asResult` puts them in `@id`. Hence the confusion. 

Format before using. We should however manage both variants . See https://github.com/libris/lxlviewer/pull/1637

### Summary of changes

* use `asResult` in `getAdjecentSearchResult`


olovy (approved):
LGTM! 

Works as expected.

PR #1639, opened 2026-06-15
fix(lxl-web): display all contributors in hasPart

olovy:
* Show all contributors in Work hasPart. Not only primary contribution.
    * Put translationOf last with an inline label. (Instead of "=" which is wrong)
    * Do more of an ISBD style with period + em-dash separating props for one part/work.
* Fix empty hasPart displayed when part only contained language
* See follow-up PR which fixes rendering of bullet lists (by actually using list element)

### Tickets involved
[LWS-437](https://kbse.atlassian.net/browse/LWS-437)

Examples
http://localhost:5173/find?_q=hasPart.translationOf%3A()

Skärmbild från 2026-06-15 13-16-17

Skärmbild från 2026-06-15 13-18-04

Fix empty hasPart disaplyed when part only contained language:
Skärmbild från 2026-06-15 13-17-14

PR #1640, opened 2026-06-15
feat(lxl-web/decorated-data): Use ul for bullet lists

olovy:
Follow-up to https://github.com/libris/lxlviewer/pull/1639 (stacked)

* use actual ul for vertical listing of values.
    * Instead of faked lists (block with ::before bullet points). Fixes alignment when line breaks inside list values. And of course more semantically correct.
    * activated by setting propertyStyle `ul` or `ul-when-multiple` (should we name them in a way that indicates that they are "special"? Didn't use `ul()` since "stylers" are all implemented in one place)
    * unfortunately big diff in `DecoratedData`. It's mostly because of the conditional `<li>` that has to be inserted. I think the recursion structure of the `DecoratedData` component can be refactored later. It became a bit more ugly now with both `isLi` (render as `<li>`) and `isLiChild` (no contentBefore/contentAfter in child of `<li>`).
* remove a lot of special casing CSS in Resource.svelte
* improve display of hasPart/relationship Work a little - semibold titles.
* bonus: semibold PrimaryContribution agent https://github.com/libris/lxlviewer/commit/9ef15b9bd95305ebbc267424b8ebfb4d74d16a37
* bonus: semibold PrimaryPublication in details (when multiple). Show marc:sequenceStatus, appliesTo, e.g."Sammanfattad utgivningstid", "Nuvarande / Senaste utgivare" https://github.com/libris/lxlviewer/commit/ed8713af7a9fe552acc74eeb447ef97fa8866e79

Skärmbild från 2026-06-15 14-17-18
Skärmbild från 2026-06-15 13-36-55

Skärmbild från 2026-06-15 14-19-07

### DIFF
--- a/lxl-web/src/lib/assets/json/display-web.json
+++ b/lxl-web/src/lib/assets/json/display-web.json
@@ -299,7 +299,9 @@
 						"startYear",
 						"endYear",
 						"date",
-						"marc:publicationStatus"
+						"marc:publicationStatus",
+						"marc:sequenceStatus",
+						"appliesTo"
 					]
 				},
 				"ProvisionActivity": {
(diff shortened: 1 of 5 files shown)


jesperengstrom (approved):
LGTM!

PR #1642, opened 2026-06-16
feat(lxl-web): display separate 'permalinks' if external resource uri

olovy:
Display two links if we are not the ones serving the resource ID.

image

image

Still one otherwise
image


johanbissemattsson (approved):
LGTM!

jesperengstrom (approved):
LGTM!

PR #1644, opened 2026-06-17
feat(lxl-web, supersearch): Shared supersearch state and history (Revised)

johanbissemattsson:
# Revised PR (which replaces reverted https://github.com/libris/lxlviewer/pull/1634) 

### Solves

Shareable state shared across SuperSearch instances for a consistent experience.

This is required for the redesigned/more accessible start page and app bar (which renders separate supersearch instances on different media breakpoints).

Proper history is also added (and is also sharable between instances) so users can undo and redo changes.

### Summary of changes

- Add support for sharable state and history across SuperSearch instances using context.
- Add a CodeMirror history extension and shared history so undo/redo persists across instances and navigation.
- Reimplemented shallow routing in `SuperSearchWrapper.svelte` (so `packages/supersearch` has no sveltekit-specfic imports)
- Simplification of the CodeMirror component and search context handling (the context now only includes properties for accessing the "active" supersearch component and the last updated editor).

#### Added since #1634
- Fix issue with spaces added on chrome
- Fix missing parens when adding multiple languages


jesperengstrom (approved):
Nice. Working nicely now as far as I can tell.

PR #1647, opened 2026-06-22
fix(lxlquery): handle unclosed quotes + pills

jesperengstrom:
### Solves

Needs to be thoroughly tested…

An attempt to fix the fact that a single quote char can break the pill rendering:

gif1.

This is of course because the `"` makes everything up until the next quote parse as a `strLiteral`, swallowing the subsequent qualifier.

We need to allow the stray `”` to parse as something allowed inside a QualifierValue, in order for the qualifier to stay intact. 
If we add a new token for the incomplete quoted string, `StrliteralIncomplete`, and include it in the definition of `String`, it now "adds upp", allowing the rest to stay intact too. (a little deceiving, because it is in fact an invalid query)

gif2.

But we still have a problem:

gif3.

The parser still sees the quoted string `” contribution:(”` instead of an incomplete string and a qualifier.

The only thing I've come up with is to disallow certain chars in quoted strings. When the parser sees `()`,  or `():=~<>` in a incomplete quoted string, it breaks and instead goes with the qualifier:

gif4.

**Downsides / tradeoffs**

More convoluted grammar
We lose the "error highlighting" of unclosed quotes, since it's now part of the grammar
Will probably never be 100% waterproof:

the query
`" title:"hello"`

Will still be parsed as a the string literal `" title:"` (as long as we don't disallow more chars here)

This should only be merged in combination with a fix to make sure the api don't wipe groups () from the query. These are now key to recognizing qualifiers in these cases.

### Summary of changes

* Update grammar

### DIFF
--- a/packages/codemirror-lang-lxlquery/src/syntax.grammar
+++ b/packages/codemirror-lang-lxlquery/src/syntax.grammar
@@ -7,7 +7,7 @@ andcomb { term ( AndOperator term | term )* }
 term { String | Group | UTerm | Qualifier }
 Group { "(" orcomb? ")" }
 UTerm { UOperator term }
-String { strliteral | str }
+String { strliteral | StrliteralIncomplete | str }
 Qualifier { 
   ( QualifierKey { String } QualifierOperator { bOperator } QualifierValue { String } ) | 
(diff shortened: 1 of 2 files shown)


johanbissemattsson:
Seems to work fine now! 🙌 LGTM!

PR #1650, opened 2026-07-23
feat(lxlweb): make library fetching more robust; warm up caches on startup

andersju:
Previously, if `startRefreshLibraries` failed (for example, because the EMM API was unavailable), lxlweb would run with no library/org data for at least 12 hours (until a regularly scheduled refresh succeeded).

This PR does a couple things:

- Loading/caching of libraries/orgs/util is done immediately at start-up instead of when the first request comes in (but it deliberately still doesn't block any requests if loading hasn't finished).
- If the initial `refreshLibraries` fails, it keeps trying every 30 seconds instead of waiting for 12 hours. Only after a successful fetch does it schedule the 12-hour refresh interval.

Related to #1649

### DIFF
--- a/lxl-web/src/hooks.server.ts
+++ b/lxl-web/src/hooks.server.ts
@@ -1,5 +1,10 @@
 import fs from 'fs';
-import { type HandleServerError, redirect, type RequestEvent } from '@sveltejs/kit';
+import {
+	type HandleServerError,
+	redirect,
+	type RequestEvent,
+	type ServerInit
+} from '@sveltejs/kit';
 import { env } from '$env/dynamic/private';
 import { defaultLocale, Locales } from '$lib/i18n/locales';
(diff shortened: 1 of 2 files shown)


johanbissemattsson (approved):
LGTM!

A slightly more SvelteKit_ish_ way would be using an [init hook](https://svelte.dev/docs/kit/hooks#init) but it should be no different than writing the initialisation logic at the top level of the module.

andersju:
> LGTM!
> 
> A slightly more SvelteKit_ish_ way would be using an [init hook](https://svelte.dev/docs/kit/hooks#init) but it should be no different than writing the initialisation logic at the top level of the module.

Good point! I adjusted it a bit.

johanbissemattsson (approved):
🙌

